PR #42, opened 2025-08-31
41 add volunteer form

### LINKED ISSUE #41: Add volunteer form
## Description 

Copypaste forms folder and add volunteer form

### this includes:

1. Copypaste files
2. Update stylings
3. Copypaste fallback option lists
4. Wire route
5. Update DTO

jagdishlove:
feat(forms): add volunteer & opportunity forms

- Added types for Volunteer and Opportunity forms
- Migrated code from React to Next.js
- Fixed layout file errors
- Added utils and helper functions
- Added custom hooks
- Added English (en) language support

Closes #41


.npmrc
```diff
@@ -0,0 +1,2 @@
+legacy-peer-deps=true
```
arturasmckwcz: for what dependencies is this needed?

.npmrc
```diff
@@ -0,0 +1,2 @@
...
+force=true
```
arturasmckwcz: this shouldn't be applied to all yarn CLI calls, so please remove it and use `--force` if it is required only

arturasmckwcz: in general I'd consider leaving those setting local by adding `.npmrc` to the `.gitignore`

public/locales/en/translations.json
```diff
@@ -18,9 +18,17 @@
       "resources": "RESOURCES",
       "faq": "FAQ",
       "legalNotice": "Legal notice",
-      "privacy": "Privacy",
       "vpa": "Volunteer Participation Agreement",
-      "contact": "Contact"
+      "privacy": "Privacy",
```
arturasmckwcz, line 22: please add German translations as well

.npmrc
aricionur: what is  this new file for ?

src/app/layout.tsx
```diff
@@ -1,55 +1,43 @@
...
+        <link
+          href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.3/dist/css/bootstrap.min.css"
+          rel="stylesheet"
+          integrity="sha384-QWTKZyjpPEjISv5WaRU9OFeRpok6YctnYmDr5pNlyT2bRjXh0JMhjY6hW+ALEwIH"
+          crossOrigin="anonymous"
+        />
+      </head>
+      <body className={figtree.variable}>
         <StyledComponentsRegistry>
-          <DeviceProvider initialScreenType={initialScreenType}>
```
aricionur, line 50: why did you remove this?

src/app/layout.tsx
```diff
@@ -1,55 +1,43 @@
...
   children: React.ReactNode;
 }>) {
-  const initialScreenType = await getInitialDeviceType();
-
   return (
     <html lang="en">
-      <body className={`${geistSans.variable} ${geistMono.variable}`}>
+      <head>
+        <link
+          href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.3/dist/css/bootstrap.min.css"
```
aricionur, line 45: do we need this bootstrap link?

.npmrc
jagdishlove: > what is this new file for ?

Many older packages still declare outdated peer deps → so npm throws errors or refuses to install.

- With legacy-peer-deps=true, npm behaves like npm v6:

- Ignores peer dependency conflicts during install.

But i forgot to remove this file and commited.
I will remove it

src/app/layout.tsx
jagdishlove: > why did you remove this?

I created `fe/src/utils/ClientWrapper.tsx` as a wrapper component to keep `layout.tsx` cleaner and more maintainable. The idea is to separate out client-side context/providers (like DeviceProvider) into a dedicated wrapper, so that `layout.tsx` focuses only on layout structure. This also makes it easier to scale later if we need to introduce additional global providers or client-only logic without cluttering the layout.

src/app/layout.tsx
jagdishlove: > do we need this bootstrap link?

Since we’re migrating from React to Next.js, I noticed that in the React project we had the Bootstrap  CDN included in `index.html`. That meant Bootstrap was globally affecting some tag-level styling (like buttons, forms, grids, etc.). When I moved the same code into Next.js, those global Bootstrap styles were missing because Next.js doesn’t use a plain `index.html`.

To make sure the layout and components looked consistent with the React version, I added the Bootstrap CDN link inside layout.tsx. This keeps the styling behavior the same across both projects. Without including it, some default styles would differ because Bootstrap’s overrides wouldn’t be applied.

src/app/layout.tsx
aricionur: I see,
I aggree with you about making layout.tsx and other high level files as cleaner as possible.

however, the serverside rendering nature of nextjs is not allowing the code set its initial screen type using hooks .
I checked your `ClientWrapper.tsx` file and you are giving initial values as `Desktop` which casues a flash effect for the tablets and mobile phones on the screen(especially logos).

you can check that effect by using browser developer tools to check how it looks like for mobile and desktops.
please select phone device and refresh the page `http://localhost:3000/dashboard/home`

Hence, we should extract the device type at first http request.

Thas why I added  getInitialDeviceType function at the` layout.tsx` file.

I will later put it another file and call it in  ` layout.tsx`

So could you revert ` layout.tsx` file and remove ClientWrapper.tsx and related codes ?

DeviceProvider is handling the screen size changes as well.

.npmrc
aricionur: I see,
Actually, we use yarn to install packages instead of yarn .

so if you use yarn, there shouldn't be any problem.

src/app/layout.tsx
aricionur: I got it.

lets keep this bootstrap link. 

later we will remove this in future after refactoring all these copy-paster form pages by using `styled-components`

aricionur (approved):

arturasmckwcz (approved):

public/locales/en/translations.json
jagdishlove: > German

Done

PR #60, opened 2025-09-29
37 set language from query param

### LINKED ISSUE #37: set language from Query param
- when a user is selected deutch, a Nextjs hydriation error occurs due to the SSR nature.
- add a query param if user selectes any language other than EN
- parse and set the languate from query param of the http reqeust
`?language=de`

aricionur:


arturasmckwcz (approved):
thank you!

PR #78, opened 2025-10-08
21 add volunteer filters

### LINKED ISSUE #21: add volunteer filters
## Description

add filters
volunteers filters designs are missing

aricionur:
closes #21


src/components/Dashboard/Volunteers/Filters/constants.tsx
```diff
@@ -0,0 +1,58 @@
+import { ByDay, OccasionalType } from "need4deed-sdk";
```
arturasmckwcz: these guys need to be shared in SDK

arturasmckwcz: and please check if you could re-use something, like `TimeSlot`

src/components/Dashboard/Volunteers/Filters/constants.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export enum TimeSlot {
+  morning = "08-11",
+  noon = "11-14",
+  afternoon = "14-17",
+  evening = "17-20",
+}
+
+export enum FilterKeys {
+  SEARCH = "search",
```
arturasmckwcz: SEARCH won't go to the API, right?
my suggestion would be:
```
import { FilterKeys as ApiFilterKeys } from "need4deed-sdk";

const fiterKeys = {
  ...ApiFilterKeys,
  SEARCH: "search",
}

type FilterKeys = typeof fiterKeys[keyof fiterKeys];
```

then `FilterKeys` will be `"search" | "engagement | ...`

something like that. WDYT?

src/components/Dashboard/Volunteers/Filters/constants.tsx
aricionur: I checked the sdk, there is no similar definition for `TimeSlot`, so created new one.

You're right, I added `TimeSlot` and `FilterKeys` to the `sdk`.

But for  `FilterKeys`. I just defined it for volunteers. after adding it into sdk,  for other listing pages we have to add new keys into it.
 
I hope, it will not be a too big enum definition in future..

src/components/Dashboard/Volunteers/Filters/constants.tsx
aricionur: Nope.

SEARCH will go to the API.

frontend do not have all data.

src/components/Dashboard/Volunteers/Filters/constants.tsx
aricionur: Ohh you alread added `FilterKeys` into sdk as `QueryParamsKeys` 

I will refactor fe branch code...

aricionur: FIXED.

arturasmckwcz (approved):
thank you!!!

PR #83, opened 2025-10-17
82 update volunteer data structure for card list

### LINKED ISSUE #82: update volunteer data structure for card list
## Description

the API has been changed so that now `GET /volunteer` serves option lists, like districts, activities, etc. not as `string[]` but as `OptionItem[]`
so frontend components dealing with that structure need to be updated.

arturasmckwcz:
closes: #82

mostly tackling typescript errors
1. downgraded `@tanstack/react-form`. they rolled over from 0 to 1 and many internal Generics don't work anymore.
2. casted types there and there
3. and added the helper for converting `OptionItem` to string in order to adapt to nes data structure from API

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "@phosphor-icons/react": "^2.1.10",
-    "@tanstack/react-form": "^1.12.3",
+    "@tanstack/react-form": "^0.26.4",
     "@tanstack/react-query": "^5.80.7",
     "axios": "^1.11.0",
     "classnames": "^2.5.1",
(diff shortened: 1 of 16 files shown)


552020 (dismissed):
@arturasmckwcz `yarn build` fails!

Image

I checkout the branch, rm -rf the node_modules, re-installed them with yarn install and then run yarn build.

arturasmckwcz:
> @arturasmckwcz `yarn build` fails!
> 
@552020 thanks for pointing that out. we didn't arrive to building yet, we only run `yarn dev` and never hit `Announcement` component to get that exception.
fixing.

need4deed (approved):

Issue #98, opened 2025-10-26
availability edit grid


arturasmckwcz:
## Description

[Figma](https://www.figma.com/design/9Kl17OiDLbWoNxgY43Rgjk/Need4Deed-Designs-2025?node-id=8714-120601&t=TG2xMN8FomvnT4Gd-4)

gets volunteer object and change function

DarrellRoberts:
hey @arturasmckwcz , I'd love to take on this issue

arturasmckwcz:
> hey [@arturasmckwcz](https://github.com/arturasmckwcz) , I'd love to take on this issue

great! please ask any question here or on slack

PR #105, opened 2025-10-30
restyle availability edit grid #98

DarrellRoberts:
Closes issue #98 

Notes: 
- Didn't see a mobile design on figma, so used responsive columns for now
- The full-width of the grid is dependant on its parent(s). The entire form redesign is a part of a bigger task I believe so I didn't want to tamper with the form's structure and just worked with a temporary solution
- For the shortening of the days of week I changed the i18n jsons rather than using a str method

#### Desktop
image

#### Mobile
image

### DIFF
--- a/src/components/forms/index.module.css
+++ b/src/components/forms/index.module.css
@@ -1,6 +1,4 @@
 .form-container {
-  /* disabled to get full width for availability grid */
-  /* width: clamp(40ch, 80vw, 65ch); */
   display: flex;
   flex-direction: column;
   align-items: center;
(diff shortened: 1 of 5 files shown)


src/components/forms/index.module.css
```diff
@@ -1,5 +1,6 @@
 .form-container {
-  width: clamp(40ch, 80vw, 65ch);
+  /* disabled to get full width for availability grid */
```
aricionur: Could you remove comments please?

DarrellRoberts: hi @aricionur , sure thing, I've removed it in the latest commit

aricionur: thank you

aricionur (approved):

PR #115, opened 2025-11-11
Improve docker build

drndivoje:
- decrease size of the docker image
- disable telemetry tracking for nextjs


package-lock.json
```diff
@@ -0,0 +1,6236 @@
+{
```
arturasmckwcz: I can't see where we use this file

arturasmckwcz (approved):

package-lock.json
drndivoje: actually it should be removed because yarn is used to build the app

PR #130, opened 2025-11-24
feat: added single custom hook for POST/PUT/PATCH calls.

### LINKED ISSUE #128: Create single custom hooks for POST,PUT and PATCH API using React-Query
### Description

- Create A global custom hook which can use for POST, PUT and PATCH, as right now we only have for POST custom hook(usePostMutation), which can extend further into these http calls and reuse it in efficient way.

### Proposed solution

Custom hook to fetch PUT, POST and PATCH.

jagdishlove:
## Description
The change is I modified usePostMutation custom hook into useMutationQuery which will have POST,PUT and PATCH http call which will make this hook more efficient to to reuse in project in any CRUD update API call.

## Related Issues
Closes #128


src/hooks/useMutationQuery.tsx
```diff
@@ -0,0 +1,75 @@
...
+  method?: HttpMethod;
+  successMessage?: string;
+  onSuccessCallback?: (data: TResponse) => void;
+  queryKeyToInvalidate?: (string | number)[]; // Optional key to refetch data on success
+}
+
+// Generic function to perform the API call
+async function mutateData<TData, TResponse>(apiPath: string, method: HttpMethod, data: TData): Promise<TResponse> {
+  // Dynamically call the appropriate axios method
+  const response = await axios[method.toLowerCase() as "post" | "patch" | "put"](apiPath, data);
```
aricionur: It would be nicer to define those as Axios methods as a type like:
type axiosMethods =  "post" | "patch" | "put"

aricionur (approved):

src/hooks/useMutationQuery.tsx
jagdishlove: > It would be nicer to define those as Axios methods as a type like: type axiosMethods = "post" | "patch" | "put"

Fixed.

Issue #88, opened 2025-10-24
create volunteer editable profile logic


arturasmckwcz:
### this includes:
1. volunteer object fetching
2. volunteer patching
3. adding comments and other timeline objects

PR #159, opened 2025-12-12
#88: volunteer object fetching

HansKre:
## Description
Uses API for volunteer-object fetching and default values

image

### DIFF
--- a/src/components/Dashboard/Profile/sections/VolunteerHeader.tsx
+++ b/src/components/Dashboard/Profile/sections/VolunteerHeader.tsx
@@ -2,7 +2,12 @@
 import Image from "next/image";
 import styled from "styled-components";
 import { formatDateTime, getImageUrl } from "@/utils";
-import { ApiVolunteerGet } from "need4deed-sdk";
+import {
+  ApiVolunteerGet,
+  VolunteerStateEngagementType,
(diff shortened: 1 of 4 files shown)


src/components/Dashboard/Profile/sections/VolunteerHeader.tsx
```diff
@@ -112,6 +117,10 @@ export function VolunteerHeader({ volunteer }: Props) {
 
   const initialAvatarUrl = getImageUrl(volunteer.person.avatarUrl || defaultAvatarVolunteerProfile);
 
+  const statusEngagement = volunteer.statusEngagement || VolunteerStateEngagementType.NEW;
```
arturasmckwcz, line 120: I'm not entirely sure we need fallback values here. I already asked Nadav and Darya on [Figma](https://www.figma.com/design/9Kl17OiDLbWoNxgY43Rgjk/Need4Deed-Designs-2025?node-id=10548-258199&t=wXCdW3iN3kFA0mDQ-4)

but anyway setting fallback value should happen on the backend IMHO

arturasmckwcz (commented):
thanks a lot!

arturasmckwcz (approved):
thank you!

Issue #90, opened 2025-10-24
volunteer contact details container


arturasmckwcz:
## Description

[general view](https://www.figma.com/design/9Kl17OiDLbWoNxgY43Rgjk/Need4Deed-Designs-2025?node-id=8708-119794&t=ULYX4dap6kmN2oaU-4)

it contains copies, "Edit" button, and 4 editable field components (developed in another issue)

### behaviour

initially fields render in display mode
by clicking "Edit" fields re-render in edit mode and instead "Edit" "Cancel" and "Save changes" buttons render

cezene:
Working on this issue

arturasmckwcz:
@cezene great! please ask anything here or on slack

PR #161, opened 2025-12-12
#90: volunteer contact details container

HansKre:
## Description
Implements https://github.com/need4deed-org/fe/issues/90

## Required backend changes:

1. PATCH-Request fails without a `city`-column on the `address.entity.ts`.

  Requires something like:

  ```ts
  @Column({ nullable: true })
  @IsString()
  @Length(100)
  city: string;
  ```

2. `waysToContact`: frontend expects and sends array of lowercase string keys, e.g.:                                                                                                                                                                           
  `["whatsapp", "telegram", "mobilePhone", "email", "sms"]`. Frontend uses those string keys for `i18n`.

## Refactoring in subsequent PR

- [ ] Rename `ContactDetails` to `ContactDetailsSection`

### DIFF
--- a/src/components/Dashboard/Profile/ProfileController.tsx
+++ b/src/components/Dashboard/Profile/ProfileController.tsx
@@ -5,7 +5,7 @@ import ProfilePage from "./ProfilePage";
 
 export function ProfileController({ volunteerId }: { volunteerId: string }) {
   const { data } = useGetQuery<ApiVolunteerGet>({
-    queryKey: ["volunteers", volunteerId],
+    queryKey: ["volunteer", volunteerId],
     apiPath: `${apiPathVolunteer}${volunteerId}`,
     staleTime: cacheTTL,
   });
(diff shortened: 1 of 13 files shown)


src/components/Dashboard/Profile/ProfileController.tsx
```diff
@@ -5,7 +5,7 @@ import ProfilePage from "./ProfilePage";
 
 export function ProfileController({ volunteerId }: { volunteerId: string }) {
   const { data } = useGetQuery<ApiVolunteerGet>({
-    queryKey: ["volunteers", volunteerId],
+    queryKey: ["volunteer", volunteerId],
```
arturasmckwcz, line 8: this is actually deeply nested eager loading on the backend side, including comments, opportunities, etc.
just FYI

arturasmckwcz (approved):
thank you!

Issue #124, opened 2025-11-21
volunteer profile document viewer


arturasmckwcz:
### Description

[Figma](https://www.figma.com/design/9Kl17OiDLbWoNxgY43Rgjk/Need4Deed-Designs-2025?node-id=10548-255703&t=6qiG4dOP3DJz3MaL-4)

Issue #123, opened 2025-11-21
volunteer profile document section container


arturasmckwcz:
### Description

[Figma](https://www.figma.com/design/9Kl17OiDLbWoNxgY43Rgjk/Need4Deed-Designs-2025?node-id=8720-122049&t=6qiG4dOP3DJz3MaL-4)

### this includes:
1. header
2. document grid with predefined copies
3. logic including document download, preview, and delete functions.

### gets:
1. volunteer object (or id only???)
2. document objects list

PR #172, opened 2025-12-25
#123 & #124: volunteer profile document section

HansKre:
## Description

Implements #123

>Note: Backend uses mock-implementation for doc-upload and returns always the same dummy-document.

image

image

Implements #124

image


arturasmckwcz:
Screenshot 2026-01-09 at 12 13 33

looks like i18n translations are missing or ~wrong arguments are given to t function~
english is okay, so something's wrong in de/translation.json

the problem:
```
node -e "console.log(require('./public/locales/de/translations.json').dashboard.volunteerProfile.documentSection)"
undefined
```

HansKre:
> Screenshot 2026-01-09 at 12 13 33
> looks like i18n translations are missing or ~wrong arguments are given to t function~ english is okay, so something's wrong in de/translation.json
> 
> the problem:
> 
> ```
> node -e "console.log(require('./public/locales/de/translations.json').dashboard.volunteerProfile.documentSection)"
> undefined
> ```

Those look like the merge-issues that I had to solve due to the concurrent refactoring of `ProfilePage`. They should be solved by now. For me, it looks like this in German:

image

However, now that you mentioned it and I paid closer attention, I noticed, that switching to English didn't work anymore indeed. It was much more subtle for me, then for you, though. Turns out, again the aforementioned merge-issue. Hopefully, we won't have concurrent refactoring in future.

arturasmckwcz (approved):
thank you!

PR #191, opened 2026-01-14
#167 after successful login

### LINKED ISSUE #167: after successful login
### Description

after successful login a user has to be redirected to dashboard which matches user role value: `admin`, `volunteer`, `coordinator`, `agent`.
OR to the home page if role is `user`

### Proposed solution

_No response_

DarrellRoberts:
## Description
redirects user to a url path according to their role and user id

## Related Issues
[closes #issue 167](https://github.com/need4deed-org/fe/issues/167)

## Changes
- Add auth.ts file
- Redirect after successful login

## Screenshots / Demos

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes


src/utils/auth.ts
```diff
@@ -0,0 +1,12 @@
+import { apiPathMe } from "@/config/constants";
+import axios from "axios";
+import { ApiUserGet } from "need4deed-sdk";
+
+export const getAuthUser = async (): Promise<ApiUserGet | null> => {
```
arturasmckwcz: I recon it'd be beneficial using:
`/src/hooks/useGetQuery.tsx`
it'd cache `ApiUserGet` object for further use, by calling query with the same `queryKey`
@aricionur, @HansKre  WDYT?

HansKre: Agree, and if we consider `ApiUserGet` as slow changing data, we can go really high with the `staleTime`. An example usage in `src/components/Dashboard/Profile/ProfileController.tsx`:

```ts
  const { data } = useGetQuery<ApiVolunteerGet>({
    queryKey: ["volunteer", volunteerId],
    apiPath: `${apiPathVolunteer}${volunteerId}`,
    staleTime: cacheTTL,
  });
```

DarrellRoberts: thank you @arturasmckwcz and @HansKre . I've commited my changes now so have switched from serverside to tanQuery fetching. I found I had to modify the useGetQuery hook for it to work so not sure if you agree. In the hook I added a conditional enabled property & added additional conditional checking to the return

src/hooks/useGetQuery.tsx
```diff
@@ -81,7 +83,7 @@ export const useGetQuery = <T,>({ queryKey, apiPath, params = {}, staleTime }: U
   }, [isError, error, t]);
 
   return {
-    data: data?.data,
+    data: (data && Object.prototype.hasOwnProperty.call(data, "data") ? data.data : data) as T,
```
HansKre: Not sure, if this is the right way to do it.

We had:

Image

Which is correct, since there could be an error.

With the casting, TypeScript from now on assumes, `data` is always defined and of type `T`, even if there's an error. Not sure, if we want this.

HansKre: If really needed, the right way to do it would propably be using discriminated unions, e.g.:

```ts
type Respone = BaseResponse & (
  {
    isError: true;
    error: SomeErrorType;
    data?: never;
  } ||
  {
    isError: false;
    error?: never;
    data: T;
  }
)
```

src/components/Login/LoginForm.tsx
```diff
@@ -22,20 +18,12 @@ interface LoginResponse {
...
     apiPath: apiPathLogin,
     successMessage: "dashboard.login.successMessage",
     onSuccessCallback: async () => {
-      const user = await getAuthUser();
-      if (user && user.role !== UserRole.USER) {
-        router.push(`/${language}/dashboard?role=${user.role}&userId=${user.id}`);
-      } else {
-        router.push(`/${language}`);
-      }
+      onLoginSuccess?.();
```
HansKre:
```suggestion
      onLoginSuccess();
```

src/components/Login/LoginController.tsx
```diff
@@ -0,0 +1,32 @@
...
+  useEffect(() => {
+    if (user && "role" in user) {
+      const path =
+        user.role !== UserRole.USER ? `/${language}/dashboard?role=${user.role}&userId=${user.id}` : `/${language}`;
+
+      router.push(path);
+    } else {
+      return;
+    }
+  }, [user, language, router]);
```
HansKre, line 25: Use early exit to reduce nesting:

```suggestion

  useEffect(() => {
    if (!(user && "role" in user)) return;
    const path =
      user.role !== UserRole.USER ? `/${language}/dashboard?role=${user.role}&userId=${user.id}` : `/${language}`;

    router.push(path);
  }, [user, language, router]);
```

src/hooks/useGetQuery.tsx
DarrellRoberts: thank you @HansKre , I've added a discriminated union to handle error, loading & success responses. Also wrote a cleaner line for the data return

DarrellRoberts:
thanks so much for reviewing @HansKre ! I've committed my changes

src/hooks/useGetQuery.tsx
```diff
@@ -36,6 +36,28 @@ interface ApiResponse<T> {
...
+        data?: never;
+      }
+    | {
+        isError: false;
+        isLoading: false;
+        error?: never;
+        data: T;
+      }
+  );
+
```
HansKre: You got the type right, bravo 👍 👏 

However, tbh, I don't really understand the problem that you are trying to solve with this. It creates a lot of overhead and the way you're using it, involves explicit casting (which is not the idea of discriminated unions usually).

I reverted the `useGetQuery` and added only the `enabled`-property and the entire project keeps building.

Unless we have a real benefit from the added complexity, I'd probably not do it.

DarrellRoberts: hi @HansKre , so sorry for the delayed response, I was away. 

Thanks for the feedback.

The reason I introduced that logic was to solve a functional issue: endpoints like /user/me which I need for the redirect return the object directly ({...}), while others return a wrapped response ({ data: {...} }). 

data.data is therefore undefined for the /user/me endpoint, and the hook fails to return the actual user data.

Is there way to handle this whilst keeping the hook simple?

HansKre: ~~I understand the motivation.
Does it make sense to you then, that project builds also without the typing?~~

I now understand the issue, @DarrellRoberts 

@arturasmckwcz should we harmonize backend-returns to avoid unnecessary fe-complexity? Could we have `GET /api/user/me` return a wrapped response `({ data: {...} })`?

Should be doing it according to the swagger:

image

But actual response looking like that:

image

arturasmckwcz:
Steve already did the change, please check PR#227
<https://github.com/need4deed-org/be/pull/227>
already merged

HansKre:
Awesome! In that case, @DarrellRoberts should be good to revert the changes to `useGetQuery`-hook, other than adding `enabled`-prop, right?

DarrellRoberts:
thank you @arturasmckwcz and @HansKre! I've now reverted the useGetQuery with exception of the enabled prop and rebased my branch

HansKre (approved):
Thank you!
LGTM 👍

Issue #89, opened 2025-10-24
create  modal for status change


arturasmckwcz:
## Description

uses toaster!
date picker is out of this scope, please use prop and for dev purpose please default it to `<input type="date" />`.

it gets
1. volunteer object
2. function for patching

[Figma](https://www.figma.com/design/9Kl17OiDLbWoNxgY43Rgjk/Need4Deed-Designs-2025?node-id=10548-258201&t=RcHjncaxm6FQ2bhU-4)

IronTroll:
Hi, I would like to work on this issue. Please assign it to me if possible. I am excited to contribute to the project.

arturasmckwcz:
> Hi, I would like to work on this issue. Please assign it to me if possible. I am excited to contribute to the project.

hi, @IronTroll it's not ready for dev yet, please check another issue

Issue #70, opened 2025-10-04
volunteer header (statuses) container


arturasmckwcz:
## Description

1. gets volunteer object
2. gets a handler for click on "Change status"

[Figma](https://www.figma.com/design/9Kl17OiDLbWoNxgY43Rgjk/Need4Deed-Designs-2025?node-id=10548-255312&t=6qiG4dOP3DJz3MaL-4)

dawole94:
hello, can I take this issue?

arturasmckwcz:
> hello, can I take this issue?

@dawole94 this is already taken by @jagdishlove only wasn't marked

arturasmckwcz:
@jagdishlove please check Figma link I updated it

PR #196, opened 2026-01-22
#70: VolunteerHeader statuses

HansKre:
## Description

Implements #70 and #89

image

image

image


arturasmckwcz (approved):
thank you!!!

PR #202, opened 2026-01-26
Fix I18nProvider

HansKre:
## Description

Fixes de + en requests for en-locale:

<img width="750" height="84" alt="image" src="https://github.com/user-attachments/assets/79c59549-5d57-4b35-8e6e-94f362b58e86" />


Changed `useEffect` → `useMemo` in `I18nProvider`.                                                                                            
                                                                                                                                          
  - `useEffect` runs AFTER render → children already fired requests with wrong language (de)                                                
  - `useMemo` runs DURING render → language is set before children render                                                                   
                                                                                                                                          
Now when navigating e.g. to `/en/dashboard/opportunities/7`, only one request will fire with `language=en`.

<img width="754" height="71" alt="image" src="https://github.com/user-attachments/assets/2228a926-cb3b-4fe9-af43-af2347dd1581" />

### DIFF
--- a/src/config/i18next.tsx
+++ b/src/config/i18next.tsx
@@ -6,7 +6,7 @@ import { initReactI18next, I18nextProvider } from "react-i18next";
 import deTranslation from "../../public/locales/de/translations.json";
 import enTranslation from "../../public/locales/en/translations.json";
 import { Env } from "@/types";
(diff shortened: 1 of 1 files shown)


arturasmckwcz (approved):
Yeah certainly makes sense thanks!

PR #212, opened 2026-02-01
Simplify pre-commit

HansKre:
## Description

Even on my M4 max with 64GB RAM, the build is quite slow and creates massive preassure on file I/O which makes my computer freeze. I don't see the value of running the build on every commit after linting and typechecking / compiling succeeded.


arturasmckwcz (approved):
thank you! there's no point of building just for type checking

Issue #213, opened 2026-02-02
Agent card: header and statuses


need4deed:
### Description

https://www.figma.com/design/9Kl17OiDLbWoNxgY43Rgjk/Need4Deed-Designs-2025?node-id=12504-113640&t=BsaUTayszJ75c2XB-4

PR #222, opened 2026-02-07
#213: Implement Agent-ProfileHeader

HansKre:
## Description

Implements #213 

For testing, start `be` and `fe` and navigate to `http://localhost:3000/en/dashboard/agents/6`

image

image

image

image

## Mock getting `/agents/:id`

For testing, add this at the end of `RoutePrefix` to `src/server/types/enums.ts`

```ts
export const RoutePrefix = {
   ...
   VOLUNTEER: "/volunteer",
   + AGENT: "/agent",
} as const;
```

Create `src/server/routes/agent.ts` with following content:

```ts
import { FastifyInstance, FastifyPluginOptions } from "fastify";
import fp from "fastify-plugin";
import { RoutePrefix } from "../types";

async function agentRoutes(
  fastify: FastifyInstance,
  options: FastifyPluginOptions,
) {
  const prefixedPath = options.prefix || RoutePrefix.AGENT;

  fastify.get(`${prefixedPath}/:id`, async (request, reply) => {
    const { id } = request.params as { id: string };
    const now = new Date().toISOString();
    return reply.status(200).send({
      message: "Agent fetched successfully",
      data: {
        id: Number(id),
        name: "Hanger 1-3",
        type: "rac",
        createdAt: now,
        statusEngagement: "new",
        volunteerSearch: "not_needed",
        trustLevel: "unknown",
      },
      count: 1,
    });
  });
}

export default fp(agentRoutes, {
  name: "agent-routes",
});
```

Add this to `function createServer` at `src/server/index.ts`

```ts
await fastifyInstance.register(agentRoutes, {
    prefix: RoutePrefix.AGENT,
  });
```

Navigate to: `http://localhost:3000/en/dashboard/agents/6`


src/components/Dashboard/Profile/sections/ProfileHeader/agent/AgentHeader.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const engagementStatusLabels = createEngagementStatusLabelMap(t);
+  const volunteerSearchLabels = createVolunteerSearchLabelMap(t);
+  const trustLevelLabels = createTrustLevelLabelMap(t);
+
+  return (
+    <HeaderCard
+      testId="agent-header"
+      avatar={
+        <IconContainer data-testid="agent-header-icon">
+          <House size={120} color="var(--color-blue-500)" weight="duotone" />
```
aricionur: this naming convention depraceted.
Could you add `Icon` to the end of the `House`?

src/components/Dashboard/Profile/sections/ProfileHeader/agent/TrustLevelDropdown.tsx
```diff
@@ -0,0 +1,114 @@
...
+  gap: var(--spacing-8);
+  padding: var(--spacing-8) var(--spacing-16);
+  font-size: var(--font-size-lg);
+  color: var(--color-blue-700);
+  cursor: pointer;
+
+  &:hover {
+    background-color: var(--color-grey-50);
+  }
+`;
```
aricionur: Could you move these styled components to the end of the file(after main component).
we decided to move all these styled components after the main component in the file.

src/components/Dashboard/Profile/ProfileController.tsx
```diff
@@ -125,5 +146,9 @@ export const ProfileController = ({ entityId, entityType }: Props) => {
     return <VolunteerProfileController entityId={entityId} />;
   }
 
+  if (entityType === "agent") {
+    return <AgentProfileController entityId={entityId} />;
+  }
+
```
aricionur, line 151: Is it possible to define here a map for components like something below:

```
const controllers = {
"volunteer" : VolunteerProfileController,
"agent": AgentProfileController,
...
}

and in function

return controllers[entityType}
```

by doing that we will be implementing Open-Closed princible, probably you already know:)

Also it would be very nice to use/define an enum for volunteer, agent ..etc for diffirent profile types. so we can use it in diffirent files..

I found something like that `formTypes.ts  `

```
export enum FormType {
  VOLUNTEER = "volunteer",
  OPPORTUNITY = "opportunity",
}
```

Maybe, you can use/extend this enum or creaate new one.

src/components/Dashboard/Profile/sections/ProfileHeader/agent/TrustLevelDropdown.tsx
HansKre: Is this decision documented?

Going to move it to separate file, as this matches the established pattern:

image

src/hooks/useProfileSections.tsx
```diff
@@ -0,0 +1,145 @@
...
+import { VolunteerProfileDocument } from "@/components/Dashboard/Profile/sections/VolunteerProfileDocument";
+import { IconName, ProfileEntityProps } from "@/components/Dashboard/Profile/types/types";
+import { ApiOpportunityGet, ApiVolunteerGet, VolunteerStateTypeType } from "need4deed-sdk";
+import { useRef } from "react";
+import { useTranslation } from "react-i18next";
+
+export const useProfileSections = ({ volunteer, opportunity, agent }: ProfileEntityProps) => {
+  const { t } = useTranslation();
+
+  const contactDetailsRef = useRef<EditableSectionRef>(null);
```
aricionur, line 26: Could you split this hook, if it is possible?
 it will be easier to maintain small hooks.

aricionur (changes requested):

src/components/Dashboard/Profile/ProfileController.tsx
HansKre: It is possible, but would cause cascading re-renders due to unstable reference of `controllers`, so we'd need to memoize `controllers`, which would add abstraction and complexity. I was considering it, but tbh, I think the current implementation is concise and simple enough.

Regarding the new `enum`:

`ProfileEntityType` is already a clearly definied union-type and usage of `enum` is very controversial in TypeScript-community, so I would leave it at that.

src/hooks/useProfileSections.tsx
HansKre: Let's do it in a separate PR, since all subsequent PRs have changes in this.

HansKre: https://github.com/need4deed-org/fe/pull/229

src/components/Dashboard/Profile/sections/ProfileHeader/agent/TrustLevelDropdown.tsx
aricionur: there is no documantation at all for this project at the moment.
but  good catch, we should add a new one for new volunteer developers

HansKre: Was going to setup and install `react-compiler` and with that, start an ADR [here](https://github.com/need4deed-org/fe/pull/231/changes#diff-54c8cc633f097183f9f77ba310b7bb8dfb8404f42b78abb3919b921e12c2f5fb). Might be a starting point.

src/components/Dashboard/Profile/ProfileController.tsx
aricionur: Thanks for the explanation,

Regarding the re-render/memoization concern: if we define the controller map outside the component function , it becomes a static reference. This would avoid any cascading re-renders or the need for useMemo entirely, as the object reference never changes.

It might look something like this:

```

const CONTROLLER_MAP: Record<ProfileEntityType, React.ComponentType<{ entityId: string }>> = {
  volunteer: VolunteerProfileController,
  agent: AgentProfileController,
  opportunity: OpportunityProfileController,
};

export const ProfileController = ({ entityId, entityType }: Props) => {
  const Controller = CONTROLLER_MAP[entityType];
  return <Controller entityId={entityId} />;
};
```


I find this keeps the main component very clean as we add more entity types (OCP), but if you still prefer the explicit `if` statements for a small number of types, I'm happy to move forward with the current version.

aricionur (approved):
I added an extra commnet regarding  `ProfileController`.

LGTM

src/components/Dashboard/Profile/ProfileController.tsx
HansKre: Agree, I also find this very clean: https://github.com/need4deed-org/fe/pull/234

Issue #198, opened 2026-01-24
Opp-card: opportunity section


arturasmckwcz:
### Description

create opportunity details section

PR #227, opened 2026-02-12
#198: Opportunity details

HansKre:
## Description

Implements #198 

image

image


src/components/EditableField/EditableField.tsx
```diff
@@ -59,13 +61,20 @@ const InputWrapper = styled.div<HasError>`
...
     &:focus {
       outline: none;
       border: ${(props) =>
         props.$hasError ? "var(--editableField-border-error)" : "var(--editableField-border-focus)"};
     }
   }
+
+  textarea {
+    resize: vertical;
+    min-height: 80px;
```
aricionur: Could you move this value into global.css?

aricionur (approved):

PR #228, opened 2026-02-13
adds rac contact details

DarrellRoberts:
## Description
Adds Agent/RAC Contact Details

## Related Issues
Closes [issue 219](219)

## Changes
- New hook for updating contact details
- New component for RAC Contact Details

## Screenshots / Demos
image
image

As we don't have "role" and "landline" in the database yet, I had to mock a handler in the backend. Also I had some conflicts with the existing PATCH /person/:id, due to the lack of address on the form. 

With this handler, you can successfully edit the details of: firstName, middleName, lastName, email and mobile.

However, if you want to successfully edit role or landline, you'll get a mock success response, but you won't be able to see the updated role/landline in the form.

`src/server/routes/person.routes.ts`

```typescript
fastify.patch<{
    Params: ParamsId;
    Reply: ReplyData<ApiPersonGet>;
    Body: ApiPersonPatch;
  }>(
    "/:id",
    {
      schema: {
        params: idParamSchema,
        body: { $ref: "ApiPersonPatch#" },
        // response: responseSchema("ApiPersonGet#"),
        response: {
          200: {
            type: "object",
            properties: {
              message: { type: "string" },
              data: personResponseSchema,
            },
            required: ["message", "data"],
          },
        },
      },
    },
    async (request, reply) => {
      const personId = request.params.id;
      const body = request.body as any;

      if (body.role || body.landline) {
        const mockData = {
          id: personId,
          firstName: body.firstName || "Mock",
          lastName: body.lastName || "User", 
          ...body,
          updatedAt: new Date().toISOString(),
          createdAt: new Date().toISOString(), 
        };

        return reply.status(200).send({
          message: "Mock Update Successful",
          data: mockData,
        });
      } else {
        const person = await fastify.db.personRepository.findOne({
          where: { id: personId },
          // relations: ["address.postcode"],
          relations: [],
        });

        if (!person) {
          throw new NotFoundError(`Person with id ${personId} not found.`);
        }

        const updatedPersonObj = deepMerge(
          {
            id: person.id,
            // addressId: person.addressId,
            // address: { postcodeId: person.address.postcodeId },
          },
          dtoParsePerson(request.body),
        );

        const updatedPerson = await updatePerson(updatedPersonObj);

        const serializedData = dtoSerializePerson(updatedPerson);

        const data = {
          ...person,
          ...serializedData,
          address: null,
          createdAt: person.createdAt || new Date().toISOString(),
          updatedAt: new Date().toISOString(),
        };

        return reply
          .status(200)
[...]

PR #230, opened 2026-02-15
Refactor: Extract display and edit modes

HansKre:
## Description

Split components that had inline display/edit mode switching into separate Display and Edit files.
Edit components access form state via useFormContext (matching AccompanyingDetails pattern). Orchestrators wrap children with `FormProvider`. Display components receive pre-computed values as `props`.

### DIFF
--- a/src/components/Dashboard/Profile/sections/Comments/common/CommentDisplay.tsx
+++ b/src/components/Dashboard/Profile/sections/Comments/common/CommentDisplay.tsx
@@ -1,0 +1,45 @@
+import { DotsThreeOutline } from "@phosphor-icons/react";
+import { useTranslation } from "react-i18next";
+import { CommentActionMenu } from "./CommentActionMenu";
+import { CommentText, MenuAction } from "./styles";
+
+type MenuState = {
+  isOpen: boolean;
+  anchorElement: HTMLButtonElement | null;
+  buttonRef: (el: HTMLButtonElement | null) => void;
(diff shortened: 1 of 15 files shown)


src/components/Dashboard/Profile/sections/Comments/common/CommentDisplay.tsx
```diff
@@ -0,0 +1,45 @@
...
+  onDelete: () => void;
+};
+
+type Props = {
+  commentId: string | number;
+  content: string;
+  menu: MenuState;
+};
+
+export function CommentDisplay({ commentId, content, menu }: Props) {
```
DarrellRoberts, line 22: I noticed within the ` /components` dir there are some components that use the `ES5 function call with default`, `ES5 function call without default` and others that use the `ES6 arrow function`. Do you think it'd be worth creating an issue to enforce consistency within the  `components` folder?

DarrellRoberts (commented):
Looks good from my end, just needs to be updated. I've used this pattern for PR #228

src/components/Dashboard/Profile/sections/Comments/common/CommentDisplay.tsx
HansKre: Good idea. Shoud probably agree on using or not using the `default`-export in the first place :-)

arturasmckwcz (approved):

PR #236, opened 2026-02-15
Hide toggle-edit-mode-btn when editing

HansKre:
## Description

Hide toggle-edit-mode-btn when editing.


DarrellRoberts (changes requested):
Looks good! However due to bad timing the Agent Contact Details is now merged, so that also needs to be added

DarrellRoberts (approved):

Issue #216, opened 2026-02-02
Agent-Card: opportunities section


need4deed:
### Description

https://www.figma.com/design/9Kl17OiDLbWoNxgY43Rgjk/Need4Deed-Designs-2025?node-id=11375-22399&t=BsaUTayszJ75c2XB-4

### Proposed solution

This issue is a bit complex but one can adjust the code based on the similar section in the volunteer card (profile)

PR #243, opened 2026-03-03
#216: Implement AgentOpportunities

HansKre:
## Description

Implements https://github.com/need4deed-org/fe/issues/216

URL for testing: http://localhost:3000/en/dashboard/agents/6

image

### DIFF
--- a/src/components/Dashboard/Profile/sections/VolunteerOpportunities/styles.ts
+++ b/src/components/Dashboard/Profile/sections/VolunteerOpportunities/styles.ts
@@ -1,0 +1,16 @@
+import styled from "styled-components";
+
+export const CountBadge = styled.span`
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  min-width: var(--volunteer-profile-opportunities-tab-count-badge-min-width);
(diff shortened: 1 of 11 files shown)


src/components/Dashboard/Profile/sections/VolunteerOpportunities/styles.ts
DarrellRoberts, line 1: would we use this file to store the other styled components within this directory? e.g. on /Tabs.tsx and /VolunteerOpportunities.tsx

DarrellRoberts (changes requested):

src/components/Dashboard/Profile/sections/VolunteerOpportunities/styles.ts
HansKre: Sure. However, I would see proactively moving pre-existing styles optional for the scope of this PR.

DarrellRoberts: good point, then it looks good to me

DarrellRoberts (approved):

PR #262, opened 2026-03-12
#247: agent list header

### LINKED ISSUE #247: agent list header
### Description

{}

### Proposed solution

_No response_

DarrellRoberts:
## Description
Adds header, searchbar and filters to Agent lists

Test on [http://localhost:3000/en/dashboard/agents](http://localhost:3000/en/dashboard/agents
)
## Related Issues
Closes #247 

## Changes

## Screenshots / Demos
image

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/src/components/Dashboard/Agents/Agents.tsx
+++ b/src/components/Dashboard/Agents/Agents.tsx
@@ -49,7 +49,7 @@ export const Agents = () => {
   };
 
   const handleClearAllFilters = () => {
-    const cleared = getClearFilter(cardsFilter) as unknown as AgentCardsFilter;
+    const cleared = getClearFilter<AgentCardsFilter>(cardsFilter);
     setCardsFilter(cleared);
     router.push(pathname + questionMark + serializeAgentFilters(cleared, searchParams));
   };
--- a/.env.sample
+++ b/.env.sample
@@ -1,3 +1,4 @@
 NODE_ENV=development
 URL_API="http://localhost:5000"
 NEXT_PUBLIC_CLOUDFRONT_URL=https://<user_id>.cloudfront.net/images
+NEXT_PUBLIC_CLOUDFRONT_DATA_URL=https://<user_id>.cloudfront.net/data
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "cSpell.words": ["Organisation"]
+}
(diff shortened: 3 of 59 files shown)


src/components/Dashboard/Agents/Agents.tsx
```diff
@@ -2,11 +2,96 @@
...
+  };
+
+  const handleClearAllFilters = () => {
+    const cleared = getClearFilter(cardsFilter) as unknown as AgentCardsFilter;
+    setCardsFilter(cleared);
+    router.push(pathname + questionMark + serializeAgentFilters(cleared, searchParams));
+  };
+
+  useEffect(() => {
+    if (!apiFilterOptions) return;
```
HansKre, line 58: A bit unfortunate we have to do this in a `useEffect`. Better way would be a `onSuccess`-handler in `useGetQuery` (just a note, not expecting any changes here)

src/components/Dashboard/Agents/Agents.tsx
```diff
@@ -2,11 +2,96 @@
...
+  };
+
+  const handleFilterUpdate = (newFilter: AgentCardsFilter | ((prev: AgentCardsFilter) => AgentCardsFilter)) => {
+    const updatedFilter = typeof newFilter === "function" ? newFilter(cardsFilter) : newFilter;
+    setCardsFilter(updatedFilter);
+    router.push(pathname + questionMark + serializeAgentFilters(updatedFilter, searchParams));
+  };
+
+  const handleClearAllFilters = () => {
+    const cleared = getClearFilter(cardsFilter) as unknown as AgentCardsFilter;
```
HansKre: Let`s avoid `as unknown as` as it defeats the purpose of TypeScript

HansKre (dismissed):
Looks good, except for the unchecked cast :-)

src/components/Dashboard/Agents/Agents.tsx
DarrellRoberts: fair point, I'll leave it for now

src/components/Dashboard/Agents/Agents.tsx
DarrellRoberts: I understand and makes sense, but I fell into a bit of a rabbit hole trying to fix the ts error. I've removed it and modified the `getClearFilter` to include a generic to satisfy the error. What do you think: 
````jsx
export const getClearFilter = <T>(filter: object): T => {
  const newFilter: Record<string, string | boolean | object> = {};

  for (const [key, val] of Object.entries(filter)) {
    if (typeof val === "boolean") newFilter[key] = false;
    else if (typeof val === "string") newFilter[key] = "";
    else if (typeof val === "object") newFilter[key] = getClearFilter(val);
    else throw new Error("Unsupported type to clear the filter");
  }

  return newFilter as T;
};
````

arturasmckwcz (approved):
thank you!

PR #264, opened 2026-03-17
#248: Agent fetching hook

### LINKED ISSUE #248: agent fetching hook
### Description

{}

### Proposed solution

_No response_

DarrellRoberts:
## Description
Adds Agent List fetching hook 

## Related Issues
Closes #248 

## Changes

- I had to tweak the `useGetQuery hook` to make it possible to opt out of `language` being included in the request url (otherwise it will trigger a server error).

## Screenshots / Demos
image

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/src/components/Dashboard/Agents/AgentListController.tsx
+++ b/src/components/Dashboard/Agents/AgentListController.tsx
@@ -22,10 +22,12 @@ type Props = {
 export const AgentListController = ({ setNumOfAgents, sortOrder, isFiltersOpen, filter, apiFilterOptions }: Props) => {
   const [currentPage, setCurrentPage] = useState(1);
 
-  const serializedFilter = serializeAgentFilters(filter, undefined, false, {
-    serializeToIDs: true,
-    apiFilterOptions,
-  }) as URLSearchParams;
+  const serializedFilter = new URLSearchParams(
+    serializeAgentFilters(filter, undefined, false, {
+      serializeToIDs: true,
+      apiFilterOptions,
+    }),
+  );
 
   const { data, count } = useGetQuery<ApiAgentGetList[]>({
     queryKey: ["agents"],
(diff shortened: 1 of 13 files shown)


src/components/Dashboard/Agents/AgentListController.tsx
```diff
@@ -1,103 +1,57 @@
...
-          de: "Mitte",
-        },
-      },
-      comments: [],
-      agentDetails: {},
-      type: "",
+  const serializedFilter = serializeAgentFilters(filter, undefined, false, {
+    serializeToIDs: true,
+    apiFilterOptions,
+  }) as URLSearchParams;
```
HansKre: Could we avoid the cast?

HansKre (dismissed):
Almost flawless 😆 👍

src/components/Dashboard/Agents/AgentListController.tsx
DarrellRoberts: thanks so much for reviewing! Without the casting, it flags errors on the `useGetQuery` hook. So in my latest commit in order to satisfy the `FilterParam` type for `filter` on the `useGetQuery`, I wrapped the `serializedFilter` function in a `URLSearchParams` rather than casting it. What do you think?
````jsx
  const serializedFilter = new URLSearchParams(
    serializeAgentFilters(filter, undefined, false, {
      serializeToIDs: true,
      apiFilterOptions,
    }),
  );
````

arturasmckwcz (approved):
thank you!

src/components/Dashboard/Agents/AgentListController.tsx
arturasmckwcz: @DarrellRoberts
it's definitely better, much better than casting which is always sweeping under the rug to some extent

DarrellRoberts: thanks @arturasmckwcz !

Issue #265, opened 2026-03-17
Agent card: trust level fix


need4deed:
### Summary

In agent cards (profile) trust level it does not work and needs to be renamed to "responsiveness" 

### Steps to Reproduce

When changing trust level it likely does not send to api (sdk) and BE or apply the selection in the FE

<img width="601" height="443" alt="Image" src="https://github.com/user-attachments/assets/0ecada1b-8fde-489e-b164-a1de97da660c" />

### Expected Behavior

to be able to change between the three options (unknown, high, low) 

### Actual Behavior

stays at one opion,  drop down opens but no section in effect

need4deed:
its a bug + a feature request - change the text there to responsiveness instead of trust level

117Isabell:
Hey I'd like to work on this. Thanks

arturasmckwcz:
> Hey I'd like to work on this. Thanks

Assigned. Thank a lot for helping us!

Cy-fox:
hello

PR #273, opened 2026-03-27
Fix agent card trust level

117Isabell:
## Description
Add trust level row to AgentCard with TrustLevelDropdown and useUpdateAgentStatus; block card navigation when interacting with the control.

## Related Issues
issue: https://github.com/need4deed-org/fe/issues/265#issuecomment-4126657116 

## Changes

## Screenshots / Demos
now:
image

before: no Responsiveness field

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "dependencies": {
     "@hookform/resolvers": "^5.2.2",
     "@phosphor-icons/react": "^2.1.10",
-    "@tanstack/react-form": "^0.26.4",
+    "@tanstack/react-form": "^1.28.5",
     "@tanstack/react-query": "^5.80.7",
     "axios": "^1.11.0",
     "classnames": "^2.5.1",
@@ -22,7 +22,7 @@
     "email-validator": "^2.0.4",
     "i18next": "^25.3.2",
     "need4deed-sdk": "^0.0.66",
-    "next": "15.3.8",
+    "next": "^15.5.14",
     "react": "^19.0.0",
     "react-day-picker": "^9.13.0",
     "react-dom": "^19.0.0",
@@ -45,7 +45,8 @@
     "husky": "^9.1.7",
     "lint-staged": "^16.2.7",
     "prettier": "^3.8.1",
-    "typescript": "^5"
+    "typescript": "^5",
+    "yarn": "^1.22.22"
   },
   "lint-staged": {
     "*": "prettier --ignore-unknown --write"
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "dependencies": {
     "@hookform/resolvers": "^5.2.2",
     "@phosphor-icons/react": "^2.1.10",
-    "@tanstack/react-form": "^1.28.5",
+    "@tanstack/react-form": "^0.26.4",
     "@tanstack/react-query": "^5.80.7",
     "axios": "^1.11.0",
     "classnames": "^2.5.1",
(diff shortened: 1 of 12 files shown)


package.json
jagdishlove, line 1: Hi, these file changes might reflect the whole code, so it would be suggested to have the same version to use them until we have some issues or bugs in the package's current version.
Could you revert, and only we are changing the need4deed-sdk version, which will be provided by @arturasmckwcz on time to time.

yarn.lock
jagdishlove, line 1: Same for this file until we change the version of need4deed-sdk, we don't need to make any changes and commit to these files.
Could you also revert this file?

src/components/Dashboard/Agents/AgentCard.tsx
```diff
@@ -48,6 +58,17 @@ export const AgentCard = ({ agent }: Props) => {
         <Heading4>{t("dashboard.agentProfile.volunteerSearch")}</Heading4>
         <StatusBadge status={agent?.volunteerSearch} label={volunteerSearchLabels[volunteerSearch]} />
       </CardDetailsInfo>
+      <CardDetailsInfo onClick={(e) => e.stopPropagation()} onKeyDown={(e) => e.stopPropagation()} role="presentation">
```
jagdishlove, line 66: Why are we using event propagation here?

src/components/Dashboard/Profile/sections/ProfileHeader/agent/AgentHeader.tsx
```diff
@@ -62,7 +64,7 @@ export const AgentHeader = ({ agent }: Props) => {
             value={agent.trustLevel}
             options={TRUST_LEVEL_OPTIONS}
             labels={trustLevelLabels}
-            onChange={() => {}}
+            onChange={(next) => patchAgent({ trustLevel: next })}
```
jagdishlove: This isn't a major, but I recommend adding a function call here so that if more things come up, they can be added to that function structurally.

src/components/Dashboard/common/statusMaps.ts
```diff
@@ -5,12 +5,15 @@ import {
...
   SparkleIcon,
   StopCircleIcon,
 } from "@phosphor-icons/react";
-import { AgentEngagementStatusType, AgentVolunteerSearchType } from "need4deed-sdk";
+import { AgentEngagementStatusType, AgentTrustType, AgentVolunteerSearchType } from "need4deed-sdk";
 
-export type StatusValue = AgentEngagementStatusType | AgentVolunteerSearchType;
+export type StatusValue = AgentEngagementStatusType | AgentVolunteerSearchType | AgentTrustType;
 
 export const statusColorMap: Record<string, string> = {
```
jagdishlove: Use Type Safety to avoid any random string input here.


```export const statusColorMap: Record<StatusValue, string>```

src/components/Dashboard/common/statusMaps.ts
jagdishlove: Here also.
use ```StatusValue``` for Type safety

jagdishlove (changes requested):
Thanks, the code looks good and robust, a few issues and changes are needed.
Please check or reply to the review comments if anything is needed.

jagdishlove (approved):
All good, Thanks @117Isabell  for the collaboration.

PR #319, opened 2026-04-15
fix: set English as default language and persist language choice (#303)

### LINKED ISSUE #303: Default languege english
### Description

make the default language on the app.need4deed english 

### Proposed solution

_No response_

Cy-fox:
## Description
The app was always defaulting to German on every reload or navigation, 
even after the user had switched to English. This fixes both the default 
language and the persistence of the user's language choice.

## Related Issues
Closes #303

## Changes
- Changed DEFAULT_LOCALE from DE to EN in middleware.ts
- Added cookie (preferred-lang) to save the user's language choice for 1 year
- Middleware now reads the cookie first before falling back to the default,
  so bare URLs like /dashboard respect the user's chosen language


DarrellRoberts (commented):
works for me and looks good! Just need to update your branch

need4deed (approved):
Cookie persistence logic is correct. Minor nit: `const redirectResponse = NextResponse.redirect(url); return redirectResponse;` can be simplified to `return NextResponse.redirect(url);` — the intermediate variable adds nothing. Otherwise looks good.

PR #326, opened 2026-04-15
fix: prevent left nav bar from overlapping dashboard content

### LINKED ISSUE #324: left menu blocking the dashboard view
### Summary

left bar menu is blocking search bar and list on the dashboard

### Steps to Reproduce

Image

### Expected Behavior

should not block anything 

### Actual Behavior

in certain resolutions left menu is blocking the other parts pf the dashboard 

### Environment

_No response_

Cy-fox:
## Summary

- Fixes the fixed left navigation bar overlapping the search bar and list area on the dashboard at certain viewport widths
- Uses CSS `max()` to set `margin-left` to whichever is larger: the sidebar width or the centred layout offset, so content is always pushed clear of the nav bar

Closes #324


need4deed:
Victor this PR is picking up all the middleware changes from #319 (default language + cookie persistence) because it looks like it was branched off that branch instead of main.

The actual fix here (the `container.tsx` margin change) is clean and good to go — it just needs to be rebased onto main so the two PRs stay separate. Once you do that this should be a very quick merge.

nadavosa (approved):
CSS fix is solid — using `max()` to clamp the left margin against the nav bar width is the right approach and avoids media query breakpoints. The SDK type fixes are the same as in #329. One minor note: there's a stray space in `var( --color-red-50)` in `statusColorMap` (carried over from before this PR), worth cleaning up at some point. LGTM.

PR #329, opened 2026-04-16
hotfix: fix TypeScript errors introduced by SDK .73 bump

Cy-fox:
Several types were renamed or restructured in the SDK bump that merged with develop, leaving the codebase in a broken state where tsc --noEmit fails on every branch:

## Description
Fix TypeScript errors introduced by the SDK .73 bump that were blocking all branches from committing.

## Changes
- AgentCard: use volunteerSearch from normalized agent, not raw ApiAgentGetList
- helpers: add fallback for volunteerSearch being undefined
- statusMaps: switch to Object.fromEntries to avoid duplicate-key errors from enums sharing string values
- AccompanyingDetails helpers: convert languageToTranslate from number to string
- AgentContactDetails: representative → representatives[0]
- useUpdateAgentContact / useUpdateOrganizationDetails: ApiRepresentativePatch → ApiRepresentativeGet


src/components/Dashboard/Profile/common/statusMaps.ts
```diff
@@ -37,74 +37,78 @@ export type StatusValue =
...
-  [AgentVolunteerSearch.SEARCHING]: "var(--color-red-50)",
-  [AgentTrustLevel.UNKNOWN]: "var(--color-grey-50)",
-  [AgentTrustLevel.LOW]: "var(--color-red-50)",
-  [AgentTrustLevel.HIGH]: "var(--color-green-100)",
-  [AgentEngagementStatusType.NEW]: "var(--color-violet-100)",
-  [AgentEngagementStatusType.ACTIVE]: "var(--color-green-100)",
-  [AgentEngagementStatusType.INACTIVE]: "var(--color-grey-50)",
-  [AgentEngagementStatusType.UNRESPONSIVE]: "var(--color-grey-50)",
-};
+// Several SDK enums share the same underlying string values (e.g. "new", "active",
```
arturasmckwcz, line 40: great move! thank you

arturasmckwcz (approved):

src/components/Dashboard/Agents/helpers.ts
```diff
@@ -29,7 +29,7 @@ export function getNormalizedAgent(agent: AgentListItem): Omit<
     ...agent,
     type: agent.type,
     district: agent.district,
-    volunteerSearch: agent.volunteerSearch,
+    volunteerSearch: agent.volunteerSearch ?? AgentVolunteerSearchType.NOT_NEEDED,
```
arturasmckwcz, line 32: okay it doesn't throw a `TypeError`, but still it renders dash instead `not needed`. it's out of the scope I understand, just pointing out

src/hooks/useUpdateOrganizationDetails.ts
```diff
@@ -1,11 +1,11 @@
 import { ApiAgentProfileGet } from "@/components/Dashboard/Profile/types";
 import { apiPathOrganization } from "@/config/constants";
 import { useMutationQuery } from "@/hooks";
-import { ApiRepresentativePatch } from "need4deed-sdk";
+import { ApiRepresentativeGet } from "need4deed-sdk";
 import { DeepPartial } from "ts-type-safe";
 
 export const useUpdateOrganization = (organizationId: string) => {
-  return useMutationQuery<DeepPartial<ApiRepresentativePatch>, ApiAgentProfileGet>({
+  return useMutationQuery<DeepPartial<ApiRepresentativeGet>, ApiAgentProfileGet>({
```
arturasmckwcz, line 8: what's the point of this refactoring?
if you'd check SDK:
```
export type ApiRepresentativePatch = Partial<ApiRepresentativeGet>;
```

arturasmckwcz: `ApiRepresentativePatch` is introduced here just in case the type for PATCH starts differing from the one for GET. now it'd imply refactoring a lot of locations instead of updating SDK IMHO.

src/components/Dashboard/Agents/helpers.ts
Cy-fox: @arturasmckwcz The new version of the SDK (.73) changed how this works, it can now be empty. I added a "default" value here so the app doesn't crash when that happens.

src/hooks/useUpdateOrganizationDetails.ts
Cy-fox: @arturasmckwcz Oh, I see it now! I had trouble finding the ApiRepresentativePatch type after the update, so I swapped it just to get the build passing. I'll change it back to keep it future-proof. Thanks for showing me!

nadavosa (approved):
Clean hotfix. The `Object.fromEntries` approach for `statusColorMap` and `statusIconMap` is the right call — duplicate computed keys in object literals are a TS error with strict SDK enums. The `representative` → `representatives[0]` and `ApiRepresentativePatch` → `ApiRepresentativeGet` changes align with the SDK bump. LGTM.

src/hooks/useUpdateOrganizationDetails.ts
Cy-fox: @arturasmckwcz `ApiRepresentativePatch` was removed in SDK .73 , that's what caused the TS errors. Reverting would break the build again.

nadavosa:
Before this gets merged, please run `yarn lint` and `yarn typecheck` locally and fix any errors. We've added a CI check (#359) that will block merging if these fail.

PR #334, opened 2026-04-17
fix: prevent redirect to login on public pages

### LINKED ISSUE #328: From phone: Become a volunteer redirects to login page to the app
### Summary

Become a volunteer redirects to login page to the app when using phone

### Steps to Reproduce

click become a volunteer from website using phone view

### Expected Behavior

https://app.need4deed.org/en/forms/volunteer should be the regular become a volunteer form with no login

### Actual Behavior

https://app.need4deed.org/en/forms/volunteer send user to login 

Image

### Environment

_No response_

ivannissimrch:
## Description
- Visiting volunteer Form while not logged in was redirecting to login
- Root cause: Header always called useCurrentUser() firing GET /api/me, which returned 401, triggering token refresh, which failed and redirected
- Fix: add isPublicPage prop, flowing from PageLayout to  Header to  useCurrentUser(enabled) so no request fires on public pages at all

## Related Issues
Closes #328 

## Screenshots / Demos
[fix-redirect.webm](https://github.com/user-attachments/assets/97530f06-879e-488a-be05-2c37797638bb)

## Test
 - [ ] Visit /en/forms/volunteer while not logged in,  should load normally, no redirect
 - [ ] Visit /en/login while not logged in, should load normally
 - [ ] Visit /en/dashboard while not logged in,  should still redirect to login
 - [ ] Visit /en/dashboard while logged in, user profile shows in header

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ next-env.d.ts
 # dev
 dev/
 dev.*
+.env.local
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "date-fns": "^4.1.0",
     "email-validator": "^2.0.4",
     "i18next": "^25.3.2",
-    "need4deed-sdk": "^0.0.73",
+    "need4deed-sdk": "^0.0.75",
     "next": "15.3.8",
     "react": "^19.0.0",
     "react-day-picker": "^9.13.0",
--- a/src/app/[lang]/forms/volunteer/page.tsx
+++ b/src/app/[lang]/forms/volunteer/page.tsx
@@ -4,7 +4,7 @@ import { PageLayout } from "@/components/Layout";
 
 export default function VolunteerPage() {
   return (
-    <PageLayout background="var(--color-white)">
+    <PageLayout background="var(--color-white)" isPublicPage>
       <BecomeVolunteer />;
     </PageLayout>
   );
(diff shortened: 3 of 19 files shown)


DarrellRoberts (commented):
Works for me and fixes the issue. 

However, I would favour using a `cookie` approach rather than adding a prop to every page layout. That way you would change only `3 files` compared to `12 files`. 

So:  

- on the `/Login/LoginController` you set a boolean cookie e.g.`is_logged_in` cookie on the `onLoginSuccess` func, 
- then on the `useCurrentUser hook` you get the cookie, ideally with a `helper function` you create on `/utils/helpers` e.g. `getCookie()` 
- On the  `useCurrentUser hook` you do `enabled: hasAuthHint (cookie condition) && enabled`

That way, the `useCurrentUser` hook on the `Header` component will only fire when this `cookie` is present. It also prevents potential bugs when `<PageLayout>` is added in the future and the `isPublicPage` prop is incorrectly defined or missing

What do you think?

ivannissimrch (commented):
Thank you for the review and changes.

DarrellRoberts (approved):

arturasmckwcz (approved):
thank you guys!!!

PR #350, opened 2026-04-20
fix: send ISO code instead of translated title for language submission

### LINKED ISSUE #347: Any language for new volunteers is Ghotuo
### Summary

Forms are filled out by volunteers as having various languages but Ghotuo is registered in the dashboard 

### Steps to Reproduce

Forms filled out in vaious langugges on app.need4deed/forms/volunteer

Image

Any language for new volunteers is Ghotuo

Image

### Expected Behavior

Reflect the real language

### Actual Behavior

The language of the new volunteers is always Ghouto

Likely root cause: parseFormStateDTOVolunteer in the FE sends l.language which holds the translated display title (e.g. "Arabisch") instead of the
  ISO code. The BE's getLanguageTitle then does findOneBy({ isoCode: "Arabisch" }), finds nothing, returns undefined, and TypeORM ignores undefined in 
  the WHERE clause — returning the first row in the Language table (Ghotuo).      

 

### Suggested fix: 
In LanguageFieldRow.tsx, change the select option value from the translated title to the item ID (which is the ISO code in the
  fallback list):                                                                                                                                      
   
  // before                                                                                                                                            
  value={item.title[i18n.language as Lang]}

  // after
  value={String(item.id)}
                                                                                                                                                       
  This makes l.language store e.g. "ar" instead of "Arabisch", which is what getLanguageTitle expects. The same change also fixes the broken           
  de-duplication logic in disabledLanguages, which already compares against String(item.id).

_No response_

nadavosa:
## Summary
- Fixes #347
- Language select options were storing the translated display title (e.g. `"Arabisch"`) as the value instead of the ISO code (e.g. `"ar"`)
- The BE's `getLanguageTitle` expects an ISO code, so all lookups returned `undefined`, falling back to the first DB row (Ghotuo)

## Change
One line in `LanguageFieldRow.tsx`:
```tsx
// before
value={item.title[i18n.language as Lang]}

// after
value={String(item.id)}
```

## Test plan
- [ ] Fill out volunteer form, select any language (e.g. Arabic)
- [ ] Submit and verify the correct language appears in the dashboard

### DIFF
--- a/src/components/forms/LanguageFields/LanguageFieldRow.tsx
+++ b/src/components/forms/LanguageFields/LanguageFieldRow.tsx
@@ -58,16 +58,13 @@ export function LanguageFieldRow({
             {t("form.becomeVolunteer.fields.languages.chooseLanguage")}
           </option>
           <option key={language.id} value={language.language}>
-            {language.language}
+            {availableLanguages.find((item) => String(item.id) === language.language)?.title[i18n.language as Lang] ?? language.language}
           </option>
           {availableLanguages.map((item) => (
             <option
               key={item.id}
               value={String(item.id)}
-              disabled={
-                (disabledLanguages.includes(String(item.id)) && language.language !== String(item.id)) ||
-                disabledLanguages.includes(item.title[i18n.language as Lang] as string)
-              }
+              disabled={disabledLanguages.includes(String(item.id)) && language.language !== String(item.id)}
             >
               {item.title[i18n.language as Lang]}
             </option>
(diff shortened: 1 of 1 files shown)


lourooo:
The fix makes sense, but there are two things worth checking:

1. Selected language shows the ID instead of the name (lines 60–61)
After this change, `language.language` will store the ID (e.g. `"5"`) instead of the translated title. Since lines 60–61 use `{language.language}` as both the value and the display text, the currently selected language in the dropdown will show the raw ID instead of the language name.

2. Line 69 becomes dead code
Since `disabledLanguages` will now contain IDs, the condition `disabledLanguages.includes(item.title[i18n.language as Lang] as string)` will never be true and can be removed.

(Victor also checked)

PR #359, opened 2026-04-21
ci: add lint and typecheck workflow on PRs to develop

nadavosa:
## Summary
- Adds a GitHub Actions workflow that runs on every PR targeting `develop`
- Runs `yarn lint` and `yarn typecheck` to catch errors before merge
- Uses yarn and Next.js build cache for faster runs

## Test plan
- [ ] Open a PR to `develop` and verify the workflow triggers
- [ ] Introduce a lint error and verify the check fails

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #360, opened 2026-04-21
fix: add name field to opportunity details for renaming

### LINKED ISSUE #340: Opportunity profile - Rename opportunity
### Description

It’s not possible to rename an opportunity (e.g. when there’s a typo)
We should be able to rename opportunities

Image

### Proposed solution

Image

Add row called name

Cy-fox:
## Description
Adds an editable "Name" field to the Opportunity Details section so coordinators can rename an opportunity directly from the profile page. The field is shown in both display and edit mode, with validation and i18n support (EN/DE).
## Related Issues
Closes #340
## Changes
- Add `useUpdateOpportunityTitle` hook (PATCH title field)
- Add `title` to `opportunityDetailsSchema` with required validation
- Add "Name" field to `OpportunityDetailsEdit` (first field, wired to mutation)
- Add "Name" field to `OpportunityDetailsDisplay`
- Add i18n keys `opportunityName`, `saveSuccess`, `validation.opportunityNameRequired` (EN + DE)
- Fix pre-existing TS error in `AgentContactDetails` (`representative` → `representatives[0]`)


nadavosa:
Before this gets merged, please run `yarn lint` and `yarn typecheck` locally and fix any errors. We've added a CI check (#359) that will block merging if these fail.

nadavosa (commented):
## Review

Good addition — clean hook, schema, and i18n keys.

**Issue — form closes before mutation completes:**

```ts
const onSubmit = (data: OpportunityDetailsFormData) => {
  if (data.title !== opp.title) {
    updateTitle({ title: data.title });
  }
  onCancel(); // fires immediately, regardless of success/failure
};
```

`onCancel()` is called right away, so the form closes while the PATCH is still in flight. If the update fails, the user sees no error. The pattern used elsewhere is to pass callbacks:

```ts
const onSubmit = (data: OpportunityDetailsFormData) => {
  if (data.title !== opp.title) {
    updateTitle({ title: data.title }, { onSuccess: onCancel });
  } else {
    onCancel();
  }
};
```

**Minor — `setValue={() => {}}` on display field:**

```tsx
<EditableField mode="display" ... setValue={() => {}} />
```

If `EditableField` accepts `setValue` as required even in display mode, a no-op is fine — but worth checking if there's a `readOnly` or display-only variant that doesn't need it.

nadavosa:
The fix is one line — move `onCancel()` into the `onSuccess` callback so the form only closes after the PATCH succeeds:

```ts
const onSubmit = (data: OpportunityDetailsFormData) => {
  if (data.title !== opp.title) {
    updateTitle({ title: data.title }, { onSuccess: onCancel });
  } else {
    onCancel();
  }
};
```

Otherwise looks good to go.

arturasmckwcz (approved):
thank you!

Issue #352, opened 2026-04-20
Languages end up being in the “Main communication field”


Cy-fox:
### Summary

Languages end up being in the “Main communication field”

### Steps to Reproduce

All the languages end up being in the “Main communication field”

### Expected Behavior

Main communication” should only contain the languages the person speaks. Language to translate should be only shown in the “Language to translate

### Actual Behavior

<img width="1576" height="671" alt="Image" src="https://github.com/user-attachments/assets/5fce40ca-8cc5-4fb4-b8d3-c6aa9df839f2" />

need4deed:
@Cy-fox this reffers to opportunity not volunteer cards
@DarrellRoberts thanks for checking this!

PR #362, opened 2026-04-21
fix: split volunteer languages into main communication and translate fields

Cy-fox:
## Description
Volunteer languages were all rendering in a single "Languages" field. Split into two separate rows "Main communication" (LangPurpose.GENERAL) and "Language to translate" (LangPurpose.TRANSLATION), matching the expected UI.
## Changes
- Add optional `purpose` filter to `formatLanguagesForDisplay`
- Replace single `languages` prop in `DisplayFields` with `mainCommunication` + `languagesToTranslate`
- Pass both filtered language strings from `VolunteerProfile`
- Add i18n keys `mainCommunication` / `languagesToTranslate` (EN + DE)

### DIFF
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -590,7 +590,8 @@
       "profileSection": {
         "title": "Freiwilligenprofil",
         "edit": "Bearbeiten",
-        "languages": "Sprachen",
+        "mainCommunication": "Hauptkommunikation",
+        "languagesToTranslate": "Sprache zum Übersetzen",
         "availability": "Verfügbarkeit",
         "districts": "Bezirke",
         "volunteerType": "Freiwilligen-Typ",
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -590,6 +590,7 @@
       "profileSection": {
         "title": "Freiwilligenprofil",
         "edit": "Bearbeiten",
+        "languages": "Sprachen",
         "mainCommunication": "Hauptkommunikation",
         "languagesToTranslate": "Sprache zum Übersetzen",
         "availability": "Verfügbarkeit",
--- a/public/locales/en/translations.json
+++ b/public/locales/en/translations.json
@@ -589,6 +589,7 @@
       "profileSection": {
         "title": "Volunteer profile",
         "edit": "Edit",
+        "languages": "Languages",
         "mainCommunication": "Main communication",
         "languagesToTranslate": "Language to translate",
         "availability": "Availability",
--- a/public/locales/en/translations.json
+++ b/public/locales/en/translations.json
@@ -589,7 +589,8 @@
       "profileSection": {
         "title": "Volunteer profile",
         "edit": "Edit",
-        "languages": "Languages",
+        "mainCommunication": "Main communication",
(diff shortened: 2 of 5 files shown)


public/locales/de/translations.json
```diff
@@ -590,7 +590,8 @@
       "profileSection": {
         "title": "Freiwilligenprofil",
         "edit": "Bearbeiten",
-        "languages": "Sprachen",
```
DarrellRoberts, line 593: I wouldn't remove the key, far safer to leave it and add what you need to add

public/locales/en/translations.json
```diff
@@ -589,7 +589,8 @@
       "profileSection": {
         "title": "Volunteer profile",
         "edit": "Edit",
-        "languages": "Languages",
```
DarrellRoberts, line 592: same here

DarrellRoberts (changes requested):
hey, good work although I'm a bit confused both on what this PR is trying to achieve and the Issue itself.

e.g. Issue #352 shows the `Opportunity` profile in the screenshot but here we're editing the `Volunteer` profile. 

I don't think `volunteer` has a `languagesToTranslate` property and it exists on `opportunity`.

I also looked at different Opportunities and couldn't see an issue. 

So if you have the `id` of an Opportunity Profile where you're seeing the issue, could you please share it? e.g. `/dashboard/opportunities/[id]`

I may be misunderstanding but just want to be clear on what the issue is

public/locales/en/translations.json
Cy-fox: Hi, thanks for the review.

To clarify: the volunteer object has a languages array where each entry has a purpose field (`LangPurpose.GENERAL` or `LangPurpose.TRANSLATION`). This PR splits the single "Languages" display field into two rows by filtering that array by purpose, so the volunteer profile shows "Main communication" and "Language to translate" separately instead of mixing them. There is no new `languagesToTranslate` property on the volunteer, `languagesToTranslate` is just the label key name in the translations file.

Issue #352 was filed with an opportunity profile screenshot by mistake, but the actual bug is on the volunteer profile display, the purpose field was being ignored and all languages were shown in one block.

Good point on the translation key, just pushed a fix that restores the languages key alongside the two new ones.

DarrellRoberts: ok thanks for clarifying. Do you mind giving me a route for me to test whereby we can see languages in the two fields? e.g. `/dashboard/volunteers/[id]` , along with a screenshot?

For me, if I go to for instance `/dashboard/volunteers/50` , it returns empty fields as its languages has no `purpose` property but could just be because I haven't updated my `be`

Cy-fox: Hey, I can't provide a screenshot right now as my test account was removed from the database. As you suspected, the empty fields are likely because your local BE hasn't been updated yet. Once you pull the latest BE changes the purpose field should be populated on volunteer languages and both fields should display correctly. Any volunteer created or edited after that update should work for testing.

DarrellRoberts:
closing this as the issue doesn't relate to volunteers as mentioned [here](https://github.com/need4deed-org/fe/issues/352)

PR #376, opened 2026-04-22
fix: preserve language purpose when saving volunteer profile

### LINKED ISSUE #367: bug: cannot edit languages in volunteer profile
## Description
When a coordinator opens a volunteer's profile and tries to edit the languages section, the changes cannot be saved. The language fields in the volunteer profile edit form are not updating correctly.

## Steps to Reproduce
1. Go to Dashboard → Volunteers
2. Open any volunteer profile
3. Click Edit on the volunteer profile section
4. Try to change/add a language
5. Click Save Changes

## Expected Behavior
The coordinator should be able to update the volunteer's languages and the changes should persist.

## Actual Behavior
Language changes cannot be saved. The form either does not submit or reverts to the previous values.

## Priority
High: affects coordinators' ability to keep volunteer data up to date.

Cy-fox:
Closes #367

## Description
When saving a volunteer's language section, the purpose field (GENERAL, TRANSLATION, RECIPIENT) was being stripped from every language entry. This meant that after saving, the languages would lose their purpose and stop appearing in the right place on the profile.

## Changes
The fix makes the form preserve the existing purpose throughout the data flow, from loading the volunteer data into the form all the way to the API payload on submit. Languages added from scratch in the edit form default to GENERAL.

### DIFF
--- a/src/components/Dashboard/Profile/sections/VolunteerProfile/formatters.ts
+++ b/src/components/Dashboard/Profile/sections/VolunteerProfile/formatters.ts
@@ -31,6 +31,7 @@ export function formatLanguages(
       id: lang.id,
       language: String(dbId),
       level: proficiencyToLevel[lang.proficiency?.toLowerCase() || "native"] || LanguageLevel.NATIVE,
+      purpose: lang.purpose,
     };
   });
 }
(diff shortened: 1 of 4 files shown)


nadavosa (commented):
## Review

Correct fix — `purpose` is now preserved in both directions (load via `formatters.ts`, save via `transformers.ts`). The `?? LangPurpose.GENERAL` fallback is sensible.

**Minor nit — inline import in `types.ts`:**

```ts
purpose?: import("need4deed-sdk").LangPurpose;
```

`LangPurpose` is already imported at the top of `volunteerProfileSchema.ts` and `transformers.ts` in this PR. Cleaner to add a top-level import in `types.ts` too rather than using the inline form — but this is cosmetic, not a blocker.

Otherwise LGTM.

PR #377, opened 2026-04-22
fix: replace district free-text with dropdown in RAC section

### LINKED ISSUE #357: Agent profile District field should be dropdown instead of free-text input
### Description

Implement a District dropdown field in the Agent profile form that allows users to select from predefined districts and saves the selection to the database.

Image

### Proposed solution

_No response_

Cy-fox:
Closes #357

## Description
The district field in the Refugee Accommodation Centre edit form was a plain text input, which made it impossible to consistently pick a valid district. On top of that, the value was not being sent to the API at all when saving, only name and address were submitted.

## Changes
The field is now a dropdown fed from the districts API, following the same pattern used in the volunteer profile. On submit, the selected district ID is sent correctly.

### DIFF
--- a/src/components/Dashboard/Profile/sections/RefugeeAccommodationCentre/RefugeeAccommodationCentre.tsx
+++ b/src/components/Dashboard/Profile/sections/RefugeeAccommodationCentre/RefugeeAccommodationCentre.tsx
@@ -8,6 +8,8 @@ import { useTranslation } from "react-i18next";
 import { FormContainer } from "../shared/styles";
 import { EditableSectionProps, EditableSectionRef } from "../shared/types";
 import { useEditingChangeNotifier } from "../shared/useEditingChangeNotifier";
+import { useApiDistricts } from "../VolunteerProfile/hooks";
+import { createMapping } from "../VolunteerProfile/mappingUtils";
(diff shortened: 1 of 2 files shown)


nadavosa (commented):
## Review

Good pattern — consistent with how other sections handle district lookups via `createMapping`.

**Issue — falsy check unsafe for numeric IDs:**

```ts
const districtId = districtMapping.titleToId[values.district];
...(districtId && { district: districtId })
```

If `districtId` is `0`, the spread is skipped even though `0` is a valid ID. Use an explicit check instead:

```ts
...(districtId !== undefined && { district: districtId })
```

**Worth verifying — initial district value:**

The form's `initialFormValues.district` needs to be the district *title* string (not its numeric ID) for the radio-list to pre-select the correct option. If the stored value is an ID, the dropdown will open blank. Please confirm this is handled, or map `details?.district` → title via `districtMapping.idToTitle` before setting default values.

Cy-fox:
Fixed the falsy check to use `districtId !== undefined`. The `initialFormValues` already sets district as the title string via `districtMapping.idToTitle` so the dropdown pre-selects correctly.

nadavosa (approved):

PR #379, opened 2026-04-22
fix: Fetch and display connected opportunities on agent profile

### LINKED ISSUE #353: Agent profile missing connected Volunteering Opportunities list
### Summary

When viewing an Agent's profile page, the associated Volunteering Opportunities (VOs) are not displayed. The profile should show all VOs connected to that specific agent.

### Steps to Reproduce

.

### Expected Behavior

The Agent's profile should list/display all Volunteering Opportunities connected to that agent.

### Actual Behavior

The Volunteering Opportunities section is missing from the profile page. Currently only showing:
- Registration date
- Engagement status
- Volunteer search status
- Responsiveness

No section for connected VOs is visible.

Image

### Environment

_No response_

Cy-fox:
Closes #353
## Description
The Opportunities section on the agent profile existed as a stub, it always showed an empty state regardless of how many opportunities the agent had. The backend endpoint GET /api/agent/:id/opportunity-linked was already in place and returning the agent's opportunities, but the frontend was never calling it.

## Changes
- AgentOpportunities.tsx: replaced the stub with a real data fetch from the agent opportunities endpoint. Each opportunity is rendered as an accordion item showing its title and status, with a link to the opportunity profile.
- useAgentProfileSections.tsx: passes the agent ID down to AgentOpportunities so it can make the fetch.

### DIFF
--- a/src/components/Dashboard/Profile/sections/AgentOpportunities/AgentOpportunities.tsx
+++ b/src/components/Dashboard/Profile/sections/AgentOpportunities/AgentOpportunities.tsx
@@ -1,32 +1,49 @@
-import { OpportunityVolunteerStatusType } from "need4deed-sdk";
+import { Heading4 } from "@/components/styled/text";
+import { apiPathAgent, cacheTTL } from "@/config/constants";
+import { useGetQuery } from "@/hooks/useGetQuery";
+import { ApiOpportunityGetList, Id } from "need4deed-sdk";
+import { useRouter } from "next/navigation";
 import { useTranslation } from "react-i18next";
+import { Accordion } from "../shared/Accordion";
 import { SectionEmptyState } from "../shared/styles";
-import { Tabs } from "../shared/Tabs";
-import { useTabTransitions } from "../shared/useTabTransitions";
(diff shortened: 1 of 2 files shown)


nadavosa (commented):
## Review

Clean implementation — good use of `useGetQuery` with `enabled: !!agentId` guard.

**One concern — translation key for status:**

```ts
subtitle={t(`dashboard.opportunities.status.${opp.statusOpportunity}`)}
```

If `opp.statusOpportunity` is `undefined` or an unexpected value, i18n will silently display the raw key path instead of a label. Worth adding a fallback:

```ts
subtitle={opp.statusOpportunity ? t(`dashboard.opportunities.status.${opp.statusOpportunity}`) : "—"}
```

Also: are the `dashboard.opportunities.status.*` translation keys already in `translations.json`? They're not added in this PR — if they're missing, all statuses will show as untranslated keys.

Otherwise looks solid.

Cy-fox:
Fixed the subtitle fallback. The dashboard.opportunities.status.* translation keys are already present in translations.json so statuses will display correctly.

nadavosa (approved):

PR #346, opened 2026-04-20
Fix: Add District to the Accompanying details section #342

lourooo:
Fix: Add District to the Accompanying details section #342

## Description
Adds the District field to the Accompanying Details section on the frontend. Backend support is still needed to persist the value.

## Related Issues
Closes #342 

## Changes
- Added `appointmentDistrict` field to the accompanying details schema
- Added district field to display and edit components
- Added translations in EN and DE

## Screenshots / Demos
image

## Checklist
- [ ] WITHIN THE SCOPE OF AN ISSUE; No unnecessary files included
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

PR #389, opened 2026-04-24
fix: rename language field to refugee language and enable multi-select

### LINKED ISSUE #388: Accompanying details: change type of list
### Summary

Lanaguge to translate should be renamed and have multible options 

### Steps to Reproduce

Back end is sending arreys and not singal value - which is good. burt in FE only one languge can be selected which is bad
Rename : language to translate to "Refugee language" 

### Expected Behavior

add additional languges and the saved values should show multiple languages 

Image

### Actual Behavior

only one 

Image

### Environment

_No response_

Cy-fox:
Closes #388
### Description
The "Language to translate" field in Accompanying Details only allowed selecting one language and the name did not match the intended purpose. It now shows as "Refugee language", allows selecting multiple languages via checkboxes, and sends the full array to the API.
### Changes:
- Renamed field label from "Language to translate" to "Refugee language" in EN and DE translations
- Changed form field from single string to string array in schema and form state
- Switched EditableField from radio-list to checkbox-list to allow multi-select
- Submit handler now sends the full selected array instead of wrapping a single value
### How to test:
1. Open an accompanying-type opportunity profile
2. Go to Accompanying Details and click Edit
3. Confirm the field is now labelled "Refugee language" and shows checkboxes
4. Select multiple languages, save, and confirm all selections are preserved

### DIFF
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -1029,6 +1029,7 @@
         "refugeeNumber": "Geflüchtetennummer",
         "refugeeName": "Name des Geflüchteten",
         "languageToTranslate": "Zu übersetzende Sprache",
+        "refugeeLanguage": "Flüchtlingssprache",
         "cancel": "Abbrechen",
         "saveChanges": "Änderungen speichern",
         "saveSuccess": "Begleitdetails erfolgreich aktualisiert",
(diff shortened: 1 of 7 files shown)


nadavosa (approved):
Correct upgrade — switching from single `radio-list` to multi-select `checkbox-list`, schema from `z.string()` to `z.array(z.string())`, and the display join logic all look right.

One heads-up: this PR touches the same AccompanyingDetails files as #346 (schema, helpers, edit, main component). Whichever merges second will need a rebase. Recommend merging this one first since #346 already has conflicts with develop.

PR #397, opened 2026-04-24
fix: split volunteer languages into main communication and translate fields

### LINKED ISSUE #396: Volunteer profile: language purpose field ignored, all languages shown in one row
## Description

On the volunteer profile page, the Languages section shows all languages in a single row regardless of their purpose. The `volunteer.languages` array has a `purpose` field on each entry (`LangPurpose.GENERAL` for main communication, `LangPurpose.TRANSLATION` for translation). This field is currently ignored, so both types are merged into one block with no distinction.

## Expected behaviour

Languages split into two separate rows:
- **Main communication**, entries with `purpose: GENERAL`
- **Language to translate**, entries with `purpose: TRANSLATION`
## Actual behaviour

All languages shown in a single "Languages" row, mixing communication and translation languages.

## Notes

Previous attempt to fix this was PR #362 (closed, was incorrectly linked to issue #352 which is about the opportunity profile, not the volunteer profile). The fix is straightforward: filter `volunteer.languages` by purpose when rendering the two rows.

Cy-fox:
Closes #396

### Description
On the volunteer profile page all languages were displayed in a single "Languages" row regardless of their purpose. The volunteer.languages array has a purpose field on each entry (LangPurpose.GENERAL for main communication, LangPurpose.TRANSLATION for translation) that was being ignored, so both types were mixed into one block with no distinction.

The section now filters the languages array by purpose and renders two separate rows: "Main communication" and "Language to translate". Translation labels updated in EN and DE.

### Changes:
- Add optional purpose filter to `formatLanguagesForDisplay`
- Replace single languages prop in DisplayFields with `mainCommunication` and `languagesToTranslate`
- Pass both filtered language strings from `VolunteerProfile`
- Add i18n keys `mainCommunication` and `languagesToTranslate` in EN and DE
- Restore languages translation key alongside new keys (keeps backwards compatibility)
How to test:
1. Open any volunteer profile at /dashboard/volunteers/[id]
2. Confirm languages section shows two rows: Main communication and Language to translate
3. Confirm each row only shows languages matching its purpose

### DIFF
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -590,7 +590,8 @@
       "profileSection": {
         "title": "Freiwilligenprofil",
         "edit": "Bearbeiten",
-        "languages": "Sprachen",
+        "mainCommunication": "Hauptkommunikation",
+        "languagesToTranslate": "Sprache zum Übersetzen",
         "availability": "Verfügbarkeit",
         "districts": "Bezirke",
         "volunteerType": "Freiwilligen-Typ",
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -590,6 +590,7 @@
       "profileSection": {
         "title": "Freiwilligenprofil",
         "edit": "Bearbeiten",
+        "languages": "Sprachen",
         "mainCommunication": "Hauptkommunikation",
         "languagesToTranslate": "Sprache zum Übersetzen",
         "availability": "Verfügbarkeit",
--- a/public/locales/en/translations.json
+++ b/public/locales/en/translations.json
@@ -589,6 +589,7 @@
       "profileSection": {
         "title": "Volunteer profile",
         "edit": "Edit",
+        "languages": "Languages",
         "mainCommunication": "Main communication",
         "languagesToTranslate": "Language to translate",
         "availability": "Availability",
--- a/public/locales/en/translations.json
+++ b/public/locales/en/translations.json
@@ -589,7 +589,8 @@
       "profileSection": {
         "title": "Volunteer profile",
         "edit": "Edit",
-        "languages": "Languages",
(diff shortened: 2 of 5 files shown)


lourooo:
Great work on the new `SelectInputField` component and the `LanguageLevel` enum, the approach is clean and well structured. A few things to address before merging.

There's a **Rules of Hooks violation** in `BecomeVolunteer.tsx` that needs fixing. The `useCallback` and `useMemo` calls are placed after the early return on line 111:

```tsx
if (showErrorAnnouncement) {
  return <ErrorAnnouncement copies={somethingWrong} />;
}

const updateLanguage = useCallback(...); // called after conditional return
const disabledLanguages = useMemo(...);  //  called after conditional return
```

React requires hooks to always be called at the top level, never after a conditional return. This will cause a runtime error. All `useCallback`/`useMemo` calls need to be moved above that early return.

There are also a couple of minor issues: two `<hr />` elements in a row inside the availability fieldset (one should be removed), a duplicate `.form-languages-grid` responsive block in `index.module.css` (the second one can be deleted), and `prevLevel` in `SelectInputField`'s `Props` interface is typed as a string literal union when `LanguageLevel` already covers those exact values — better to use the enum there for consistency:

```ts
// before
prevLevel: "languagesNative" | "languagesFluent" | "languagesIntermediate";

// after
prevLevel: LanguageLevel;
```

nadavosa (approved):
Clean split — correct use of the `purpose` filter on `formatLanguagesForDisplay`, translation keys look good in both locales.

One thing worth confirming: a language entry with no `purpose` set will be invisible in both fields (filtered out by both `LangPurpose.GENERAL` and `LangPurpose.TRANSLATION`). If all languages are guaranteed to have a purpose this is fine — just worth a sanity check on the data.

Issue #385, opened 2026-04-23
feat: show briefed checkmark next to volunteer type tag for accompanying volunteers


need4deed:
## Description

For volunteers whose type is **Accompanying** or **Regular & Accompanying**, show a **checkmark (✓) icon** right-aligned next to their volunteer type tag — but only if they have been marked as **briefed** in the communication section of their volunteer profile.

## Where it should appear

All three places where volunteer type is displayed:
1. **Volunteer profile header** — upper section of the volunteer profile page
2. **Volunteer search cards** — in the volunteer list/search view
3. **Volunteer section on opportunity cards** — where volunteers are listed under an opportunity

## Behaviour

- Only shown for `Accompanying` and `Regular & Accompanying` volunteer types (not Regular or Events)
- Only shown if the volunteer is marked as briefed in the communication section
- The ✓ icon sits directly to the right of the volunteer type tag
- No icon shown if not briefed

## Notes

- The "briefed" flag already exists in the communication section — implementor needs to identify the field/value used there and read it in the three display locations above

lourooo:
I can take this!

PR #403, opened 2026-04-25
Fix hero overlap caused by fixed header by adding layout spacing in Landing page

Ihiechi:
## Description
Fixes an issue where the hero section was rendered under the fixed header.

The header uses position: fixed which removes it from normal document flow, causing the hero content to overlap.

This fix adds layout spacing in the Landing page to offset the header height and ensure proper rendering of all sections.

## Related Issues
Thanks for pointing that out — you're right, this fix is not related to issue #385. I've removed the reference from the PR description.

## Changes
- Added top spacing in Landing layout to prevent hero overlap with fixed header

## Screenshots / Demos
N/A (layout fix)

## Checklist
- [ x] WITHIN THE SCOPE OF AN ISSUE; No unnecessary files included
- [ ] Tests added/updated
- [ ] Documentation updated
- [ x] CI passes

### DIFF
--- a/src/components/Website/Landing/Landing.tsx
+++ b/src/components/Website/Landing/Landing.tsx
@@ -18,8 +18,9 @@ import { Hero } from "./Hero";
 export function Landing({ lang }: { lang: Lang }) {
   const screenType = useScreenType();
   const isBurgerMenu = screenType === ScreenTypes.MOBILE;
+
   return (
-    <AppContainer id="app-container" style={{ paddingTop: "80px" }}>
+    <AppContainer id="app-container" style={{ paddingTop: "var(--layout-static-page-header-height)" }}>
       <Header
         logo={<N4DLogo />}
         isBurgerMenu={isBurgerMenu}
--- a/src/components/Website/Landing/Landing.tsx
+++ b/src/components/Website/Landing/Landing.tsx
@@ -19,7 +19,7 @@ export function Landing({ lang }: { lang: Lang }) {
   const screenType = useScreenType();
   const isBurgerMenu = screenType === ScreenTypes.MOBILE;
   return (
-    <AppContainer id="app-container">
+    <AppContainer id="app-container" style={{ paddingTop: "80px" }}>
       <Header
         logo={<N4DLogo />}
         isBurgerMenu={isBurgerMenu}


ivannissimrch:
I noticed it's linked to close issue #385  (briefed checkmark on volunteer type tags) but the change looks like a Landing page layout fix just wanted to flag it in case the wrong issue number got linked by mistake.

nadavosa (changes requested):
**PR #403 Review: Fix landing page overlap with fixed header**

The bug is real and the approach is understandable, but the implementation is fragile:

**1. Hardcoded `80px` breaks on mobile/tablet (main issue)**
The project already defines a responsive CSS variable for the header height:
- Mobile: `--layout-static-page-header-height: 64px`
- Tablet: `--layout-static-page-header-height: 76px`
- Desktop: `--layout-static-page-header-height: 80px`

`style={{ paddingTop: "80px" }}` only matches the desktop value — on mobile it over-compensates by 16px, on tablet by 4px. The fix is:
```tsx
style={{ paddingTop: "var(--layout-static-page-header-height)" }}
```
This stays in sync with the header height automatically across all breakpoints.

**2. Inline style bypasses the design system**
The project uses styled-components throughout. Applying layout spacing via an inline `style` prop is inconsistent — consider moving this padding into the `AppContainer` styled component or a wrapper, so it's part of the design token system.

**3. PR checklist**
"Tests added/updated" and "Documentation updated" are checked, but the diff is a single-line style change with no test or documentation changes.

Ihiechi:
Thanks for the feedback! I've updated the implementation to use the responsive CSS variable so it adapts across breakpoints, and kept the change minimal to avoid affecting layout structure. Also corrected the checklist items.

src/components/Website/Landing/Landing.tsx
```diff
@@ -18,8 +18,9 @@ import { Hero } from "./Hero";
 export function Landing({ lang }: { lang: Lang }) {
   const screenType = useScreenType();
   const isBurgerMenu = screenType === ScreenTypes.MOBILE;
+
   return (
-    <AppContainer id="app-container">
+    <AppContainer id="app-container" style={{ paddingTop: "var(--layout-static-page-header-height)" }}>
```
arturasmckwcz, line 23: we avoid in-line styling

nadavosa:
## Review — Fix landing page overlap with fixed header

The CSS variable `var(--layout-static-page-header-height)` is correct and responsive across breakpoints — this addresses the original review feedback.

One remaining note: the fix uses an inline `style` prop, which is inconsistent with the codebase convention of using `styled-components` for all layout styling. The cleaner approach is to add `padding-top: var(--layout-static-page-header-height)` directly to the `AppContainer` styled component. This is a minor style issue, not a blocker — but worth aligning with Hans's pattern before merge.

nadavosa (approved):

PR #407, opened 2026-04-27
fix: convert appointment time from UTC to local time for display

### LINKED ISSUE #393: Fix: Accompanying opportunity appointment time displays 2 hours early
## Bug

Appointment times for accompanying opportunities created via the new form are displayed 2 hours earlier than the time the user entered.

## Root Cause

In `src/components/forms/utils.ts:159`, the `datetime-local` input value is converted to UTC before being sent to the API:

```js
const accomp_datetime = isAccompanying ? (form.dateTime ? new Date(form.dateTime).toISOString() : null) : null;
```

`toISOString()` converts the user's local time to UTC (e.g. `10:30` Berlin/CEST → `08:30Z`). The backend stores and returns the time as-is, so the display shows `08:30` instead of `10:30`.

## Evidence that UTC conversion is wrong

1. **`onetime_date_time` (same file, line 197) sends the raw `datetime-local` value directly** — no `.toISOString()`. Same field type, same form, no UTC conversion. This is the correct pattern.
2. **`getDateLocalTooUTC` utility (`src/utils/index.ts:6`) exists but is never called anywhere** — suggesting UTC handling was considered and abandoned.
3. **The update (edit) form sends `appointmentTime` as a raw `HH:MM` string** and displays correctly — confirming the backend stores and returns local time as-is.

## Fix

Remove the UTC conversion and pass the `datetime-local` value directly, consistent with `onetime_date_time`:

```js
// before
const accomp_datetime = isAccompanying ? (form.dateTime ? new Date(form.dateTime).toISOString() : null) : null;

// after
const accomp_datetime = isAccompanying ? (form.dateTime || null) : null;
```

## Affected file

`src/components/forms/utils.ts` line 159

Cy-fox:
Closes #393

Appointment times were stored as UTC HH:mm strings by the API but rendered without any timezone conversion, causing them to appear 2 hours early for users in CEST (UTC+2). For example, an appointment at 10:30 CEST would show as 08:30.

Changes:
- `parseTime` in `AccompanyingDetails` helpers now converts a UTC HH:mm string to local time by setting UTC hours on a Date object and extracting via `toTimeString`, so the profile detail view shows the correct local time
- `formatAccompanyingDate` in OpportunityCard.helpers now formats the raw `appointmentDate` ISO string with date-fns and applies the same UTC-to-local conversion for `appointmentTime`, so the card view shows a human-readable date and the correct local time

### DIFF
--- a/src/components/Dashboard/Opportunities/OpportunityCard.helpers.tsx
+++ b/src/components/Dashboard/Opportunities/OpportunityCard.helpers.tsx
@@ -1,4 +1,5 @@
 import { ConfettiIcon, PersonSimpleWalkIcon, ShootingStarIcon, TranslateIcon } from "@phosphor-icons/react";
+import { utcHhmmToLocal } from "@/utils";
 import { format } from "date-fns";
 import { ApiVolunteerOpportunityGetList, OpportunityStatusType, ProfileVolunteeringType } from "need4deed-sdk";
 import { JSX } from "react";
@@ -19,17 +20,7 @@ export function formatAccompanyingDate(details?: {
   const date = new Date(details.appointmentDate);
   const formattedDate = isNaN(date.getTime()) ? details.appointmentDate : format(date, "dd.MM.yyyy");
 
-  let formattedTime: string | null = null;
-  if (details.appointmentTime) {
-    const [h, m] = details.appointmentTime.split(":").map(Number);
-    if (!isNaN(h) && !isNaN(m)) {
-      const d = new Date();
-      d.setUTCHours(h, m, 0, 0);
-      formattedTime = d.toTimeString().slice(0, 5);
-    } else {
-      formattedTime = details.appointmentTime;
-    }
-  }
+  const formattedTime = details.appointmentTime ? utcHhmmToLocal(details.appointmentTime) : null;
 
   return [formattedDate, formattedTime].filter(Boolean).join(" ");
 }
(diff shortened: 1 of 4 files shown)


nadavosa (changes requested):
**PR #407 Review: Convert appointment time from UTC to local time for display**

The display-only paths (OpportunityCard card view, `AccompanyingDetailsDisplay`) are correct. However there is a critical bug in the edit flow:

**Critical: Round-trip data corruption in the edit form**

`parseTime` in `AccompanyingDetails/helpers.ts` is used by `getInitialFormValues` to populate the React Hook Form default values, including `appointmentTime`. The `<TimeInput>` is pre-filled with the already-converted local time. On submit, `AccompanyingDetails.tsx` sends `values.appointmentTime` directly back to the API unchanged. This means:

1. API stores `08:30` (UTC)
2. `parseTime` converts it to `10:30` (CEST local)
3. User opens edit form — sees `10:30` ✓
4. User saves without changing — API receives `10:30` instead of `08:30`
5. Next read: `parseTime` converts `10:30` → `12:30`

**Each save shifts the appointment time forward by the UTC offset.** The fix should only convert for display, not for form state — either keep the raw UTC string in form state and convert only in display components, or convert back to UTC before submitting.

**Medium: Duplicated conversion logic**
The same 5-line UTC-to-local block appears in both `OpportunityCard.helpers.tsx` and `AccompanyingDetails/helpers.ts`. Should be extracted into a shared utility.

**Low: `toTimeString().slice(0, 5)` relies on implementation-defined format**
Works in all major browsers but `String(d.getHours()).padStart(2, "0") + ":" + String(d.getMinutes()).padStart(2, "0")` would be more explicit.

**Info: DST awareness**
When only `HH:mm` is stored without a date, DST transitions (summer vs winter) may produce a 1-hour discrepancy. Inherent limitation of the storage format.

PR #409, opened 2026-04-27
fix: implement save mutation in opportunity details edit form

### LINKED ISSUE #401: Opportunity profile: edited information in opportunity details is not saved
## Description

When editing fields in the Opportunity details section of an opportunity profile and saving, the changes are not persisted. After saving, the form reverts to the previous values.

## Steps to reproduce

1. Open an opportunity profile
2. Edit one or more fields in the Opportunity details section
3. Click Save

## Expected behaviour

The updated values are saved and shown after the page reloads or the section refreshes.

## Actual behaviour

The edits are lost, the fields show the original values after saving.

Cy-fox:
Closes #401

The save button in OpportunityDetails had a placeholder onSubmit that only closed the form without persisting any changes. Added the PATCH mutation and conversion helpers to send form values to the API.

Changes:
- Add `useUpdateOpportunityDetails` hook that PATCHes the opportunity endpoint
- Implement onSubmit in `OpportunityDetailsEdit` to call the mutation with description, `numberVolunteers`, languages, activities, skills and schedule
- Add `toLangOptionItems` and `toOptionItems` helpers to convert form state to the `OptionItem` shape expected by `ApiOpportunityPatch`
- Add `saveSuccess` i18n keys to EN and DE translation files

### DIFF
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -1073,6 +1073,7 @@
         "descriptionHint": "Max. {{max}} Zeichen",
         "cancel": "Abbrechen",
         "saveChanges": "Änderungen speichern",
+        "saveSuccess": "Opportunity-Details erfolgreich aktualisiert",
         "validation": {
           "descriptionRequired": "Beschreibung ist erforderlich",
           "descriptionTooLong": "Beschreibung ist zu lang",
--- a/public/locales/en/translations.json
+++ b/public/locales/en/translations.json
@@ -1072,6 +1072,7 @@
         "descriptionHint": "Max. {{max}} characters",
         "cancel": "Cancel",
         "saveChanges": "Save changes",
+        "saveSuccess": "Opportunity details updated successfully",
         "validation": {
           "descriptionRequired": "Description is required",
           "descriptionTooLong": "Description is too long",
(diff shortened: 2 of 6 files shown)


nadavosa (changes requested):
**PR #409 Review: Implement save mutation in opportunity details edit form**

**Critical: `eventDate`/`eventTime` are never sent to the API for event-type opportunities**
`OpportunityDetailsEdit.tsx` `onSubmit`: When `isEventType` is true, the form shows `eventDate` and `eventTime` fields, but the mutation payload only includes `description`, `numberVolunteers`, languages, activities, skills, and `schedule`. Event date/time are collected from the user and then silently discarded.

**High: `toLangOptionItems` matches by translated display name, not by canonical ID**
`toLangOptionItems` compares `field.value` (a *translated* language name in the user's locale) against `apiLanguage.title` (the English/canonical name). In German locale, "Arabisch" won't match "Arabic" — the language gets silently dropped from the payload. The lookup should use language `id` throughout, not the string title.

**Medium: `parseTime` timezone bug propagated to two new locations**
`d.setUTCHours(h, m)` followed by `d.toTimeString().slice(0, 5)` reads UTC-set hours in local time — in UTC+2 this displays 2 hours ahead. This was a pre-existing bug, but the PR copies it into `formatAccompanyingDate` in `OpportunityCard.helpers.tsx`. Should use `String(d.getUTCHours()).padStart(2,'0') + ":" + String(d.getUTCMinutes()).padStart(2,'0')` instead.

**Low: `formatAccompanyingDate` added to `OpportunityCard.helpers.tsx` but not used**
The new function is added but no call site in the diff uses it. If speculative, defer to the PR that actually needs it.

PR #412, opened 2026-04-27
Fix waysToContact not displaying when API returns a string value

### LINKED ISSUE #399: Bug Opportunity profile: contact details section missing phone number
## Description

The contact details comment on an opportunity profile does not display the refugee's phone number. Coordinators need the phone number to be able to contact them.

Example: /dashboard/opportunities/812

## Expected behaviour

Phone number is shown in the contact details section of the opportunity profile.

## Actual behaviour

Phone number is missing from the contact details comment.

Cy-fox:
Closes #399
The API returns `waysToContact` as a plain string (e.g. `"mobilePhone"`) instead of an array. The existing guard only checked `Array.isArray`, so a string value was silently dropped and the field always displayed "–" even when data existed.
Changes:
- `OpportunityContactDetails.tsx`: handle `waysToContact` as either a `PreferredCommunicationType[]` or a single string, wrapping the string case in a one-element array
- Made the other contact fields (`name`, `phone`, `email`) explicitly default to `""` instead of relying on the spread to avoid undefined values slipping through

### DIFF
--- a/src/components/Dashboard/Profile/sections/ContactDetails/opportunity/OpportunityContactDetails.tsx
+++ b/src/components/Dashboard/Profile/sections/ContactDetails/opportunity/OpportunityContactDetails.tsx
@@ -39,13 +39,20 @@ export const OpportunityContactDetails = forwardRef<EditableSectionRef, Props>(f
 
   const schema = createOpportunityContactDetailsSchema(t);
 
-  const initialFormValues = useMemo(
-    () => ({
-      ...opportunity.contact,
-      waysToContact: Array.isArray(opportunity.contact.waysToContact) ? opportunity.contact.waysToContact : [],
-    }),
-    [opportunity.contact],
-  );
+  const initialFormValues = useMemo(() => {
+    const raw = opportunity.contact.waysToContact;
+    const waysToContact: PreferredCommunicationType[] = Array.isArray(raw)
+      ? raw
+      : typeof raw === "string" && raw
+        ? [raw as PreferredCommunicationType]
+        : [];
+    return {
+      name: opportunity.contact.name ?? "",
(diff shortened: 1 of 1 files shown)


nadavosa (changes requested):
**PR #412 Review: Fix waysToContact not displaying when API returns a string**

The core fix is correct — wrapping a bare string in an array so it renders. A few issues worth addressing:

**1. No enum validation before the cast (correctness)**
The `typeof raw === "string" && raw` guard only checks non-empty, not that the value is a valid `PreferredCommunicationType`. An unexpected value (wrong casing, deprecated key) gets cast and silently reaches the form or API. Consider:
```ts
const isValid = (v: string): v is PreferredCommunicationType =>
  Object.values(PreferredCommunicationType).includes(v as PreferredCommunicationType);
const waysToContact = Array.isArray(raw)
  ? raw.filter(isValid)
  : isValid(raw) ? [raw] : [];
```
This would protect the array branch too.

**2. Explicit field listing may drift from schema**
The old spread (`...opportunity.contact`) included all contact fields automatically. The new explicit list (`name, phone, email, waysToContact`) will silently miss any new fields added to `OpportunityContactDetailsFormData`. A comment noting the list must stay in sync with the schema would help.

**3. `createOpportunityContactDetailsSchema(t)` called outside `useMemo` (pre-existing)**
This creates a new schema object on every render, causing the `useForm` resolver to change each render. Not introduced by this PR but it's in the touched area — worth fixing in a follow-up.

nadavosa (approved):
Both issues addressed correctly:
- Enum validation now uses `new Set<string>(COMMUNICATION_TYPES)` — covers both the string and array branches cleanly
- Sync comment added to the explicit field list

Approving.

PR #414, opened 2026-04-27
Fix opportunity details form blocked when no schedule is set

### LINKED ISSUE #400: Opportunity profile: opportunity details not saved unless schedule is also filled in
## Description

When editing an opportunity profile, changes made in the Opportunity details section are not saved if the Schedule section is left empty. The save action silently fails or is blocked by schedule validation, even when the user only intends to update the details fields.

This does not apply to Accompanying type opportunities, which have no schedule section.

## Steps to reproduce

1. Open an opportunity profile of type other than Accompanying (e.g. a regular volunteering opportunity)
2. Edit a field in the Opportunity details section
3. Leave the Schedule section empty
4. Save

## Expected behaviour

Opportunity details are saved independently of whether a schedule is entered.

## Actual behaviour

Changes to opportunity details are not saved when the schedule section is empty.

Image

Cy-fox:
Closes #400

The Save Changes button on the opportunity details edit form was permanently
disabled for opportunities with no schedule set. The Zod schema for the
availability field ran a custom validator that required at least one time slot
to be selected, but `apiToFormAvailability()` always returns a full grid
(all slots deselected) even when the opportunity has no schedule, making the
form invalid on load with no way to fix it.

Since schedule is an optional field for opportunities, the custom validator is
not needed here. Removed it so the form is valid regardless of whether a
schedule has been entered.

Changes:
- `opportunityDetailsSchema.ts`: replaced the custom availability validator
  (which required ≥1 selected slot) with z.custom().nullable().optional(),
  making the field truly optional

### DIFF
--- a/src/components/Dashboard/Profile/sections/OpportunityDetails/opportunityDetailsSchema.ts
+++ b/src/components/Dashboard/Profile/sections/OpportunityDetails/opportunityDetailsSchema.ts
@@ -34,17 +34,7 @@ export const createOpportunityDetailsSchema = (t: (key: string) => string) =>
     }),
     mainCommunication: languagesValidator(t),
     residentsSpeak: languagesValidator(t),
-    availability: z
-      .custom<Availability>(
-        (data) => {
-          if (data === null || data === undefined) return true;
-          if (!Array.isArray(data)) return false;
-          return data.some((day) => day.timeSlots.some((slot: { selected: boolean }) => slot.selected));
-        },
-        t(`${i18nPrefix}.availabilityRequired`),
-      )
-      .nullable()
-      .optional(),
+    availability: z.custom<Availability>().nullable().optional(),
     eventDate: z.date().nullable().optional(),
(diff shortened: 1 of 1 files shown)


nadavosa (approved):
Fix is correct and unblocks the form. Two minor points worth confirming:

**1. Clearing all slots now submits an empty schedule — is that intentional?**

The old validator blocked this, which was also the source of the bug. Now a user can load a form with an existing schedule, deselect everything, and submit. Please confirm that `formToApiAvailability` converts the empty grid correctly and the API handles an empty/null schedule gracefully.

**2. `z.custom<Availability>()` with no predicate accepts any runtime value**

The type parameter is TypeScript-only — Zod won't validate the shape at runtime. Not a blocker since `AvailabilityGrid` controls the output, but a structural check like `z.array(z.any()).nullable().optional()` would be safer if the component ever changes.

Approving since the fix is correct and the points above are low-risk, but worth a quick confirmation on point 1.

PR #415, opened 2026-04-27
Fix: Missing comments row on Volunteer Profile #413

### LINKED ISSUE #413: Missing comments row on Volunteer Profile
### Description

When the volunteer leave a comment on https://app.need4deed.org/en/forms/volunteer the comment does not appear on the Volunteer Profile

### Proposed solution

Add comments row below skills & experience

Image

lourooo:
Fix: Missing comments row on Volunteer Profile #413

## Description
The volunteer profile was missing the Comments field from the display view, this adds the missing row.

## Related Issues
Closes #413 

## Changes
Added comments translation key to EN and DE locale files under dashboard.volunteerProfile.profileSection
Added infoAbout prop to DisplayFields and rendered it as a new field row below "Skills & experience"
Passed volunteer.infoAbout from VolunteerProfile into DisplayFields

## Screenshots / Demos
image

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -607,6 +607,7 @@
         "volunteerType": "Freiwilligen-Typ",
         "activities": "Aktivitäten",
         "skills": "Fähigkeiten & Erfahrungen",
+        "comments": "Kommentare",
         "cancel": "Abbrechen",
         "saveChanges": "Änderungen speichern",
         "saveSuccess": "Profil erfolgreich aktualisiert",
--- a/public/locales/en/translations.json
+++ b/public/locales/en/translations.json
@@ -606,6 +606,7 @@
         "volunteerType": "Volunteer type",
         "activities": "Activities",
         "skills": "Skills & experience",
+        "comments": "Comments",
         "cancel": "Cancel",
         "saveChanges": "Save changes",
         "saveSuccess": "Profile updated successfully",
--- a/src/components/Dashboard/Profile/sections/VolunteerProfile/DisplayFields.tsx
+++ b/src/components/Dashboard/Profile/sections/VolunteerProfile/DisplayFields.tsx
@@ -48,7 +48,7 @@ type Props = {
   volunteerTypeStatus: StatusValue | undefined;
   activities: string[];
   skills: string[];
-  infoAbout: string;
+  comments: string | undefined;
   t: (key: string) => string;
 };
 
@@ -61,7 +61,7 @@ export function DisplayFields({
   volunteerTypeStatus,
   activities,
   skills,
-  infoAbout,
+  comments,
   t,
 }: Props) {
   return (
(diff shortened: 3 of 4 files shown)


nadavosa (commented):
**PR #415 Review: Fix missing comments row on Volunteer Profile**

The fix is correct and minimal — `infoAbout` was simply never passed to `DisplayFields`, so it never rendered. The wiring is straightforward. A few minor notes:

**Minor: Misleading prop name**
The prop is named `infoAbout` (the SDK's internal field name), but the label shown is `"Comments"` and the locale key is `dashboard.volunteerProfile.profileSection.comments`. Naming it `comments` would be consistent with the locale key and less confusing for future maintainers.

**Minor: Redundant double-default**
`VolunteerProfile.tsx` passes `infoAbout={volunteer.infoAbout || ""}`, and `DisplayFields` then checks `{infoAbout || <EmptyPlaceholder />}`. The `|| ""` in the parent is redundant since an empty string is falsy and the child handles it — pick one location for the fallback.

**Note: No edit support for this field**
Users can see their comments but cannot edit them through this form. The mutation hook does list `"comments"` as an allowed key, so the infrastructure exists. If the intent is to keep it permanently read-only, a comment to that effect would help.

**Tests:** The PR checklist marks tests as not updated — if there are snapshots for `DisplayFields`, they'll need updating.

nadavosa (changes requested):
Two small things to fix before merging:

**1. Rename prop `infoAbout` → `comments`**
The label shown is "Comments" and the locale key is `dashboard.volunteerProfile.profileSection.comments`. Using the SDK's internal field name as the prop name is inconsistent with how every other prop in `DisplayFields` is named (they all match the label/locale key). Rename the prop to `comments` in both `DisplayFields.tsx` and the call site in `VolunteerProfile.tsx`.

**2. Remove redundant `|| ""` in `VolunteerProfile.tsx`**
`VolunteerProfile.tsx` passes `infoAbout={volunteer.infoAbout || ""}`, but `DisplayFields` already guards against empty with `{infoAbout || <EmptyPlaceholder />}`. An empty string is falsy, so the parent default is redundant. Keep the guard in one place — the child is the right spot since it owns the display logic.

Also note: this PR touches `DisplayFields.tsx` and `VolunteerProfile.tsx` — please rebase on `develop` before merging, as those files have been changed there.

lourooo:
Updated

nadavosa (approved):
Both fixes applied correctly — prop renamed to `comments`, redundant `|| ""` removed from the parent.

One thing to be aware of: this PR touches `DisplayFields.tsx` and `VolunteerProfile.tsx`, which have an in-flight hotfix on another branch. You may need to rebase once that merges before this can land cleanly.

PR #416, opened 2026-04-27
fix: show Farsi/Dari instead of Persian in the language filter

### LINKED ISSUE #404: Dashboard opportunities: "Farsi/Dari" filter returns no results because language is stored as "Persian"
## Description

Filtering opportunities by "Farsi/Dari" in the Dashboard returns no results. The filter value used in the UI is "Farsi/Dari" but the language stored on opportunity cards is "Persian" (visible in the Main communication field, e.g. /dashboard/opportunities/804). Because the strings do not match, the filter never finds anything.

The label needs to be consistent everywhere, either always "Farsi/Dari" or always "Persian".

## Steps to reproduce

1. Go to Dashboard > Opportunities
2. Apply the "Farsi/Dari" language filter
3. Observe that no results are returned even though opportunities with that language exist

## Expected behaviour

Filtering by "Farsi/Dari" returns all opportunities where the language is Farsi/Dari, regardless of how it is stored internally.

## Actual behaviour

No results are returned because the display label in the filter ("Farsi/Dari") does not match the value stored on the opportunity card ("Persian").

Image

Cy-fox:
Closes #404

The opportunity language filter was showing "Persian" as the checkbox label
because it used the raw title returned by the API options endpoint. However,
the rest of the UI (volunteer profiles, opportunity profiles) refers to this
language as "Farsi/Dari" using the languageNames translation section. The
mismatch meant users looking for Farsi/Dari speakers could not find the
correct filter option.

Changes:
- Opportunities/Filters/helpers.ts: language label resolver now looks up
  the API title in the languageNames translation namespace (with fallback
  to the raw title), making filter labels consistent with how languages
  appear in profiles
- translations (EN/DE): added "persian": "Farsi/Dari" to languageNames so
  the API title "Persian" maps to the display name "Farsi/Dari"

### DIFF
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -1655,6 +1655,7 @@
     "farsi/dari": "Farsi/Dari",
     "farsi": "Farsi",
     "dari": "Dari",
+    "persian": "Farsi/Dari",
     "turkish": "Türkisch",
     "russian": "Russisch",
     "ukrainian": "Ukrainisch",
--- a/public/locales/en/translations.json
+++ b/public/locales/en/translations.json
@@ -1287,6 +1287,7 @@
     "farsi/dari": "Farsi/Dari",
     "farsi": "Farsi",
     "dari": "Dari",
+    "persian": "Farsi/Dari",
     "turkish": "Turkish",
     "russian": "Russian",
     "ukrainian": "Ukrainian",
--- a/src/components/Dashboard/Opportunities/Filters/helpers.ts
+++ b/src/components/Dashboard/Opportunities/Filters/helpers.ts
@@ -20,7 +20,11 @@ export const createOpportunityFilterItems = (
     filter[QueryParamsKeys.LANGUAGE],
     setFilter,
     QueryParamsKeys.LANGUAGE,
-    (key) => key,
+    (key) => {
+      const translationKey = `languageNames.${key.toLowerCase()}`;
+      const translated = t(translationKey);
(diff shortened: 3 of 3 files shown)


nadavosa (commented):
**PR #416 Review: Show Farsi/Dari instead of Persian in the language filter**

Fix is correct. `key.toLowerCase()` maps the API's `"Persian"` → `"persian"` → locale key `languageNames.persian` → `"Farsi/Dari"`. Both EN and DE locale files are updated. No missing locales, no broken imports.

One latent issue worth being aware of:

**Low: Fallback detection assumes i18next returns the key string verbatim on a miss**
`translated !== translationKey` works correctly with the current i18next configuration (`saveMissing: false`, no `parseMissingKeyHandler`). If those are ever configured, the fallback breaks silently. A more defensive approach is to maintain a `Set` of known language keys, but it's low-risk with the current setup.

**Info: Dot-in-key risk (pre-existing)**
i18next uses `.` as its key separator. If the API ever returns a language name containing a dot, the lookup would be misinterpreted as a nested key. Not introduced by this PR and low-probability given the current data — just worth knowing since this PR now relies on the pattern for correctness.

nadavosa (approved):
Fix is correct and complete. The i18next key-based fallback handles the Persian→Farsi/Dari mapping cleanly, and both locale files are updated. The two notes in my earlier comment (fallback detection assumption, dot-in-key risk) are low/informational — nothing blocking. Approving.

PR #417, opened 2026-04-27
Feature 411: adds newest opportunities and volunteers to dashboard home page

### LINKED ISSUE #411: Add most recent opportunities and volunteers to dashboard page
### Description

`/dashboard` Home page displays 2 newly created opportunities and 2 newly registered volunteers until their status changes from `NEW`

### Proposed solution

- Fetch Opportunities List with query selector so that we're fetching opportunities that are the newest and limited to a length of two e.g: 
`opportunity?limit=2&page=1&sortOrder=new-old&filter%5Bstatus%5D=opp-new&language=en`
- Display the two opportunities

- Fetch Volunteers List with query selector so that we're fetching volunteers that are the newest and limited to a length of two e.g.: 
`volunteer?limit=2&page=1&sortOrder=new-old&filter%5Bengagement%5D=new&language=en`
- Display the two volunteers

- Use existing Card structure to display the opportunities and volunteers like we do for the Lists
e.g. 
#### Newest Opportunities
Image

#### Newest Volunteers
Image

DarrellRoberts:
## Description
 Adds newest opportunities and volunteers to dashboard home page (2 of each)

## Related Issues
Closes #411 

## Changes
- Fetches two of newest opportunities and volunteers on `/dashboard` page

## Screenshots / Demos
image

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -399,6 +399,12 @@
       }
     },
     "home": {
+      "content": {
+        "header": "Dashboard Startseite",
+        "newOpportunities": "Neueste Möglichkeiten",
+        "newVolunteers": "Neueste Freiwillige",
+        "loading": "Wird geladen..."
+      },
       "sidebar": {
         "home": "Startseite",
         "volunteers": "Freiwillige",
--- a/public/locales/en/translations.json
+++ b/public/locales/en/translations.json
@@ -399,6 +399,12 @@
       }
     },
     "home": {
+      "content": {
+        "header": "Dashboard Home",
+        "newOpportunities": "Newest Opportunities",
+        "newVolunteers": "Newest Volunteers",
+        "loading": "Loading..."
+      },
       "sidebar": {
         "home": "Home",
         "volunteers": "Volunteers",
--- a/src/app/[lang]/globals.css
+++ b/src/app/[lang]/globals.css
@@ -1305,6 +1305,7 @@ html {
 
     --dashboard-home-container-padding: 40px 120px 100px 132px;
     --dashboard-home-container-gap: 20px;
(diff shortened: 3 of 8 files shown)


nadavosa (changes requested):
Three issues that need fixing:

**1. Missing `return` + wrong condition in both new components**

`NewestOpportunities.tsx` line 24 and `NewestVolunteers.tsx` line 22 — the early-exit block never returns anything:

```tsx
if (opportunities?.length === 0) {
  <Heading4>{t("dashboard.home.content.loading")}</Heading4>;  // no return!
}
```

Two problems:
- Missing `return` keyword — the message is never rendered
- `length === 0` means data loaded empty, not loading. Use `isLoading` from the query hook for the loading state, and handle the empty case separately

**2. Non-specific React Query keys risk cache collisions**

`queryKey: ["opportunities"]` and `queryKey: ["volunteers"]` will share the cache with the full list pages if they use the same keys. The dashboard fetches only 2 newest with `NEW` status — this needs a distinct key, e.g. `["opportunities", "newest"]` and `["volunteers", "newest"]`.

**3. `DashboardContentContainer` height removed in `HomeContent.tsx`**

The previous `height: 300px` is gone in the new `styles.ts`. If this is intentional please confirm — if children don't have intrinsic heights the container may collapse.

DarrellRoberts:
thanks @nadavosa! All very good points and well-spotted. As for point 3, I removed the height because the Newest Volunteers are vertically stacked on top of the New Opportunites, so it exceeds the height of 300px and spills into the Footer (and beyond). I added a `min-height` with a css var of 300px instead. What do you think?

nadavosa (approved):

PR #418, opened 2026-04-27
fix: wire up opportunity details save and fix UTC time round-trip bug

### LINKED ISSUE #401: Opportunity profile: edited information in opportunity details is not saved

nadavosa:
## Summary

- Adds `useUpdateOpportunityDetails` hook and wires it into `OpportunityDetailsEdit` so saving the form actually PATCHes the API (fixes https://github.com/need4deed-org/fe/issues/401)
- Fixes `toLangOptionItems` to reverse the `languagesToFormValues` translation lookup — without this, languages initialised from the API in German locale were silently dropped on save because translated names ("Arabisch") never matched canonical API titles ("Arabic")
- Adds `formatTimeForDisplay` (UTC→local) as a separate function from `parseTime` (form state stays raw), so `AccompanyingDetailsDisplay` shows local time without corrupting the saved value on each edit (fixes the round-trip drift bug in https://github.com/need4deed-org/fe/pull/407 and https://github.com/need4deed-org/fe/pull/409)
- Fixes `formatAccompanyingDate` UTC conversion to use explicit `getHours`/`getMinutes` instead of `toTimeString()` which is implementation-defined
- Removes unused `formatAccompanyingDate` function that was added but never called

## Related issues / PRs

- Fixes https://github.com/need4deed-org/fe/issues/401
- Supersedes the opportunity-details save work in https://github.com/need4deed-org/fe/pull/409
- Fixes the UTC round-trip bug introduced in https://github.com/need4deed-org/fe/pull/407 and https://github.com/need4deed-org/fe/pull/409

## Notes

`eventDate` and `eventTime` (shown for EVENTS-type opportunities) are not included in the save payload — there is no corresponding field in `ApiOpportunityPatch` and the display component already shows `—` for those fields. This needs a BE change to support.

## Test plan

- [ ] Open an opportunity profile, edit description/languages/activities/skills/numberOfVolunteers and save — changes persist after reload
- [ ] Edit in German locale — languages save correctly (not dropped)
- [ ] Open an accompanying opportunity — appointment time shows in local timezone in display mode
- [ ] Edit accompanying details, save without changing the time — time does not drift on re-open

### DIFF
--- a/src/components/Dashboard/Opportunities/OpportunityCard.helpers.tsx
+++ b/src/components/Dashboard/Opportunities/OpportunityCard.helpers.tsx
@@ -1,4 +1,5 @@
 import { ConfettiIcon, PersonSimpleWalkIcon, ShootingStarIcon, TranslateIcon } from "@phosphor-icons/react";
+import { format } from "date-fns";
 import { ApiVolunteerOpportunityGetList, OpportunityStatusType, ProfileVolunteeringType } from "need4deed-sdk";
 import { JSX } from "react";
 
@@ -14,7 +15,23 @@ export function formatAccompanyingDate(details?: {
   appointmentTime?: string;
 }): string | null {
   if (!details?.appointmentDate) return null;
-  return [details.appointmentDate, details.appointmentTime].filter(Boolean).join(" ");
+
+  const date = new Date(details.appointmentDate);
+  const formattedDate = isNaN(date.getTime()) ? details.appointmentDate : format(date, "dd.MM.yyyy");
+
+  let formattedTime: string | null = null;
+  if (details.appointmentTime) {
+    const [h, m] = details.appointmentTime.split(":").map(Number);
+    if (!isNaN(h) && !isNaN(m)) {
+      const d = new Date();
+      d.setUTCHours(h, m, 0, 0);
+      formattedTime = `${String(d.getHours()).padStart(2, "0")}:${String(d.getMinutes()).padStart(2, "0")}`;
+    } else {
+      formattedTime = details.appointmentTime;
+    }
+  }
+
+  return [formattedDate, formattedTime].filter(Boolean).join(" ");
 }
 
 export const statusColorMap: Record<OpportunityStatusType, string> = {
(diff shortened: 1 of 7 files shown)


lourooo:
Good fixes overall, the UTC time separation and the language reverse mapping bug are both real issues and the solutions look correct. The new hook follows the same pattern as the rest of the codebase, no issues there.

Two small things worth checking: eventDate and eventTime are editable in the form but won't be persisted pending backend support, worth making them read-only/disabled in edit mode to avoid silent data loss. Also, does the new hook include title in its payload? If so, the existing useUpdateOpportunityTitle call might end up firing a redundant second PATCH when the title changes.

Cy-fox:
One tiny edge-case I spotted in the time display: `formatTimeForDisplay` builds the conversion date using new Date() (today). If an appointment is months from now in a different DST period, the local time could be off by an hour. Super easy fix, just pass in the actual `appointmentDate` from the details and use that as the base. It’ll make the conversion rock-solid without touching the rest of the logic.

arturasmckwcz:
> eventDate and eventTime are editable in the form but won't be persisted pending backend support
@lourooo 
hey Rodrigo to which API endpoint and what `body` is sent for writing?

lourooo:
It's the same PATCH /api/opportunity/:id endpoint already used by useUpdateOpportunityDetails. The SDK type ApiOpportunityPatch already has the accompanyingDetails field for it:

accompanyingDetails: {
  appointmentDate?: string;
  appointmentTime?: string;
  // ...
}

So the fix would just be adding accompanyingDetails to the updateOpportunityDetails call in onSubmit:

accompanyingDetails: isEventType ? undefined : {
  appointmentDate: values.eventDate,
  appointmentTime: values.eventTime,
},

Issue #392, opened 2026-04-24
sending patch data from agent profile


need4deed:
### Summary

data sent from agent profile not good

### Steps to Reproduce

a

### Expected Behavior

a

### Actual Behavior

a

nadavosa:
Closing — issue body was empty and not actionable. If this is still a problem, please open a new issue with steps to reproduce and expected vs actual behaviour.

Issue #354, opened 2026-04-20
Agent profile edit only allows updating Name field - Address and District missing


Cy-fox:
### Summary

The Agent profile edit form only allows updating the Name field. The Address and District fields are either missing from the form, disabled, or not saving when edited.

### Steps to Reproduce

.

### Expected Behavior

The form should allow editing and saving of:
- Name
- Address  
- District

### Actual Behavior

Only the Name field can be updated. Address and District fields are not available in the edit form or do not save when present.

https://github.com/user-attachments/assets/f7901aea-2e5b-449c-ac4a-e05bab2b049f

DarrellRoberts:
So for the example you showed: `id 485`, there is no representative associated with this RAC. Therefore there is no representative id which is sent in the `Request` hence the validation fail:

<img width="615" height="252" alt="Image" src="https://github.com/user-attachments/assets/ed1a163f-3557-4ee1-9663-3edcf183b1e1" />

I'm guessing this is a test/seeded agent? Are real RACs with representatives attached seeing the same issue?

DarrellRoberts:
hey @arturasmckwcz , is there a controller on the backend to add `representative` to `Agents` if they currently don't exist? Or would the `representative` be automically added when the Agent is created?

arturasmckwcz:
> is there a controller on the backend to add representative to Agents if they currently don't exist? Or would the representative be automically added when the Agent is created?

`representative` is a getter class field actually, it calculates the value each time one refers to it
you definitely cannot patch its value directly. it should provide id which is person id, so you can `PATCH /person/id`

DarrellRoberts:
ah ok, thank you!

PR #419, opened 2026-04-27
fix: correct agent profile save mutations and cache invalidation

nadavosa:
## Summary

Fixes two broken save hooks in the agent profile that caused edits to silently fail or not refresh the UI:

**Contact details** (https://github.com/need4deed-org/fe/issues/392):
- `useUpdateAgentContact` was invalidating `["agent", personId]` (the representative's person ID) instead of `["agent", agentId]` (the agent entity ID). The cache key used when fetching the profile is `["agent", entityId]` where `entityId` is the agent's own ID, so the profile never refreshed after saving.

**Organisation details** (https://github.com/need4deed-org/fe/issues/354):
- `useUpdateOrganization` was calling `PATCH /api/organization/:operator` where `operator` is a free-text string (e.g. "Bayern Rotes Kreuz"), not a numeric ID — every save was a 404 that silently failed.
- The hook also used `DeepPartial<ApiRepresentativeGet>` (a person type) as the request body, had a double-underscore typo in the success message key, and invalidated the wrong cache key.
- Fixed to call `PATCH /api/agent/:agentId` which accepts `about` and `website`. The organisation's numeric ID is not exposed in the current agent API response, so the `address` field cannot yet be saved — this needs a separate BE change to expose the org ID.

## Test plan

- [ ] Edit agent contact details (name, role, email, phone) — profile refreshes after save
- [ ] Edit organisation details (about, website) — profile refreshes after save showing updated values
- [ ] Confirm address field still displays correctly (read-only for now, pending BE change)

### DIFF
--- a/src/components/Dashboard/Profile/sections/OrganisationDetails/OrganisationDetails.tsx
+++ b/src/components/Dashboard/Profile/sections/OrganisationDetails/OrganisationDetails.tsx
@@ -24,7 +24,7 @@ export const OrganisationDetails = forwardRef<EditableSectionRef, Props>(functio
 ) {
   const { t, i18n } = useTranslation();
   const [isEditing, setIsEditing] = useState(false);
-  const { mutate: updateOrganization /*, isPending */ } = useUpdateOrganization(String(agent?.operator));
+  const { mutate: updateOrganization /*, isPending */ } = useUpdateOrganization(String(agent?.id));
 
   useEditingChangeNotifier(isEditing, onEditingChange);
   const { data: apiLanguages } = useApiLanguages();
@@ -56,12 +56,15 @@ export const OrganisationDetails = forwardRef<EditableSectionRef, Props>(functio
   };
 
   const onSubmit = (values: OrganisationDetailsFormData) => {
-    updateOrganization(values as unknown as ApiAgentProfileGet, {
-      onSuccess: () => {
-        reset(values);
-        setIsEditing(false);
+    updateOrganization(
+      { about: values.about, website: values.website },
+      {
+        onSuccess: () => {
+          reset(values);
+          setIsEditing(false);
+        },
       },
-    });
+    );
   };
 
   return (
--- a/src/hooks/useUpdateOrganizationDetails.ts
+++ b/src/hooks/useUpdateOrganizationDetails.ts
@@ -1,14 +1,16 @@
 import { ApiAgentProfileGet } from "@/components/Dashboard/Profile/types";
-import { apiPathOrganization } from "@/config/constants";
+import { apiPathAgent } from "@/config/constants";
 import { useMutationQuery } from "@/hooks";
-import { ApiRepresentativeGet } from "need4deed-sdk";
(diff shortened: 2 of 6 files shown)


lourooo:
Great fixes, all bugs correctly identified and the solutions are well-aligned with the rest of the codebase.

One thing worth flagging: OrganisationDetailsFormData has fields like address, organizationType, operator, services, and clientLanguages marked as required in the schema, but ApiAgentPatch only accepts about and website. 

A user could edit those fields, pass validation, and save but the data would silently not be persisted. Worth either making them read-only in edit mode or removing the required validation for anything not being sent to the API.

Cy-fox:
Might be worth adding a quick guard before the mutation fires if agent?.id is missing. Right now the component uses String(agent?.id) which can happily turn undefined into the literal string "undefined", and the API call ends up hitting /agent/undefined. A small early return or a if (!id) check in the hook would keep things clean and avoid those weird requests showing up in the network tab.

src/components/Dashboard/Profile/sections/OrganisationDetails/OrganisationDetails.tsx
```diff
@@ -24,7 +24,7 @@ export const OrganisationDetails = forwardRef<EditableSectionRef, Props>(functio
 ) {
   const { t, i18n } = useTranslation();
   const [isEditing, setIsEditing] = useState(false);
-  const { mutate: updateOrganization /*, isPending */ } = useUpdateOrganization(String(agent?.operator));
+  const { mutate: updateOrganization /*, isPending */ } = useUpdateOrganization(String(agent?.id));
```
DarrellRoberts, line 27: if `isPending` is no longer needed, then can remove the comment

src/hooks/useUpdateOrganizationDetails.ts
```diff
@@ -1,14 +1,16 @@
...
-    apiPath: `${apiPathOrganization}${organizationId}`,
+// Patches agent-level org details (about, website) via the agent endpoint.
+// Note: address and district require the organization's numeric ID which is not
+// currently exposed in the agent API response — those fields need a BE change.
+export const useUpdateOrganization = (agentId: string) => {
+  return useMutationQuery<Pick<ApiAgentPatch, "about" | "website">, ApiAgentProfileGet>({
+    apiPath: `${apiPathAgent}/${agentId}`,
     method: "patch",
-    successMessage: "__dashboard.agentProfile.contactDetails.saveSuccess",
-    queryKeyToInvalidate: ["organization", organizationId],
```
DarrellRoberts, line 12: I think I mentioned this during the `fe` refactor but the `organization` label confuses me. Would have said it made me more sense to call this hook `useUpdateAgent` but in the `/Profile/sections` dir we already have Organization components so I see why it's named as so

DarrellRoberts (approved):
looks good as far as I can tell. Currently in my dev environment I can't test the hooks as can't find an Agent with a representative attached (which I know is another issue). 

Wrote a couple of comments

PR #434, opened 2026-05-04
Fix: Creating an Event page #287

### LINKED ISSUE #287: Creating an Event page
### Description

Create event registration page derived from the event icon on the left menue on the dashbord. 

Image

https://www.figma.com/design/9Kl17OiDLbWoNxgY43Rgjk/Need4Deed-Designs-2025?node-id=15071-32477&t=xqtnjEppIO1EcUW0-4

### Proposed solution

_No response_

lourooo:
## Description
Adds a calendar page to the dashboard, accessible from the calendar icon in the left navigation bar. The page displays a monthly calendar view with day selection and a "Create an event" button that leads to a new multi-step event creation form.

## Related Issues
Closes #287 

## Changes
New `/dashboard/calendar` page with a monthly calendar grid, prev/next month navigation, and clickable day selection
- "Create an event" button in the top bar, if a day is selected, its date is pre-filled in the form
- New `/dashboard/calendar/create` multi-step form with 3 steps:
  - Step 1 — Event title and description
  - Step 2— Location (street, house number, postcode)
  - Step 3 — Date and time
- Progress bar at the top of the creation form reflecting the current step
- i18n support for both `en` and `de`

## Screenshots / Demos
image
image
image
image

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes


nadavosa:
## Review — Creating an Event page (#287)

### Critical — No API integration
`StepDateTime`'s submit button is `onClick={() => {}}` — it does nothing. The form collects data across 3 steps but never submits to the BE. The BE endpoint for events (`POST /event`) is tracked in BE #458. This PR is a UI shell and should not merge until the submit handler is wired up.

### Hardcoding — does not follow Hans's pattern
The entire codebase uses CSS variables for every dimension. This PR introduces a large number of raw pixel values:

**`CalendarContent.tsx`**
- `padding: 40px 48px 100px` → should use `var(--spacing-*)` tokens
- `gap: 110px` in `TopRow` → unexplained magic number, must be a CSS variable
- `width: 560px` in `CalendarBox` → hardcoded width
- `width: 40px; height: 40px` in `DayCell` → hardcoded
- `font-size: 20px`, `font-size: 14px` → use `var(--text-*)` tokens
- `border: 2px solid` → use `var(--border-width-*)`
- `var(--border-radius-lg, 16px)` and `var(--spacing-4, 4px)` — the pixel fallbacks suggest the tokens may not exist; confirm and define them instead of inlining fallbacks

**`ActionButton.tsx`**
- `padding: 12px 32px`, `border-radius: 100px`, `font-size: 16px`, `font-weight: 600`, `border: 2px solid` → all raw values
- The existing `Button` component is already used in `CalendarContent.tsx`. `ActionButton` appears to be a duplicate — use `Button` throughout or extract a variant, don't create a parallel component

**`styles.ts`**
- `font-size: 14px / 12px / 16px / 18px`, `font-weight: 600`, `border: 1.5px solid`, `border-radius: 8px`, `padding: 12px 16px`, `min-width: 160px`, `height: 6px`, `border-radius: 3px` → all should use design tokens

### Naming collision
`ActionButton` is already exported from `src/components/core/common/Table/styles.ts` (Hans's table component). Introducing another `ActionButton` in the Calendar directory creates confusion and import ambiguity.

### Locale issue — calendar week starts Sunday
`DAY_LABELS = ["S", "M", "T", "W", "T", "F", "S"]` and `getDay()` (which returns 0 for Sunday) means the grid starts on Sunday. Germany uses Monday as the first day of the week. This is wrong for the primary audience. Use `Date.toLocaleDateString` or adjust the `firstDow` offset.

### i18n — day labels are hardcoded English
`DAY_LABELS` is hardcoded English. Day abbreviations should come from `toLocaleString(i18n.language, { weekday: 'narrow' })` — the same pattern already used for `monthLabel`.

### Summary
Needs BE integration before merge, all pixel values replaced with design tokens, `ActionButton` deduplicated, and the week-start and day-label issues fixed.

nadavosa (changes requested):
Thanks for the contribution — the feature is well-structured overall and the component split, i18n coverage, and styled-component patterns are clean. There are a few issues that need to be fixed before this can merge.

---

## Critical: Pre-filled date is never read from the query string

The PR description states: *"if a day is selected, its date is pre-filled in the form"*. `CalendarContent` correctly serialises the selected day into a `?date=` query param when navigating to `/create`, but `CreateEvent` never reads that param back. `useSearchParams` is not imported or called anywhere in `CreateEvent.tsx`, so the date field is always initialised to `""` regardless of which day was clicked.

Fix: read the param on mount and use it to initialise `formData.date`.

```tsx
// CreateEvent.tsx
import { useSearchParams } from "next/navigation";

// inside CreateEvent()
const searchParams = useSearchParams();
const prefilledDate = searchParams.get("date") ?? "";

const [formData, setFormData] = useState<EventFormData>({
  ...
  date: prefilledDate,
  ...
});
```

---

## Bug: `buildCalendarCells` — trailing-days row is always added when the grid is already full

```ts
const remaining = 7 - (cells.length % 7);
if (remaining < 7) {
  for (let d = 1; d <= remaining; d++) {
    cells.push({ day: d, currentMonth: false });
  }
}
```

When `cells.length` is already a multiple of 7 (which happens whenever the first day of the month lands exactly after filling the previous-month padding, e.g. February 2026), `cells.length % 7 === 0`, so `remaining = 7 - 0 = 7`. The `remaining < 7` guard is then **false**, so no extra row is appended — this part is fine. But consider a month that ends mid-week: the condition fires and appends trailing days correctly. However the guard `remaining < 7` should arguably be `remaining !== 7` or `remaining > 0` — using `< 7` is equivalent in practice but is semantically confusing. More importantly: test with a month whose last day falls on a Sunday (e.g. March 2025 ends on a Sunday): `cells.length % 7 === 0`, `remaining === 7`, guard is false — **no trailing row is appended, which is correct**. This is actually fine, but worth a comment explaining the logic.

This is a low-severity nit; just add a brief comment so future readers aren't confused.

---

## Missing `"use client"` directive in step components

`StepDateTime.tsx`, `StepLocation.tsx`, and `StepTitle.tsx` contain interactive `onChange` handlers and call `t()` from `react-i18next`, yet none of them declare `"use client"`. `CreateEvent.tsx` is `"use client"` and imports them, so they inherit the client bundle, but it is better practice — and more consistent with the rest of the codebase (see `CalendarContent.tsx`, `Calendar.tsx`) — to add `"use client"` to each file explicitly so the intent is clear.

---

## `StepTitle`: `FieldLabel` is used as a visible label but its `htmlFor` duplicates the accessible name

In `StepTitle.tsx`:
```tsx
<Heading3>{t("dashboard.calendar.createForm.titleLabel")}</Heading3>
<FieldLabel htmlFor="event-title">{t("dashboard.calendar.createForm.titlePlaceholder")}*</FieldLabel>
<StyledInput
  id="event-title"
  placeholder={t("dashboard.calendar.createForm.titlePlaceholder")}
  ...
/>
```

The `FieldLabel` text and `placeholder` are the same string (`"Title"`). This means the visible label and the placeholder both say "Title", which is redundant and may confuse users. The placeholder can be left as a hint (`titleHelper`), and the label should use `titleLabel` (the more descriptive string) or vice-versa. Additionally the asterisk `*` for required state is only visible — there is no `aria-required="true"` on the input and no `aria-label` that includes the asterisk.

---

## Hard-coded pixel bottom padding (`100px`) repeated in two components

`CalendarContent.tsx` and `CreateEvent.tsx` both use a literal `100px` bottom padding:
```ts
padding: var(--spacing-32) var(--spacing-48) 100px;
```
This should use a CSS custom property (or the existing `--spacing-*` scale) to remain consistent with the design system. If this is a workaround for a fixed footer, it should at least be extracted to a shared constant.

---

## `handleSubmit` leaves a `console.log` in production-bound code

```ts
const handleSubmit = async () => {
  console.log("Submit event:", formData);   // ← must be removed
  router.push(`/${i18n.language}/dashboard/calendar`);
};
```

The TODO comment referencing BE https://github.com/need4deed-org/fe/issues/458 is helpful context, but the `console.log` itself should not be in the PR. Remove it or replace with a proper loading/error state placeholder.

---

## `CalendarBox` has a hard-coded pixel width

```ts
const CalendarBox = styled.div`
  width: 560px;
  ...
`;
```

All other dimension tokens in the codebase use CSS custom properties or `%`/`max-width`. This should be `max-width: 560px; width: 100%;` to avoid overflow on smaller viewports (the calendar is otherwise unresponsive on mobile screens).

---

## Minor / nit: `t` prop drilling is unnecessary

All three step components receive `t` as a prop, but `useTranslation()` is already available as a hook. Passing `t` as a prop is an anti-pattern in this codebase — other components (e.g. `Opportunities.tsx`) call `useTranslation()` directly inside each component. Remove the `t` prop from the step interfaces and call `useTranslation()` at the top of each step instead.

---

## Summary

| Severity | Issue |
|---|---|
| Critical | Pre-filled date from `?date=` query param is never read in `CreateEvent` |
| Bug | `console.log` left in `handleSubmit` |
| Bug | Hard-coded `560px` width breaks mobile layout |
| Quality | `"use client"` missing on step components |
| Quality | `t` passed as prop instead of using the hook |
| Nit | `100px` literal padding should use a CSS variable |
| Nit | Redundant label/placeholder text + missing `aria-required` in `StepTitle` |

Please address at minimum the critical and bug-level items before merging.

lourooo:
Updated

nadavosa (commented):
## Follow-up

Good progress — most of the issues from the previous review are fixed:

- ✅ `useSearchParams` now reads the `?date=` param and pre-fills `formData.date`
- ✅ `console.log` removed from `handleSubmit`, TODO comment references BE https://github.com/need4deed-org/fe/issues/458
- ✅ `CalendarBox` is now `max-width: 560px; width: 100%` — responsive on small screens
- ✅ `ActionButton` removed, uses the shared `Button` component throughout
- ✅ Week starts Monday (`(rawDow + 6) % 7` is correct)
- ✅ Day labels use locale (`getDayLabels(i18n.language)`)
- ✅ `"use client"` on all step components
- ✅ `t` prop drilling removed — each step calls `useTranslation()` directly
- ✅ `aria-required="true"` on the title input
- ✅ Bottom padding now `var(--spacing-48)`

Three small hardcoded pixel values remain — please replace with CSS tokens:

**`CalendarContent.tsx` — `DayCell`:**
```ts
width: 40px;   // → var(--spacing-40) or equivalent
height: 40px;  // → var(--spacing-40) or equivalent
```

**`CreateEvent.tsx` — `ProgressSegment`:**
```ts
height: 6px;        // → var(--spacing-6) or var(--border-width-*)
border-radius: 3px; // → var(--border-radius-xs) or similar
```

**`styles.ts` — `FieldRowLabel`:**
```ts
min-width: 160px; // → var(--spacing-160) or a layout-specific token
```

One structural nit in `StepTitle`: there is a `<Heading3>` immediately above `<FieldLabel>` and both render the same `titleLabel` string. The `Heading3` is redundant — the `FieldLabel` (linked to the input via `htmlFor`) is the correct accessible label element. Remove the `Heading3` for that field group so the section heading and the input label are not duplicated.

Fix those four items and this is ready to merge.

nadavosa (approved):
All 4 issues fixed: duplicate Heading3 removed, DayCell uses --spacing-40, ProgressSegment uses component tokens, FieldRowLabel uses --create-event-field-label-min-width.

PR #435, opened 2026-05-04
fix: add Inactive option to opportunity status filter

### LINKED ISSUE #431: Opportunities dashboard: "Inactive" status missing from the status filter
### Summary

The status filter on the Opportunities dashboard does not include an "Inactive" option. Coordinators have no way to filter the list to show only inactive opportunities. Inactive opportunities are visible in the list (the status tag renders, though its i18n key may also be missing), but cannot be targeted via the filter panel.

### Steps to Reproduce

1. Open the Opportunities dashboard (`/dashboard/opportunities`)
2. Open the Filters panel
3. Expand the Status section
4. Observe available options

### Expected Behavior

"Inactive" should appear as a selectable status filter option alongside "New", "Searching", and "Active".

### Actual Behavior

"Inactive" is not listed. There is no way to filter for inactive opportunities.

Image

nadavosa:
Closes #431

## Summary
- Added `OpportunityStatusType.INACTIVE` to the opportunity filter default state
- Added `"opp-inactive"` translation key to EN (`"Inactive"`) and DE (`"Inaktiv"`) locales under `dashboard.opportunities.filters.status`

The serializer/deserializer already handles any key in `filter.status` dynamically, so no other changes needed.

### DIFF
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -746,6 +746,7 @@
           "opp-new": "Neu",
           "opp-searching": "Suche läuft",
           "opp-active": "Aktiv",
+          "opp-inactive": "Inaktiv",
           "opp-past": "Vergangen"
         },
         "schedule": {
--- a/public/locales/en/translations.json
+++ b/public/locales/en/translations.json
@@ -789,6 +789,7 @@
           "opp-new": "New",
           "opp-searching": "Searching",
           "opp-active": "Active",
+          "opp-inactive": "Inactive",
           "opp-past": "Past"
         },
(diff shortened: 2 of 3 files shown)


ivannissimrch (commented):
I tested locally, and the FE fix works correctly. The Inactive option shows in the filter and sends opp-inactive to the BE. The only remaining piece is that the BE needs opp-inactive added to the OpportunityStatusType enum in be/src/server/schema/sdk-types.json. Currently, Fastify rejects it with a 400 validation error. Once that's added, this is fully working end-to-end.

Issue #430, opened 2026-04-30
Volunteers dashboard: selecting "Temporarily Unavailable" in the Engagement filter returns 0 results


Cy-fox:
### Summary

On the Volunteers dashboard, checking the "Temporarily Unavailable" option under the Engagement filter causes the results list to show 0 volunteers, even though volunteers with that status exist. A filter chip appears confirming the selection, but no results are returned.

### Steps to Reproduce

1. Open the Volunteers dashboard (`/dashboard/volunteers`)
2. Open the filters panel
3. Expand the Engagement section
4. Check "Temporarily Unavailable"
5. Observe the results

### Expected Behavior

Volunteers whose engagement status is "Temporarily Unavailable" should be displayed.

### Actual Behavior

0 volunteers are shown. The filter chip appears correctly but the API returns no results.

<img width="1125" height="714" alt="Image" src="https://github.com/user-attachments/assets/7e6f03f5-53fb-4467-9c40-c923ac48cda2" />

lourooo:
I can take care of this

PR #466, opened 2026-05-04
fix: 'Temporarily Unavailable' engagement filter not working

nadavosa:
## Summary
- The BE prepends `vol-` to filter keys, so `temporarilyUnavailable` became `vol-temporarilyUnavailable` which doesn't match the enum value `vol-temp-unavailable`
- All other engagement filter keys (`new`, `active`, `available`, `inactive`, `unresponsive`) happened to produce correct `vol-*` values
- Fix: rename the filter key to `temp-unavailable` so `vol-temp-unavailable` matches the enum
- Old translation key kept alongside new one to avoid breaking any existing references

Fixes https://github.com/need4deed-org/fe/issues/430

## Test plan
- [ ] Open volunteer filter panel, select "Temporarily Unavailable" — list should filter correctly
- [ ] Verify other engagement filters still work

### DIFF
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -356,6 +356,7 @@
           "new": "Neu",
           "available": "Verfügbar",
           "temporarilyUnavailable": "Vorübergehend nicht verfügbar",
+          "temp-unavailable": "Vorübergehend nicht verfügbar",
           "unresponsive": "Reagiert nicht"
         },
         "preferredAv": {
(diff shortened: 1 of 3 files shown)


public/locales/de/translations.json
```diff
@@ -356,6 +356,7 @@
           "new": "Neu",
           "available": "Verfügbar",
           "temporarilyUnavailable": "Vorübergehend nicht verfügbar",
```
arturasmckwcz, line 358: looks like this has become a garbage

public/locales/en/translations.json
```diff
@@ -356,6 +356,7 @@
           "new": "New",
           "available": "Available",
           "temporarilyUnavailable": "Temporarily Unavailable",
```
arturasmckwcz, line 358: the same as in `de`

PR #467, opened 2026-05-04
fix: filters reset when navigating back from volunteer/opportunity/agent profile

### LINKED ISSUE #356: Dashboard filters reset when returning from Opportunity/Volunteer Profile
### Summary

When applying filters on the Dashboard (Opportunities or Volunteers) and then opening a specific Opportunity or Volunteer Profile, the filters are lost upon returning to the Dashboard. The filter state is not saved during navigation.

### Steps to Reproduce

.

### Expected Behavior

The previously applied filters should still be active when returning to the Dashboard from a Profile page.

### Actual Behavior

All filters are reset/removed when returning to the Dashboard. User must re-apply filters every time they view a profile and go back.

https://github.com/user-attachments/assets/d19fc20b-e1df-4fb3-bc27-117cf7238357

nadavosa:
## Summary
- "Back to dashboard" used `href="."` which navigates to the parent path with no query params, discarding all active filter state
- Fix: intercept the click with `router.back()` so the browser restores the previous URL (including filter params)
- Falls back to `href="."` if there is no browser history (e.g. user navigated directly to a profile URL)
- Applies to all profile types: volunteer, opportunity, agent (all share `ProfilePage`)

Fixes https://github.com/need4deed-org/fe/issues/356

## Test plan
- [ ] Set one or more filters on the volunteers page
- [ ] Click a volunteer card to open the profile
- [ ] Click "Back to dashboard" — filters should still be active
- [ ] Repeat for opportunities and agents

### DIFF
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -356,6 +356,7 @@
           "new": "Neu",
           "available": "Verfügbar",
           "temporarilyUnavailable": "Vorübergehend nicht verfügbar",
+          "temp-unavailable": "Vorübergehend nicht verfügbar",
           "unresponsive": "Reagiert nicht"
         },
         "preferredAv": {
--- a/public/locales/en/translations.json
+++ b/public/locales/en/translations.json
@@ -356,6 +356,7 @@
           "new": "New",
           "available": "Available",
           "temporarilyUnavailable": "Temporarily Unavailable",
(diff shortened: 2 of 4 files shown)


nadavosa:
Closing — the back-navigation fix was superseded by PR https://github.com/need4deed-org/fe/pull/497 which has been merged. The `temp-unavailable` filter key rename should be tracked separately if still needed.

PR #468, opened 2026-05-04
Fix language lookup by ID when saving opportunity details

### LINKED ISSUE #463: bug: opportunity profile — languages and activities not saved correctly in opportunity details
Editing opportunity details (e.g. languages, activities) in the opportunity profile does not persist correctly. Changes appear to save but are either not reflected on reload or are partially lost.

Cy-fox:
Closes #463

When a user changes a language in the opportunity details form, `LanguageFieldRow` stores the selected value as a numeric ID string (e.g. "5"). The previous lookup in `toLangOptionItems` only matched by translated name, so any language the user changed was silently dropped from the PATCH payload. This fix adds an ID-based lookup first so both cases (initial-load translated names and user-selected IDs) resolve correctly to the right OptionItem. Also removes references to `OpportunityStatusType.INACTIVE` which was dropped from the SDK (PAST is the equivalent).

Changes:
- OpportunityDetailsEdit.tsx: added numeric ID lookup before name-based lookup in `toLangOptionItems` so changed languages are included in the save payload
- Filters/constants.ts, OpportunityCard.helpers.tsx, statusMaps.ts, ProfileHeader/opportunity/constants.ts: removed `OpportunityStatusType.INACTIVE` (SDK breaking change), map manual INACTIVE status to PAST

How to test:
1. Open an opportunity profile as an agent
2. Click Edit on the Opportunity Details section
3. Change one of the main communication languages to a different language
4. Save
5. Verify the new language appears after saving (previously it would revert to the original)
6. Repeat for residents speak languages


nadavosa (commented):
The language ID lookup fix in `OpportunityDetailsEdit.tsx` is correct and should be kept — good catch.

However, the other changes need to be reverted:

- `OpportunityCard.helpers.tsx` — removes `INACTIVE` from `Record<OpportunityStatusType, ...>` maps → causes **TS2741** lint failure (`Property '[OpportunityStatusType.INACTIVE]' is missing in type ... but required in type 'Record<OpportunityStatusType, string>'`)
- `statusMaps.ts` — same issue, removes INACTIVE from color/icon maps
- `src/components/Dashboard/Volunteers/Filters/constants.ts` — removes `[OpportunityStatusType.INACTIVE]: false` — this undoes the work from PR https://github.com/need4deed-org/fe/pull/435 which added INACTIVE to the filters
- `ProfileHeader/opportunity/constants.ts` — maps `INACTIVE → PAST` which is semantically wrong

Please revert those 4 files and keep only the `OpportunityDetailsEdit.tsx` change. The CI failure is caused by the INACTIVE removals.

nadavosa:
## Follow-up Review

The `OpportunityDetailsEdit.tsx` language ID fix remains correct and well-structured — the numeric ID path is now handled before the name-matching fallback, which is the right order.

The other four files still need to be reverted. Specifically:

- **`Filters/constants.ts`** — removing `[OpportunityStatusType.INACTIVE]: false` undoes PR https://github.com/need4deed-org/fe/pull/435. `INACTIVE` is a valid SDK enum value and must remain in the filter map.
- **`statusMaps.ts`** — `INACTIVE` must remain in both `statusColorMap` and `statusIconMap`. These are typed as `Record<StatusValue, ...>` which requires full enum coverage — removing a key is a TypeScript error.
- **`ProfileHeader/opportunity/constants.ts`** — mapping `INACTIVE → PAST` in `SDK_TO_MANUAL` and `MANUAL_TO_SDK` is semantically wrong. `INACTIVE` and `PAST` are distinct states.

Please revert those four files to their state before this PR and keep only the `OpportunityDetailsEdit.tsx` change.

nadavosa (commented):
## Follow-up

The four extra files have been reverted — only `OpportunityDetailsEdit.tsx` remains in the diff. The ID-based lookup fix is clean and the comment explains the two lookup paths clearly. LGTM, this is ready to merge.

nadavosa (approved):
LGTM — clean fix, only OpportunityDetailsEdit.tsx changed.

PR #475, opened 2026-05-05
Fix: Volunteers dashboard: selecting "Temporarily Unavailable" in the Engagement filter returns 0 results #430

### LINKED ISSUE #430: Volunteers dashboard: selecting "Temporarily Unavailable" in the Engagement filter returns 0 results

lourooo:
Fix: Volunteers dashboard: selecting "Temporarily Unavailable" in the Engagement filter returns 0 results #430

## Description
The "Temporarily Unavailable" engagement filter was sending temporarilyUnavailable to the API. The backend has a workaround that prepends vol- to all engagement filter values before querying the database, so it was looking for vol-temporarilyUnavailable which doesn't exist, returning 0 results. The fix aligns the filter key with what the backend expects.

## Related Issues
Closes #430 

## Changes
- Fix `temporarilyUnavailable` filter key to `temp-unavailable` so the backend correctly resolves it to `vol-temp-unavailable`
- Update EN and DE translation keys accordingly
- Fix layout shift in the volunteer cards list during loading by adding `flex: 1` to `VolunteerCardListContainer`

## Screenshots / Demos
image

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/src/components/Dashboard/Volunteers/Filters/constants.tsx
+++ b/src/components/Dashboard/Volunteers/Filters/constants.tsx
@@ -1,4 +1,4 @@
-import { ByDay, OccasionalType, QueryParamsKeys, TimeSlot } from "need4deed-sdk";
+import { ByDay, OccasionalType, QueryParamsKeys, TimeSlot, VolunteerStateEngagementType } from "need4deed-sdk";
 import { CardsFilter } from "./types";
 
 export const defaultVolunteerCardsFilter: CardsFilter = {
@@ -7,12 +7,12 @@ export const defaultVolunteerCardsFilter: CardsFilter = {
   [QueryParamsKeys.DISTRICT]: {},
   [QueryParamsKeys.LANGUAGE]: {},
   [QueryParamsKeys.ENGAGEMENT]: {
-    new: false,
-    active: false,
-    available: false,
-    "temp-unavailable": false,
-    inactive: false,
-    unresponsive: false,
+    [VolunteerStateEngagementType.NEW]: false,
+    [VolunteerStateEngagementType.ACTIVE]: false,
+    [VolunteerStateEngagementType.AVAILABLE]: false,
+    [VolunteerStateEngagementType.TEMP_UNAVAILABLE]: false,
+    [VolunteerStateEngagementType.INACTIVE]: false,
+    [VolunteerStateEngagementType.UNRESPONSIVE]: false,
   },
   [QueryParamsKeys.AVAILABILITY]: {
     times: {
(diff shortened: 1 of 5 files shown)


src/components/Dashboard/Volunteers/Filters/constants.tsx
```diff
@@ -10,7 +10,7 @@ export const defaultVolunteerCardsFilter: CardsFilter = {
     new: false,
     active: false,
     available: false,
-    temporarilyUnavailable: false,
+    "temp-unavailable": false,
```
DarrellRoberts: it's not your fault at all, but not sure why we're using a different pattern here than in `Opportunities`. 
I think it would be better to use the sdk types. It'd prevent further disconnects between the `fe` and `be`.
See [here](https://github.com/need4deed-org/fe/blob/develop/src/components/Dashboard/Opportunities/Filters/constants.ts) to see what I mean.

Image

This is also reflective in the i18n for Opportunities:

Image

DarrellRoberts (changes requested):
A good approach you used but I think this would be a better opportunity to align the constants with the `sdk` similar to how the constants in `Opportunities` are built.

lourooo:
Updated

src/components/Dashboard/Volunteers/helpers.ts
```diff
@@ -107,10 +107,11 @@ export function serializeFilters(
   });
 
   // 2. Clear all existing 'engagement' params
+  // Strip the "vol-" prefix because the backend's engagementWorkaround re-adds it
   params.delete(QueryParamsKeys.ENGAGEMENT);
   Object.entries(filter.engagement).forEach(([key, value]) => {
     if (value === true) {
-      params.append(QueryParamsKeys.ENGAGEMENT, key);
+      params.append(QueryParamsKeys.ENGAGEMENT, key.replace(/^vol-/, ""));
```
DarrellRoberts, line 114: ah it would appear the `be` handles filters from opportunities and volunteers differently. Nonetheless, great workaround!

DarrellRoberts (approved):
thank you!

PR #480, opened 2026-05-06
Add address field to agent contact details form

Cy-fox:
Closes #354

The agent contact details edit form was missing the address field entirely. The address is stored on the representative (person) and the existing PATCH /person/:id endpoint already handles it via dtoParsePerson, so this is a pure frontend fix.

Changes:
- Add agentAddressUtils.ts with formatAddress and parseAddress helpers, mirroring the volunteer contact details pattern
- Extend agentContactDetailsSchema with an optional address string field
- Memoize initialFormValues and sync form on agent data refresh via useEffect (same pattern as VolunteerContactDetails)
- Parse address string and include the structured address object in the PATCH /person/:id payload on submit
- Add address Controller to AgentContactDetailsEdit for editing
- Show address field in AgentContactDetailsDisplay
- Add "address" label key to EN and DE translation files

How to test:
1. Log in as a coordinator and navigate to an agent profile
2. Open the Contact Details section and verify the address is displayed
3. Click Edit and verify the address field appears pre-filled
4. Change the address (format: "Street, City, Postcode") and save
5. Verify the updated address is shown after saving

PR #481, opened 2026-05-06
Show district in agent profile header

### LINKED ISSUE #458: bug: agent profiles missing address and district information
Agent profiles do not show address or district, making it impossible to find contact information when the opportunity card is also missing details. All agent profiles should include district, address, and relevant contact info where available.

Cy-fox:
Closes #458

Agent profiles were not displaying the district field anywhere on the page. The district is available on the API response (agent.district) but was never rendered in the UI, making it impossible to see which area an agent serves.

Note: address visibility is covered by PR #480 (AgentContactDetailsDisplay now shows the address field).

Changes:
- Add a District row to AgentHeader using StatusRowField, rendering the title in the current UI language with EN fallback
- Import Lang type from need4deed-sdk for type-safe language key access

How to test:
1. Log in as a coordinator and navigate to an agent profile that has a district assigned
2. Verify the District row appears in the header section below Trust Level
3. Switch the UI language between EN and DE and verify the district title updates accordingly
4. Navigate to an agent profile with no district set and verify the row shows the empty placeholder

### DIFF
--- a/src/components/Dashboard/Profile/sections/ProfileHeader/agent/AgentHeader.tsx
+++ b/src/components/Dashboard/Profile/sections/ProfileHeader/agent/AgentHeader.tsx
@@ -2,7 +2,6 @@
 import { EMPTY_PLACEHOLDER_VALUE } from "@/config/constants";
 import { formatDateTime } from "@/utils";
 import { HouseIcon } from "@phosphor-icons/react";
-import { Lang } from "need4deed-sdk";
 import { useTranslation } from "react-i18next";
 import { useUpdateAgentStatus } from "@/hooks";
 import { AgentTrustLevel, ApiAgentProfileGet } from "../../../types/agent";
@@ -12,6 +11,7 @@ import {
   createEngagementStatusLabelMap,
   createTrustLevelLabelMap,
   createVolunteerSearchLabelMap,
+  getDistrictLabel,
   TRUST_LEVEL_OPTIONS,
 } from "./constants";
 import { TrustLevelDropdown } from "./TrustLevelDropdown";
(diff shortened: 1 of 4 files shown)


src/components/Dashboard/Profile/sections/ProfileHeader/agent/AgentHeader.tsx
```diff
@@ -72,6 +73,18 @@ export const AgentHeader = ({ agent }: Props) => {
           />
         }
       />
+
```
DarrellRoberts, line 77: here you can take advantage of the `/constants.ts` file and create a formatting function like so: 
```typescript
// /constants.ts
export const getDistrictLabel = (district: ApiAgentProfileGet["district"], language: string): string => {
  const { title } = district || {};

  return title?.[language as Language] || title?.en || EMPTY_PLACEHOLDER_VALUE;
};
```

Then in `/agents/AgentHeader.tsx` all you need to do is import the function, capture its value and pass it as a prop
```typescript
  const districtLabel = getDistrictLabel(agent.district, i18n.language);
// in the tsx
      <StatusRowField title={t("dashboard.agentProfile.district")} showIcon={false} extra={districtLabel} />
```

What do you think?

DarrellRoberts (changes requested):
looks good, just one suggested change to make it neater

src/components/Dashboard/Profile/sections/ProfileHeader/agent/AgentHeader.tsx
Cy-fox: Good call, done. Moved the logic into `getDistrictLabel` in constants.ts and simplified the component.

DarrellRoberts (approved):
thanks! Just need to resolve some merge conflicts

PR #497, opened 2026-05-07
Fix: bug: dashboard list pagination resets to page 1 when returning from a profile #487

### LINKED ISSUE #487: bug: dashboard list pagination resets to page 1 when returning from a profile
When navigating to a volunteer, opportunity, or agent profile from a paginated list (e.g. page 3) and then going back, the list resets to page 1 instead of returning to the page the user was on.

**Steps to reproduce:**
1. Navigate to any paginated list (volunteers, opportunities, or agents)
2. Go to page 2 or higher
3. Open any item profile
4. Use the browser back button or in-app navigation to return to the list

**Expected behavior:** The list should restore the previously selected page so the user can continue where they left off.

**Current behavior:** The list always resets to page 1.

**Reporter:** Senya (06.05.2026)

lourooo:
Fix: bug: dashboard list pagination resets to page 1 when returning from a profile #487

## Description
Page number was stored in component state, so navigating to a profile and returning always reset the list to page 1. Page is now persisted in the URL (?page=X), so both the browser back button and the in-app "Back to Dashboard" button restore the correct page.

## Related Issues
Closes #487 

## Changes
Added usePageParam hook to read/write the current page as a URL search param

Replaced useState(1) with usePageParam() in OpportunityListController, VolunteerListController, and AgentListController

Added params.delete("page") to the three filter serializers so changing filters always resets to page 1

Changed "Back to Dashboard" to use router.back() instead of href="." so it restores the full previous URL including the page param

## Screenshots / Demos

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/src/components/Dashboard/Agents/AgentListController.tsx
+++ b/src/components/Dashboard/Agents/AgentListController.tsx
@@ -1,7 +1,7 @@
 import { ApiAgentGetList, ApiOptionLists, SortOrder } from "need4deed-sdk";
 import { AgentCardList } from "./AgentCardList";
-import { useEffect, useState } from "react";
-import { useGetQuery } from "@/hooks";
+import { useEffect } from "react";
+import { useGetQuery, usePageParam } from "@/hooks";
 import { apiPathAgent, cacheTTL } from "@/config/constants";
 import { serializeAgentFilters } from "./helpers";
 import { AgentCardsFilter } from "./Filters/types";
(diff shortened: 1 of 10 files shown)


nadavosa:
## Review

**Overview:** Fixes pagination reset on back-navigation by persisting the current page in the URL (`?page=X`) via a new `usePageParam` hook, and fixes the back button in `ProfilePage` to use `router.back()` instead of `href="."`.

### What's good
- `usePageParam` is clean and well-structured. `useCallback` deps are correct.
- `params.delete("page")` in all three serialize-filter helpers correctly resets to page 1 when filters change.
- `router.back()` is the right fix — `href="."` always navigated to the base path, discarding the stored `?page=X`.
- `backNavStyles` extraction in `styles.tsx` is a clean refactor to share styles between `BackLink` and `BackButton`.

### One suggestion
`setCurrentPage` uses `router.push`, which adds a history entry on every page click. This means if a user pages through 1 → 2 → 3, pressing the browser back button takes them through all three pages before leaving the list. Consider `router.replace` instead — it keeps the URL in sync without polluting the history stack, and `router.back()` from a profile would still land on the last visited page since that page was the one pushed when navigating into the profile.

Otherwise LGTM.

nadavosa (approved):

PR #498, opened 2026-05-07
Fix: skill field required when volunteer has no skill #485

### LINKED ISSUE #485: bug: volunteer profile skill field required when volunteer has no skill
When a coordinator edits information in a volunteer profile (e.g. volunteering type), the skill field is marked as required. This forces the coordinator to select a skill even when the volunteer has not chosen one, leading to incorrect data being saved.

**Steps to reproduce:**
1. Open a volunteer profile that has no skill set
2. Enter edit mode for any section that includes the skill field
3. Attempt to save without selecting a skill
4. The form blocks submission, requiring a skill to be selected
**Expected behavior:** Skill should be optional. If the volunteer has not set a skill, the field should accept an empty/unset value and allow saving.

**Reporter:** Kenza (05.05.2026)

Cy-fox:
Closes #485

The `skills` field in the volunteer profile form had a `min(1)` Zod validation, making it required. This blocked coordinators from saving any profile section when the volunteer had no skill set, even though the skill attribute is optional.

Removing the `min(1)` constraint allows the form to submit with an empty skills array.

Changes:
- Remove `min(1)` from the `skills` field in `volunteerProfileSchema.ts`
How to test:
1. Open a volunteer profile that has no skill set
2. Enter edit mode for the profile section
3. Attempt to save without selecting a skill
4. Confirm the form submits successfully

### DIFF
--- a/src/components/Dashboard/Profile/sections/VolunteerProfile/volunteerProfileSchema.ts
+++ b/src/components/Dashboard/Profile/sections/VolunteerProfile/volunteerProfileSchema.ts
@@ -54,7 +54,7 @@ export const createVolunteerProfileSchema = (t: (key: string) => string) => {
     activities: z
       .array(z.string())
       .min(1, t("dashboard.volunteerProfile.profileSection.validation.activitiesRequired")),
-    skills: z.array(z.string()).min(1, t("dashboard.volunteerProfile.profileSection.validation.skillsRequired")),
+    skills: z.array(z.string()),
   });
 };
 


nadavosa:
## Review

**Overview:** Removes the `.min(1)` validation constraint from the `skills` field in the volunteer profile schema, making it optional.

One-line fix, correct and targeted. `activities` still enforces at least one entry; `skills` now correctly allows empty arrays for volunteers with no skills. LGTM.

nadavosa (approved):

PR #509, opened 2026-05-08
fix: enforce email validation on form submit in opportunity creation

### LINKED ISSUE #421: Opportunity application form: email address validation is not enforced
### Summary

The email field on the opportunity application form no longer validates input. Any string is accepted and the form submits successfully with an invalid email address.

### Steps to Reproduce

1. Open any opportunity
2. Open the application/submit form
3. Enter an invalid email address (e.g. notanemail, abc@, test123)
4. Submit the form

### Expected Behavior

The form should reject invalid email addresses and show a validation error before allowing submission.

### Actual Behavior

The form accepts any value in the email field and submits without validation.

### Environment

_No response_

Cy-fox:
Closes #421

The email field in the Add Opportunity form only validated onChange and onChangeAsync. If a user submitted the form without ever interacting with the email field, the validation was skipped and the form could be submitted with an empty or invalid email.

Changes:
- Switched email field from individual validator props to the unified `validators` object on SimpleInputField
- Added an `onSubmit` validator that enforces required + format checks at submission time
- Preserved existing `onChange` and `onChangeAsync` validators for real-time feedback
How to test:
1. Go to Add Opportunity form
2. Fill all fields except email
3. Click submit without touching the email field
4. Verify that a validation error appears on the email field
5. Enter a valid RAC email and verify the form submits correctly

### DIFF
--- a/src/components/forms/AddOpportunity/AddOpportunity.tsx
+++ b/src/components/forms/AddOpportunity/AddOpportunity.tsx
@@ -146,17 +146,28 @@ export default function AddOpportunity() {
             name="email"
             FieldTag={formOpportunity.Field}
             label={t("form.addOpportunity.fields.contactGroup.email.label")}
-            onChangeValidator={({ value }) => {
-              if (!value) {
-                return t("form.error.required");
-              }
-              if (!validateEmail(value as string)) {
-                return t("form.error.email");
-              }
-              return undefined;
-            }}
-            onChangeAsyncValidator={({ value }) => {
-              return validateRACEmail(value as string, t("form.error.badEmail"));
+            validators={{
+              onChange: ({ value }) => {
+                if (!value) {
+                  return t("form.error.required");
+                }
+                if (!validateEmail(value as string)) {
+                  return t("form.error.email");
+                }
+                return undefined;
+              },
+              onChangeAsync: ({ value }) => {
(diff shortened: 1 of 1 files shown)


lourooo:
onSubmit doesn't include the async validateRACEmail check, a paste without yping scenario could bypass the badEmail validation. Likely intentional to avoid a network call on submit, but worth confirming. LGTM.

nadavosa (commented):
## Review

The refactor from individual `onChangeValidator`/`onChangeAsyncValidator` props to the `validators` object is correct, and adding `onSubmit` closes the gap where a user could submit without touching the field.

One edge case as Rodrigo noted: `onSubmit` doesn't run `validateRACEmail` (the async check). A user who pastes a non-RAC email and immediately hits submit without triggering `onChangeAsync` would bypass the bad-email check. Whether this is acceptable depends on how likely that flow is — if the field is pre-focused on page load the async fires on every keystroke anyway.

If you want to close the gap completely, you could call `validateRACEmail` synchronously inside `onSubmit` (it returns a string or undefined, the async wrapper just wraps a fetch). But if the intent is to keep submit-time validation lightweight and rely on onChange for the RAC check, add a comment explaining that so the next person doesn't wonder.

Otherwise the fix is correct and well-scoped. LGTM.

nadavosa (approved):
LGTM — email validation enforced on submit, async wrapper in place.

PR #510, opened 2026-05-08
fix: display operator name in agent organisation details

### LINKED ISSUE #490: bug: agent profile does not display operator (Träger) information
The agent profile page does not show the operator (Träger) that the agent belongs to. This information is important for coordinators to understand which organisation the agent represents.

**Expected behavior:** The agent profile should display the operator/Träger name (and ideally a link to the operator profile if applicable).

**Current behavior:** No operator/Träger information is visible anywhere on the agent profile page.

**Reporter:** Senya (06.05.2026)

Cy-fox:
Closes #490

The API returns the operator name at two levels: `agent.operator` (root) and `agent.agentDetails.operator` (nested). When the nested field is empty the OrganisationDetails form showed a blank value because it only read from `agentDetails`. Now the form initializer falls back to the root-level operator.

Changes:
- Added fallback chain `details?.operator || agent.operator || ""` in OrganisationDetails form initial values
How to test:
1. Open an agent profile where the operator name is set at root level but not in agentDetails
2. Verify the operator field shows the correct value in the Organisation Details section
3. Verify editing and saving still works correctly

### DIFF
--- a/src/components/Dashboard/Profile/sections/OrganisationDetails/OrganisationDetails.tsx
+++ b/src/components/Dashboard/Profile/sections/OrganisationDetails/OrganisationDetails.tsx
@@ -35,6 +35,7 @@ export const OrganisationDetails = forwardRef<EditableSectionRef, Props>(functio
 
   const initialFormValues = {
     ...details,
+    operator: details?.operator || agent.operator || "",
     clientLanguages: apiLanguagesToFormValues(details?.clientLanguages),
   };
 


lourooo:
|| treats "" as falsy, if details.operator is an empty string, it falls through to agent.operator. Since the field is min(1) in the schema this is harmless in practice, but ?? would be more precise. LGTM.

nadavosa (commented):
## Review

The fallback chain is correct and covers the described case. One precision nit echoing Rodrigo's comment: `details?.operator || agent.operator` treats an empty string as falsy, so if `details.operator` is `""` it falls through to `agent.operator`. Since the schema enforces `min(1)` this won't occur in practice, but `??` is semantically stricter (null/undefined only):

```ts
operator: details?.operator ?? agent.operator ?? "",
```

Otherwise this is a clean, minimal fix. LGTM.

nadavosa (approved):
LGTM — operator name display fixed.

PR #511, opened 2026-05-08
fix: invalidate complementary cache on volunteer-opportunity matching

### LINKED ISSUE #300: Matching doesnt work without refreshing
### Summary

after matching the volunteer from profile the matching doesnt go to the opportunity and also after refreshing doesnt saty at the volunteers profile - it needs refreshing to see it

### Steps to Reproduce

after opp in pending - match button

### Expected Behavior

volunteer in opp as matched - without need to refrresh 
opp in voluteer as matched 

### Actual Behavior

moves the opp at the volunteer profile to matched at first but not in the opp profile and then it does also stayed matched at the volunteer profile. the volunteer is MD. MIAJEE. he stayes pending - it only works after refreshing pages 

Image

Image

Image

### Environment

_No response_

Cy-fox:
Closes #300

When matching a volunteer from the volunteer profile page, only the volunteer-opportunities cache was invalidated. The opportunity-volunteers cache stayed stale, so navigating to the opportunity profile still showed the volunteer as pending until a manual page refresh.

Changes:
- After a successful status update or delete in the matching hooks, also invalidate the complementary query prefix (opportunity-volunteers when called from volunteer view, and vice versa)
- Added `getComplementaryPrefix()` helper to determine the other cache key
- Both `useUpdateOpportunityVolunteerStatus` and `useDeleteOpportunityVolunteer` now invalidate both sides
How to test:
1. Open a volunteer profile with a pending opportunity match
2. Click Match on an opportunity
3. Navigate to the opportunity profile without refreshing
4. Verify the volunteer now shows as matched (no refresh needed)
5. Repeat from the opportunity side: match a volunteer, then check the volunteer profile

### DIFF
--- a/src/hooks/useUpdateOpportunityVolunteerStatus.ts
+++ b/src/hooks/useUpdateOpportunityVolunteerStatus.ts
@@ -1,6 +1,7 @@
 import { apiPathOpportunityVolunteer } from "@/config/constants";
 import { useMutationQuery } from "@/hooks";
 import { OpportunityVolunteerStatusType } from "need4deed-sdk";
+import { useQueryClient } from "@tanstack/react-query";
 import axios from "axios";
 
 type StatusUpdatePayload = {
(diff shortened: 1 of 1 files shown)


lourooo:
getComplementaryPrefix silently falls back to "opportunity-volunteers" for any unrecognised key. Not a bug today, but worth an explicit guard if other query keys are ever passed. Otherwise LGTM.

nadavosa (commented):
## Review

The fix is correct and minimal — both `useUpdateOpportunityVolunteerStatus` and `useDeleteOpportunityVolunteer` now invalidate the complementary cache on success, so either side of the match reflects updates without a manual refresh.

One thing to harden: `getComplementaryPrefix` silently falls back to `"opportunity-volunteers"` for any key that isn't exactly `"opportunity-volunteers"`. This means an unexpected caller would silently invalidate the wrong cache. Since both hooks are always called with one of these two known keys this is fine in practice, but an explicit guard would make the intent clear:

```ts
function getComplementaryPrefix(queryKey: string[]): string {
  if (queryKey[0] === "volunteer-opportunities") return "opportunity-volunteers";
  if (queryKey[0] === "opportunity-volunteers") return "volunteer-opportunities";
  throw new Error(`getComplementaryPrefix: unexpected key "${queryKey[0]}"`);
}
```

This is a nit — the PR is correct as-is. LGTM.

nadavosa (approved):
LGTM — complementary cache invalidated correctly on match status change.

Issue #478, opened 2026-05-05
feat: agents list view


ivannissimrch:
Part of #441
Prerequisite: refactor/extract-entity-table-list must be merged first                                                                                                     
                  
 ## What this adds

 A flat table view on the agents' search page, toggled via the existing
 List/Cards tabs — same pattern as the volunteers list view (#500)

Name, type, volunteer search, district, active volunteers, email, number of opportunities

nadavosa:
Sub-issue of https://github.com/need4deed-org/fe/issues/441 (list view for all search pages).

ivannissimrch:
Documenting available fields from ApiAgentGetList for this table.
 Available now: title, type, volunteerSearch, district with title, activeVolunteers.
Missing from  API: email and number Of Opportunities. I will show em-dash, same as the volunteers pattern, until these are added on the backend.

Issue #441, opened 2026-05-04
feat: list view for all search pages (volunteers, opportunities, agents)


nadavosa:
Add a list/table view toggle to the volunteers, opportunities, and agents search pages alongside the current card view.

Name-Type-Engagment status-Matching status-Language-Postcode (idealy distric generated from it)-email

One of the issues is that we many times want to copy many email together to email volunteers (after filtering) so there should be an option to do that in the list view

PR #529, opened 2026-05-12
refactor: extract generic EntityTableList component

### LINKED ISSUE #526: refactor: extract generic EntityTableList component before #477 and #478
Part of #441                                                                                                                                                          
                  
  ## Context                                                                                                                                                            
            
  PR #500 shipped the volunteers list view with a hardcoded table structure                                                                                             
  (wrapper, header, pagination) inside `VolunteerTableList.tsx`. Before building
  the same pattern for opportunities (#477) and agents (#478), extract the shared
  logic into a reusable generic component to avoid duplicating it three times.   

## Why before #477 and #478                                                                                                                                           
                                                                                                                                                                        
  Without this, `OpportunityTableList` and `AgentTableList` would each duplicate                                                                                        
  the same wrapper, header rendering, and pagination logic. This makes #477                                                                                             
  and #478 a clean one-day task each.                                                                                                                                   
                                                                                                                                                                        
  ## References                                                                                                                                                         
               
  Suggested in Navad's PR #500 review (Looking ahead section)                                                                                                           
  Prerequisite for #477, #478

ivannissimrch:
## Description
Extracts a generic `EntityTableList<T>` component to `Dashboard/common/EntityTableList/` so opportunities (#477) and agents (#478) list views can reuse it instead of duplicating the table/header/pagination boilerplate

Also: 
- Replaces `selectedTabIndex` prop on `VolunteerListController` with `viewMode: ViewMode` ,  controller no longer knows about tab indices
- Adds `VOLUNTEER_COL_WIDTHS` as single source of truth for column widths

**Layout note:** With filters open the filter panel overflows off screen (table takes full width). Tried `flex: 1` on the container  so filters sit side by side but some columns get clipped and also tried an overlay approach  neither is clean. Figma has no spec for this state. What's the preferred behavior?

## Related Issues
Closes #526 Part of #441

## Changes
- Add `EntityTableList<T>` to `Dashboard/common/EntityTableList/`
- Add `ViewMode` type to `Dashboard/common/types.ts`
- Add `volunteerTableColumns.ts` with `createVolunteerTableColumns(t)` 
- Migrate `VolunteerTableList` to use `EntityTableList`
- Replace `selectedTabIndex` with `viewMode: ViewMode` on `VolunteerListController`

## Screenshots / Demos

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes

### DIFF
--- a/src/components/Dashboard/Volunteers/Volunteers.tsx
+++ b/src/components/Dashboard/Volunteers/Volunteers.tsx
@@ -21,6 +21,7 @@ import {
   serializeFilters,
 } from "./helpers";
 import { VolunteerListController } from "./VolunteerListController";
+import { ViewMode } from "../common/types";
 
 export function Volunteers() {
   const { t } = useTranslation();
@@ -34,7 +35,7 @@ export function Volunteers() {
   const pathname = usePathname();
   const router = useRouter();
   const tabs = [t("dashboard.volunteers.tabs.tab1"), t("dashboard.volunteers.tabs.tab2")];
-
+  const viewMode: ViewMode = selectedTabIndex === 0 ? "list" : "cards";
   const opportunityId = searchParams.get("opportunity") ?? undefined;
   const opportunityFilter = useGetOpportunity(opportunityId);
 
@@ -110,7 +111,7 @@ export function Volunteers() {
             filter={cardsFilter}
             apiFilterOptions={apiFilterOptions}
             opportunityId={opportunityId}
-            selectedTabIndex={selectedTabIndex}
+            viewMode={viewMode}
           />
           <Filters
             isFiltersOpen={isFiltersOpen}
--- a/src/components/Dashboard/Volunteers/Volunteers.tsx
+++ b/src/components/Dashboard/Volunteers/Volunteers.tsx
@@ -35,7 +35,7 @@ export function Volunteers() {
   const pathname = usePathname();
   const router = useRouter();
   const tabs = [t("dashboard.volunteers.tabs.tab1"), t("dashboard.volunteers.tabs.tab2")];
-  const viewMode: ViewMode = selectedTabIndex === 0 ? "list" : "cards";
+  const viewMode = Object.values(ViewMode)[selectedTabIndex];
   const opportunityId = searchParams.get("opportunity") ?? undefined;
   const opportunityFilter = useGetOpportunity(opportunityId);
 
--- a/src/components/Dashboard/common/types.ts
+++ b/src/components/Dashboard/common/types.ts
@@ -1,0 +1,1 @@
+export type ViewMode = "cards" | "list" | "map";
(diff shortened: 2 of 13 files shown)


nadavosa:
Two small things worth addressing:

**CSS variable in a generic component**
`styles.ts` uses `var(--opportunities-container-gap)` inside `EntityTableList`, which lives in `common/`. If that variable isn't declared in non-opportunity contexts this will silently collapse to zero. Either introduce a generic variable (e.g. `--entity-table-gap`), add a fallback value (`var(--opportunities-container-gap, 16px)`), or accept a `gap` prop.

**`viewMode === "list"` repeated in `VolunteerListController`**
It appears twice — once for the limit, once for the conditional render. Extracting `const isListView = viewMode === "list"` (same as the old `isListView`) keeps the intent clear and avoids the duplication.

ivannissimrch:
Thank you for the feedback. I implemented the following changes.

- I  added ` --entity-table-gap` as a generic CSS variable in the base: root of globals.css  and replaced the dependency on --opportunities-container-gap in the shared component.
- I added `const isListView = viewMode === "list"` at the top of VolunteerListController to replace the two repeated checks.

src/components/Dashboard/common/types.ts
```diff
@@ -0,0 +1 @@
+export type ViewMode = "cards" | "list" | "map";
```
arturasmckwcz: here I'd suggest to use `enum`:
```
export enum ViewMode {
  LIST = "list",
  CARDS = "cards",
  MAP = "map",
};
```

src/components/Dashboard/Volunteers/Volunteers.tsx
```diff
@@ -34,7 +35,7 @@ export function Volunteers() {
   const pathname = usePathname();
   const router = useRouter();
   const tabs = [t("dashboard.volunteers.tabs.tab1"), t("dashboard.volunteers.tabs.tab2")];
-
+  const viewMode: ViewMode = selectedTabIndex === 0 ? "list" : "cards";
```
arturasmckwcz: this is totally okay for 2 tabs, but I'd define mapping object, in this case just an array: `[...Object.values(ViewMode)][selectedTabIndex]`
`ViewMode` has to be an `enum` in such a case of course.

arturasmckwcz: or you may list values in an order you need instead of using `Object.values()`

arturasmckwcz: and no need for spreading as `Object.values()` creates a new array anyway, so it'd look: `Object.values(ViewMode)[selectedTabIndex]`

arturasmckwcz (commented):
LGTM
pls check comments and address if you recon they are to the point

src/components/Dashboard/Volunteers/Volunteers.tsx
ivannissimrch: Done , I used Object.values(ViewMode)[selectedTabIndex].
Also updated VolunteerListController to use ViewMode.LIST instead of the string literal.

src/components/Dashboard/common/types.ts
ivannissimrch: Done I converted ViewMode to an enum

PR #538, opened 2026-05-18
fix: render appointmentDistrict next to postcode and drop from PATCH payload

### LINKED ISSUE #533: Render appointmentDistrict next to postcode (display only); drop from PATCH payload
## Context

`appointmentDistrict` on opportunities is **server-calculated** from `appointmentPostcode` and is therefore read-only on the client. The current frontend doesn't surface it at all, and the PATCH payload type misleadingly suggests it can be sent.

### Display mode
- Render `appointmentDistrict` immediately after `appointmentPostcode`, separated by a comma.
- Single row, value formatted as: `"<postcode>, <district title>"` (e.g. `"10115, Mitte"`).
- District title comes from `appointmentDistrict.title` localized to the current i18n language (fall back to `de`, then to nothing if unavailable).
- If district is missing/empty, show just the postcode.

### Edit mode
- District field is **not** rendered. Only `appointmentPostcode` is editable.
- On save, the backend recalculates `appointmentDistrict` from the new postcode; the client re-reads it from the GET response.

### PATCH payload
- Remove `appointmentDistrict?: string` from `OpportunityAccompanyingDetailsUpdateData` in `src/hooks/useUpdateOpportunityAccompanyingDetails.ts`.
- The field must never be sent in PATCH requests.

## Files likely touched

- `src/hooks/useUpdateOpportunityAccompanyingDetails.ts` — drop `appointmentDistrict?: string`
- `src/components/Dashboard/Profile/sections/AccompanyingDetails/AccompanyingDetails.tsx` — compute combined `postcode, district` string from `opportunity.accompanyingDetails.appointmentDistrict.title[lang]`
- `src/components/Dashboard/Profile/sections/AccompanyingDetails/AccompanyingDetailsDisplay.tsx` — accept the combined value via prop and render it in the postcode row
- `AccompanyingDetailsEdit.tsx` — no change (no district field)

## Related

Originally spun out from review of #532; rescoped from pure type cleanup to include display rendering.

Cy-fox:
Closes #533

appointmentDistrict is server-calculated from appointmentPostcode and is read-only on the client. The frontend was incorrectly treating it as an editable string, including it in the form schema and PATCH payload, and failing to display it. This fix reads the localized district title directly from the API response and renders it inline with the postcode as a single row, removes the district field from the edit form, and drops it from the PATCH payload.

Changes:
- useUpdateOpportunityAccompanyingDetails: remove appointmentDistrict from payload type
- accompanyingDetailsSchema: remove appointmentDistrict field
- helpers: remove appointmentDistrict from getInitialFormValues
- AccompanyingDetailsEdit: remove district Controller and district-related props
- AccompanyingDetails: derive postcodeDisplay from opportunity.accompanyingDetails using localized Option title with de fallback; drop appointmentDistrict from onSubmit
- AccompanyingDetailsDisplay: merge postcode and district into one row; remove separate district row

### DIFF
--- a/src/components/Dashboard/Profile/sections/AccompanyingDetails/AccompanyingDetails.tsx
+++ b/src/components/Dashboard/Profile/sections/AccompanyingDetails/AccompanyingDetails.tsx
@@ -1,9 +1,9 @@
 "use client";
-import { useApiDistricts, useApiLanguages } from "@/components/Dashboard/Profile/sections/VolunteerProfile/hooks";
+import { useApiLanguages } from "@/components/Dashboard/Profile/sections/VolunteerProfile/hooks";
 import { useUpdateOpportunityAccompanyingDetails } from "@/hooks/useUpdateOpportunityAccompanyingDetails";
 import { zodResolver } from "@hookform/resolvers/zod";
 import { de, enUS } from "date-fns/locale";
-import { ApiOpportunityGet, LangPurpose } from "need4deed-sdk";
+import { ApiOpportunityAccompanyingDetails, ApiOpportunityGet, Lang, LangPurpose, Option } from "need4deed-sdk";
 import { forwardRef, useEffect, useImperativeHandle, useMemo, useState } from "react";
 import { FormProvider, useForm } from "react-hook-form";
 import { useTranslation } from "react-i18next";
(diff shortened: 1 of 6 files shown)


nadavosa (approved):
Good approach — district is server-calculated from postcode so making it display-only is correct.

One minor note: `rawDetails` is cast as `ApiOpportunityAccompanyingDetails & { appointmentPostcode?: string; appointmentDistrict?: Option }` to access fields not yet in the SDK type. This is the same workaround pattern used elsewhere in the codebase and is fine short-term. Should be cleaned up when the SDK type catches up.

The fallback chain `title?.[lang] ?? title?.de ?? ""` is correct. The `postcodeDisplay` logic (shows `postcode, district` or just `postcode` if no district yet) handles the edge cases cleanly.

PR #541, opened 2026-05-19
fix: add loading placeholder to dashboard list pages

### LINKED ISSUE #424: UX Feature: Add loading placeholder for Dashboard lists
### Description

When we're loading data on the` /dashboard` routes: Agents, Opportunity and  Volunteer, we get a layout shift from when the data is null to when it's populated from the backend.

This is most notable with Volunteers on `/dashboard/volunteer` , and it can give the impression that there are no results.

The height also causes an issue and can conflict with the layout of the footer.

Image

### Proposed solution

Use tanStack Query's `isLoading` state to conditionally display a placeholder. 

Could be as simple as 
```typescript
<Heading4> {t("dashboard.loading")} <Heading4>
``` 
but a lot better to create `Skeleton Cards` which greatly softens the layout shift.

Also propose using a responsive height for the `DashboardLayout`. This prevents the overlap of other components. In the `global.css` there is already the css var: `--dashboard-home-container-min-height: 300px;` which is used as a `min-height` on the `DashboardLayout` for the `Home` component

Cy-fox:
Closes #424

Dashboard Volunteer, Opportunity, and Agent lists had no loading state, causing a layout collapse and giving the impression of empty results while data was being fetched from the API. Each list controller now returns a centred loading message with a min-height container while isLoading is true, preventing the layout shift.

Changes:
- Add shared DashboardListLoading component (common/) with min-height and centred layout
- VolunteerListController: destructure isLoading from useGetQuery, return placeholder while loading
- OpportunityListController: same
- AgentListController: same

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,93 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project overview
+
+Need4Deed frontend — a Next.js 15 PWA for a Berlin NGO connecting volunteers with refugee accommodation centers. Deployed on AWS CloudFront. Open-source with volunteer contributors of varying experience levels.
+
+**Domain vocabulary (required reading):**
+
+- **Deed** — task posted by a center (e.g. "German tutoring", "translation")
+- **Volunteer** — registers via public form, has skills, languages, location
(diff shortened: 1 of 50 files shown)


nadavosa (approved):
Clean and correct. One thing to verify before merging: confirm that the translation key `dashboard.home.content.loading` exists in both locale files — it is not in this diff. If it is already there, good to go.

PR #555, opened 2026-05-22
fix: remove +49 phone number format restriction

nadavosa:
## Summary

- Removes `PHONE_NUMBER_REGEX` (`/^\+[0-9\s]+$/`) validation from phone fields in all three contact detail forms: volunteer, opportunity, and agent
- Phone fields now only require non-empty input — any format is accepted
- Affects: volunteer profile contact details, opportunity contact details, agent contact details (mobile + landline)

## Test plan

- [ ] Volunteer contact details: enter a number without `+` prefix (e.g. `01234567`) → saves without validation error
- [ ] Opportunity contact details: same
- [ ] Agent contact details (mobile + landline): same
- [ ] Phone field still rejects empty string

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "date-fns": "^4.1.0",
     "email-validator": "^2.0.4",
     "i18next": "^25.3.2",
-    "need4deed-sdk": "^0.0.85",
+    "need4deed-sdk": "^0.0.95",
     "next": "15.3.8",
     "react": "^19.0.0",
     "react-day-picker": "^9.13.0",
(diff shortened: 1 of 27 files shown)


ivannissimrch (commented):
The validation change works correctly across all three forms. However, the error message strings still describe the OLD rules  "must start with + and contain only digits and spaces," which no longer match the new regex; + is no longer required; parens, dashes, and slashes are now allowed.

[Screencast from 2026-05-25 12-14-39.webm](https://github.com/user-attachments/assets/dee8a5ec-9c77-4bd7-9f37-fce3a47b65fd)

PR #565, opened 2026-05-29
feat: show logged-by username in communication tracker

### LINKED ISSUE #455: feat: communication tracker — show username next to each contact entry
In the volunteer profile communication tracker, there is no indication of who made the contact. Each entry should display the username of the coordinator who logged the contact.

Cy-fox:
Closes #455

The communication tracker table had no indication of who logged each contact entry. A new "By" column is added that shows the coordinator's full name for entries made by the currently logged-in user, falling back to a numeric user ID for entries by other users.

The fallback is intentional: the communication GET endpoint currently returns only userId (not the full user object), so resolving names for other coordinators requires a backend update to load the user relation and embed fullName in the response.

Changes:
- CommunicationTracker: import useCurrentUser, add getLoggedByLabel helper, add By column header and cell to the table
- locales EN/DE: add dashboard.communicationSection.loggedBy key ("By" / "Von")

How to test:
1. Open any volunteer or agent profile with communication tracker entries
2. The table should now have a "By" column as the 4th column
3. Entries logged by the current user should show their full name
4. Entries logged by other users show their numeric user ID (e.g. #3) until the BE embeds user info

### DIFF
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -1262,6 +1262,7 @@
       "whatPlatform": "Welche Plattform haben Sie verwendet?",
       "howDidYouSend": "Wie haben Sie Ihre Nachricht gesendet?",
       "contactMethodLabel": "Kontaktmethode",
+      "loggedBy": "Von",
       "contactMethodRequired": "Kontaktmethode*",
       "communicationTypeRequired": "Kommunikationstyp*",
       "communicationTypes": {
--- a/public/locales/en/translations.json
+++ b/public/locales/en/translations.json
@@ -1260,6 +1260,7 @@
       "whatPlatform": "What platform did you use?",
       "howDidYouSend": "How did you send your message?",
       "contactMethodLabel": "Contact method",
+      "loggedBy": "By",
       "contactMethodRequired": "Contact method*",
       "communicationTypeRequired": "Communication type*",
       "communicationTypes": {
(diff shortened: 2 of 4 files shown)


nadavosa:
Two issues:

**1. `getLoggedByLabel` only resolves the name for the current user**
For any other coordinator's entry, it falls back to `#${userId}` (a raw ID number, not a name). The feature title says "show logged-by username" but users will see `#42` for entries logged by their colleagues. This requires the BE to return the user's name (e.g. `fullName`) in the communication response — worth flagging as a follow-up issue.

**2. `getLoggedByLabel` should be in `helpers.ts`**
It's a pure function with no component dependencies — it belongs in the existing helpers file alongside `getDisplayLabel` and `getContactMethodLabel`, not inline in the component.

nadavich:
## Code Review

Clean, well-scoped feature. The acknowledged fallback to `#userId` for other users is fine. A couple of things to fix:

**1. `entry.userId` may be null on legacy entries**

If `userId` is `null`/`undefined` on older communication records, the cell renders `#null` or `#undefined`. Guard it:
```tsx
{entry.userId ? getLoggedByLabel(entry.userId) : "—"}
```

**2. `currentUser.fullName` could be empty**

```ts
if (currentUser && userId === currentUser.id) return currentUser.fullName;
```
If `fullName` is empty/null the cell renders blank with no fallback. Suggest:
```ts
if (currentUser && userId === currentUser.id) return currentUser.fullName || `#${userId}`;
```

**3. `useCurrentUser(true)` — clarify the argument**

The `true` arg isn't obvious. If it means "required" and throws when unauthenticated, it could cause an error during auth loading. Worth confirming or adding a short comment.

**Minor**
- `getLoggedByLabel` defined inside render — a `useCallback` would be cleaner as the component grows
- No tests

**Verdict:** approve with fix for #1 — `#null` rendering on legacy entries is a visible bug. #2 and #3 are lower priority.

need4deed (approved):

PR #556, opened 2026-05-24
Add agents list view (#478)

ivannissimrch:
## Description
 Adds the list-view (table) variant for the agents dashboard.

## Related Issues
Closes #478 

## Changes

- New `AgentTableList` and `AgentTableRow` wired into `AgentListController` via `viewMode`
- 7 agent table columns (title, type, volunteerSearch, district, activeVolunteers, numberOfOpportunities, email) with EN and DE translations.
- District lookup via `districtsList` (table and  card paths)  fixes empty district render
- Email rendering with temporary cast workaround (see SDK note below)
- New shared `COLUMN_WIDTH` constants (`XS`/`SM`/`MD`/`LG`), both agent and volunteer tables migrated
- New shared `WrapAnywhereCell` for long unbreakable text (used by agent email column)
- Card constant globals consolidation: `CARD_COLUMNS`/`CARD_ROWS` moved to `src/config/constants.ts` (touches Agent, Volunteer and Opportunity controllers)
- `createAgentTypeMap` moved out of `icon.tsx` into `Agents/constants.ts` 
- DE translation fix: `dashboard.agents.*` now uses **Bezirk**

 ## SDK 0.0.91 note (relevant to email column)
 Tried to bump to `need4deed-sdk@0.0.91` to pick up the typed `email` field on `ApiAgentGetList`. 
0.0.91 introduced `Voidable<T>` (= `T | null | undefined`) on many fields where they were previously `T | undefined`, which broke 10 typecheck errors across 6 unrelated files. I stayed on `0.0.90` for this PR and used a cast workaround for `email` (with a TODO comment explaining when to remove).

- `numberOfOpportunities` column currently renders `0,` but currently the BE is not sending the number of opportunities; we have `activeVolunteers and numActiveVolunteers`  (which appear to be the same value, possibly duplicate metrics).

## Screenshots / Demos
Screenshot from 2026-05-24 12-14-24

## Checklist
- [x] WITHIN THE SCOPE OF AN ISSUE; No unnecessary files included
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

PR #578, opened 2026-05-31
Add opportunities list view (#477)

### LINKED ISSUE #477: feat: opportunities list view

ivannissimrch:
## Description
Adds the table (list view) for the opportunities dashboard, toggled via the existing List/Cards tabs

 **Known limitations / decisions:**
  - **Horizontal scroll lives in the shared `EntityTableList`** (10 columns are wider than the dashboard).The volunteer table now scrolls too. I didn't adjust its column widths here to avoid conflicts with the active agent table PR (#556).
  - **Overlaps #556:** both branches add `COLUMN_WIDTH` / `CARD_*` constants and touch
    `EntityTableList`, so whichever merges second needs a small rebase to dedupe. Follow-up after
    #556 lands: dedupe shared constants, tidy volunteer/agent widths under the scroll, add a sticky title column, and swap the local `TitleCell` for the shared `WrapAnywhereCell`.

## Related Issues
Closes #477 

## Changes
  - New `OpportunityTableList`, `OpportunityTableRow`, wired into `OpportunityListController` via `viewMode`
  - 10 columns (title, type, current status, matching status, languages, activities, district,
    schedule, number of volunteers, agent) with translations
  - District, activities rendered via `districtsList`, `activitiesList` id lookups (same as the card)
  - Schedule derived from `accompanyingDetails` (accompanying) or `availability` (regular)
  - Long titles wrap via `overflow-wrap: anywhere`
  - Em-dash placeholder for missing values
  - Widened `formatAccompanyingDate` to accept the voidable `appointmentDate`.
  - Added horizontal scroll to the shared `EntityTableList`; introduced shared `COLUMN_WIDTH` + `CARD_*` constants

## Screenshots / Demos
[Screencast from 2026-05-31 17-16-02.webm](https://github.com/user-attachments/assets/711a6896-9278-403d-a5c0-0e0715761030)

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes


nadavosa:
Good work. Two blockers before merging:

**1. Must merge after https://github.com/need4deed-org/fe/pull/556**
Both PRs add `CARD_COLUMNS`, `CARD_ROWS`, `CARD_LIMIT` to `constants.ts`. This PR also relies on `TABLE_LIMIT` being present in constants, which #556 moves to the bottom. Merge #556 first, then rebase this branch to remove the duplicate constant additions.

**2. `agentTitle` and `numberOfVolunteers` on `ApiVolunteerOpportunityGetList`**
`OpportunityTableRow.tsx` destructures `agentTitle` and `numberOfVolunteers` from the opportunity object, but these fields are not in the current SDK. This PR doesn't bump the SDK version — it needs the SDK bump from #556 (0.0.85 → 0.0.90) to land first, and even then check whether these fields exist in 0.0.90. If they don't, the TypeScript build will fail.

nadavosa:
Conflicts resolved and pushed — TABLE_LIMIT from #556, ScrollableTable/WrapAnywhereCell merged (both kept), ViewMode import restored. Ready to review again.

src/components/Dashboard/Opportunities/OpportunityListController.tsx
```diff
@@ -84,6 +90,20 @@ export function OpportunityListController({
 
   if (isLoading) return <DashboardListLoading />;
 
+  if (isListView) {
+    return (
+      <OpportunityTableList
```
DarrellRoberts, line 95: I noticed that when the Filter is open, there is a horizontal overflow creating a scrollbar on the window. We could pass the `isFilterOpen` prop down here and recalculate the columns if the Filter is open?

src/components/Dashboard/common/EntityTableList/styles.ts
```diff
@@ -8,6 +9,15 @@ export const Wrapper = styled.div`
   width: 100%;
```
DarrellRoberts: alternatively, to the Filter issue I mentioned on `‎/OpportunityListController.tsx`, we could add a `min-width` here instead of `width` for a css-only solution and say something like:
```typescript
export const Wrapper = styled.div`
  display: flex;
  flex-direction: column;
  gap: var(--entity-table-gap);
  min-width: 85%;
`;
```

DarrellRoberts (commented):
really cool feature! I just had one point on the `Filter` causing an overflow on the horizontal scrollbar of the browser

Image

src/components/Dashboard/common/EntityTableList/styles.ts
ivannissimrch: Thank you, I went with the CSS min-width option, the window overflow's gone, and the filter's visible now.

PR #592, opened 2026-06-03
feat(#479): copy filtered volunteer emails to clipboard

### LINKED ISSUE #479: feat: bulk email copy from filtered list
Allow coordinators to copy all volunteer emails from the current filtered list view to clipboard for bulk emailing. Fully blocked until email is added to the volunteerListSerializer and the SDK is updated.

ivannissimrch:
## Description
 Adds a "Copy emails" button to the volunteer list (in the email column header). Clicking it copies the email addresses of **all volunteers matching the current filter** to the clipboard.

**How it works:** The list endpoint caps at 120/page, so a `useCopyVolunteerEmails` hook fetches the filtered list imperatively (`queryClient.fetchQuery`), pages through if there are more than 120, extracts the emails, and writes them to the clipboard. `CopyButton` and `copyEmails` are generic/shared, ready to reuse for the agents table later.

**Known limitation:**  works well on short filtered lists. On the long unfiltered list, the BE 503s/ times out; it can't serve the heavy paginated full-object queries fast enough. This is a backend limit. Later on will use a dedicated emails-only, unpaginated endpoint. The hook would then swap the page-loop for a single call.

## Related Issues
Closes #479 

## Changes
  - `useCopyVolunteerEmails` hook, imperative paged fetch, extract, copy emails, loading state, toasts
  - `CopyButton` ,  `copyEmails` clipboard helper in `common/`
  - `Column.headerAction` support in `EntityTableList`, copy button in the email column header
  - Exported `fetchData` / `getReducedFilter` from `useGetQuery` for reuse
  - `MAX_PAGE_LIMIT` constant (the BE page cap)
  - en/de translations for the tooltip and toasts
  - Minor: volunteer column-width rebalance; dropped the `statusMatch`/`email` cast (SDK now exposes them)

## Screenshots / Demos
[Screencast from 2026-06-03 06-47-42.webm](https://github.com/user-attachments/assets/f2e0177f-1dec-43e2-93b6-eb56a6f0bc1a)

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes


nadavich:
## Code Review

### What's Good
- Hook design is solid — imperative `queryClient.fetchQuery` is the right tool here
- Error handling: try/catch/finally correctly resets `isCopying`; toasts for both success and error
- `headerAction?: ReactNode` is a clean, minimal extension to `EntityTableList`
- SDK cast removed — nice cleanup
- Both en/de translations covered including the interpolated count

---

### Issues

**1. Unnecessary arrow wrapper**

In `VolunteerTableList.tsx`:
```tsx
// current
onClick={() => onCopyEmails()}
// should be
onClick={onCopyEmails}
```

**2. `fetchQuery` may return stale cached data**

`fetchEmailPage` doesn't set `staleTime: 0`, so if the same key is cached from a prior click it returns stale data without a network call:
```ts
queryClient.fetchQuery({
  queryKey: ["volunteer-emails", serializedFilter.toString(), page],
  staleTime: 0,  // add this
  queryFn: () => fetchData(...)
})
```

**3. `ActionButtonWithTooltip` import path is a layering violation**

`CopyButton` (in `Dashboard/common/`) imports from `Dashboard/Profile/sections/VolunteerProfileDocument/`. A shared common component shouldn't reach into a feature subtree. `ActionButtonWithTooltip` should be moved to `Dashboard/common/` or `CopyButton` should inline a simpler implementation.

**4. `email` not narrowed after `filter(Boolean)`**

```ts
// current — still infers (string | undefined)[]
.filter(Boolean)

// fix
.filter((e): e is string => Boolean(e))
```

**5. Generic error message misleads on non-timeout failures**

"Try narrowing the filter" is correct for the 503 case but confusing for network errors or permission failures. Consider a neutral fallback or checking the error type.

**6. No tests**

The async pagination path (`fetchAllFilteredEmails`: first page → derive total pages → parallel fetch rest) is the most risk-prone logic here. Worth at minimum: single-page case, multi-page case, and error case.

---

### Minor
- `copyEmails` lives in `Dashboard/common/` but is a pure utility — `src/utils/` would be more natural
- `MAX_PAGE_LIMIT = 120` could use a short comment (`// BE hard cap per page`)

---

**Verdict:** Approve with minor fixes — the two worth addressing before merge are `staleTime: 0` (#2) and the `ActionButtonWithTooltip` import path (#3). No security concerns, no regressions to existing columns.

PR #595, opened 2026-06-04
perf(volunteers): send listType so backend loads only needed relations

arturasmckwcz:
## What

`VolunteerListController` now sends `listType` on `GET /volunteer`:
- table view → `listType=table`
- card view → `listType=card`

The backend uses this to load (and return) only the relations each view renders — the table needs just languages + locations, so it no longer over-fetches activities/skills/availability.

## Why

Part of the `GET /volunteer` performance fix (was ~18-20s). Backend: need4deed-org/be#634.

## ⚠️ Deploy ordering

Requires **be#634 to be deployed first** — that backend's query schema uses `additionalProperties: false`, so an old backend would reject the `listType` param with a 400. Once be#634 is live, `listType` defaults to `card` server-side if absent, so there's no hard coupling beyond ordering.


need4deed (approved):
Approved: send listType so backend loads only the relations each view renders. Pairs with be#634 (now merged).

PR #591, opened 2026-06-02
fix: add waysToContact field to opportunity contact details

Cy-fox:
Closes #579

ApiOpportunityContact has 5 fields (id, name, phone, email, waysToContact) but the contact details section only displayed and edited 3. The section also used a piped-comment workaround inherited from early development to read and save contact data, bypassing the proper contact API that has since been available.

This PR adds the missing waysToContact field and replaces the comment-based approach with opportunity.contact and useUpdateOpportunityContact (which already existed in the codebase).

Changes:
- opportunityContactDetailsSchema: add waysToContact as optional PreferredCommunicationType array
- OpportunityContactDetails: remove comment parsing logic; read initial values from opportunity.contact; save via useUpdateOpportunityContact; wire useEnumTranslation for waysToContact label mapping
- OpportunityContactDetailsDisplay: add waysToContact as a checkbox-list display row
- OpportunityContactDetailsEdit: add waysToContact Controller with checkbox-list and full label mapping

:)

PR #601, opened 2026-06-05
🐛 fix(opportunity): source contact details from .contact object

arturasmckwcz:
## What

`OpportunityContactDetails` now sources contact name/phone/email from the fetched `ApiOpportunityGet.contact` object instead of parsing `<|>`-delimited strings out of the `comments` array.

## Why

The opportunity GET response already exposes a proper `contact` object (`{ id, name, phone, email, waysToContact }`). The section was ignoring it and instead reverse-engineering the values from a legacy system comment, then saving via comment create/update.

## Changes

- Read initial form values directly from `opportunity.contact`.
- Persist edits through the existing (previously unused) `useUpdateOpportunityContact` hook, which `PATCH`es the opportunity. `contact.id` and `contact.waysToContact` are preserved; only the three edited fields are overwritten.
- Removed all `<|>` comment-parsing logic and the `useCreateComment` / `useUpdateComment` usage from this section. Verified no other reader/writer of the `<|>` delimiter remains in `src/`.

## Checks

- `yarn typecheck` ✅
- `yarn lint` ✅

### DIFF
--- a/src/components/Dashboard/Profile/sections/ContactDetails/opportunity/OpportunityContactDetails.tsx
+++ b/src/components/Dashboard/Profile/sections/ContactDetails/opportunity/OpportunityContactDetails.tsx
@@ -1,6 +1,5 @@
 "use client";
-import { useCreateComment } from "@/hooks/useCreateComment";
-import { useUpdateComment } from "@/hooks/useUpdateComment";
+import { useUpdateOpportunityContact } from "@/hooks/useUpdateOpportunityContact";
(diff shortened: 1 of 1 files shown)


arturasmckwcz:
Superseded by #599 (read opportunity contact from `opportunity.contact`) and #591 (waysToContact editing), both already merged to develop. This PR fixes the same root cause but drops the legacy `<|>`-comment fallback and the waysToContact UI those PRs added, so it would regress. Closing as redundant.

Issue #597, opened 2026-06-05
fix: cap multi-value fields at 2 with +N overflow on opportunity card and table views


nadavosa:
## Problem

Opportunity cards and table rows expand vertically when an opportunity has many activities or languages, causing inconsistent row/card heights and hard-to-scan lists.

### Card view (`OpportunityCard.tsx`)
- `activityTitles` → pass `max={2}` to `<Tags>` (already supports the prop from #586 fix)

### Table view (`OpportunityTableRow.tsx`)
- `activityTitles.join(", ")` → `truncateList(activityTitles, 2)` with `+N` overflow
- Add `$noWrap` to all `TableCell` components to lock row height

## Expected behaviour

Same as the volunteer list/card fix in #583 and #586 — at most 2 items shown with `+N` indicator for the rest.

thnndnly:
Hey, I've implemented this as described and it works well — one question about the title column before I open the PR.

  Done as specified:
  - Card view: max={2} passed to <Tags>, so activities cap at 2 with the +N badge from the #586 fix.
  - Table view: truncateList(activityTitles, 2) with +N overflow, and $noWrap on the TableCell components to lock the row height. I also wrapped the cell content in TruncatedText (the pattern from #610), so
  clipped values end with an ellipsis instead of being cut mid-character.´

<img width="1920" height="1032" alt="Image" src="https://github.com/user-attachments/assets/471f2da0-856a-4436-812c-9f4a4a14bbb3" />

The question: Taking "add $noWrap to all TableCell components" literally also truncates the title, so e.g. "Help with German lessons" renders as "Help with Ge…" — the main identifier of the row
  becomes unreadable. Two things make me think the title should be excluded:

  - In the volunteer fix (#583 / PR #584) the name column was the one column that did not get $noWrap — the length problem there was solved on the data side (first name only). An opportunity title has no short
  version like that.
  - The title cell currently has a deliberate overflow-wrap: anywhere from an earlier fix ("wrap long opportunity titles to prevent column overflow"), so long titles wrap instead of overflowing.

  My suggestion: keep $noWrap + truncation on all other columns (which fixes the actual row-height problem, since title wrapping is rare), but let the title keep its current wrapping behaviour. Happy to
  truncate the title too if you prefer strict fixed-height rows — just let me know.

need4deed:
@thnndnly you are totally right, no title truncation should happen

need4deed:
[...]

PR #618, opened 2026-06-09
Opportunities table: drop filterable columns

### LINKED ISSUE #612: Opportunities table: drop filterable columns
## Summary
  Apply the opportunities-table changes requested in the 2026-06-06 standup.

 ## Changes
  - Drop **type**, **current status**, **activities** (all filterable)
  - Keep **matching status** (not filterable)
  - Position **Schedule** on the  2nd column
  - Single-line
  -  Shrink the Volunteers Needed column.

ivannissimrch:
## Description
Opportunities table: drop filterable columns

## Related Issues
Closes #612 

## Changes
  - Drop **type**, **current status**, **activities** (all filterable)
  - Keep **matching status** (not filterable)
  - Move **Schedule** to 2nd column
  - Single-line rows (truncate with ellipsis)
  - Shrink the **Volunteers Needed** column (numbers only)
  - Schedule uses a local formatSchedule helper (translated German day names
  - Move `formatAccompanyingDate` into `Opportunities/helpers`

 ## Notes/follow-ups.
  - **Schedule format:** The column feels a bit narrow to show enough info. @nadavosa, how do you want this? Show more across multiple lines, or format it differently?
 **Formatter:** The opportunities table uses a new local formatSchedule in Opportunities/helpers rather than reusing either existing formatter, the volunteer-profile one (translated, but coupled to getScheduleState) or the opportunities-card one (decoupled, but missing translations). The local one gives us both: translated and standalone. Folding all three into one shared helper may be a separate cleanup.

## Screenshots / Demos
Screenshot from 2026-06-08 20-14-06

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes


ivannissimrch:
**Updated:** switched the schedule to a local formatSchedule (decoupled from getScheduleState) instead of the formatAvailability, function and refreshed the description. Synced with develop.

thnndnly:
Cross-reference from #597: @need4deed wrote there that titles should not be truncated ("no title truncation should happen" — https://github.com/need4deed-org/fe/issues/597#issuecomment-4683231059). In this PR the title cell gets `$noWrap` + `TruncatedText` and the `overflow-wrap` TitleCell is removed, so long titles render as e.g. "Help with Ge…". If that applies here as well, the title cell can simply be excluded from the `$noWrap` change — the other columns are unaffected.

ivannissimrch:
Thanks @thnndnly, I just pushed a fix, added a WrappedText primitive (overflow-wrap: anywhere) next to TruncatedText and used it for the title so it wraps instead of truncating. Long single-word titles break properly now, too.
Heads up on the schedule: I'm wrapping that one on purpose as well. Came out of today's standup: Nadav wants the full date and time visible, so it's intentional. The other columns still truncate like before.

Screenshot from 2026-06-12 19-12-07

thnndnly (approved):
Looks good, thanks. The last commit addresses my truncation note from #597 — title and schedule now wrap instead of getting cut off.

One follow-up (non-blocking): as you mentioned, the new `formatSchedule` is almost identical to the existing `formatAvailability` in the volunteer profile. Worth folding into one shared helper as a separate cleanup, not here.

PR #626, opened 2026-06-09
Agent table: email copy per-row, copy-all and shared useCopyEmails refactor

### LINKED ISSUE #611: feat: email copy buttons on agent table (per-row and copy-all)
## Summary
Add email copy to the agent dashboard table, matching what volunteers already have (#479 and per-row copy in #610). Decided in the 2026-06-06 standup: single email copy per row **and** copy-all, for both volunteers and agents ("so we have both options").

ivannissimrch:
## Description
Adds email copy to the agent (NGO) table, per-row single copy and copy-all, matching what volunteers already have. Building it surfaced that the copy-email hook, translations were volunteer-specific, but are now shared, so I generalized them.

## Related Issues
Closes #611 

## Changes
  **Feature agent email copy**
  - Per-row copy button in the email cell (reuses the shared `CopyEmail` component).
  - Copy-all button in the email column header (via `headerAction`).

  **Refactor, one shared hook**
  - Generalized `useCopyVolunteerEmails` to  `useCopyEmails(apiPath, cacheKey, serializedFilter)`.
 
  **Refactor shared translations**
  - Moved the `copyEmails` block `dashboard.volunteers`  to  `dashboard.common` (en and de) since both tables use it; repointed all references.

## Screenshots / Demos
[Screencast from 2026-06-09 16-39-35.webm](https://github.com/user-attachments/assets/e5664d31-53ef-467f-92fc-f4e57bc3279b)

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes


nadavosa (commented):
1. **Trailing slash** — hook does \`${apiPath}/\`; if \`apiPathAgent\` already ends with \`/\`, the fetch URL breaks. Verify in \`constants.ts\`.

2. **aria-label = tooltip** on copy-all button — both use the same translation key, no extra info for screen readers. Use a distinct key or string.

3. **\`district\` column narrowed** from \`LG\` → \`MD\` — side-effect of widening email. Check it doesn't clip long district names.

4. **\`EmailRecord\` local type** — check the SDK doesn't already have a base type with an \`email\` field before keeping a local one (per CLAUDE.md).

ivannissimrch:
**Issues addressed** 

**Trailing Slash:**
Verified in constants.ts - apiPathAgent is /api/agent, no trailing slash, so ${apiPath}/ resolved to /api/agent/ (single slash, no // break). I also refactored the hook to pass apiPath through unchanged so it doesn't force the slash at all. Each caller now passes the path that matches its own list fetch. 

**Aria Label:**
- Fixed the issue by creating dedicated `copyAriaAll` keys for each table, available in both English and German.

**Width:**
- Verified that double-line rows properly wrap long names without clipping at the medium (MD) breakpoint. The setting has been left as medium.

**Email Record:**
- I couldn't find an SDK type for this.

PR #629, opened 2026-06-10
fix: wrap volunteer form in Suspense to prevent hydration crash

### LINKED ISSUE #628: Volunteer form crashes for users arriving via opportunity link (useSearchParams hydration mismatch)
## Bug

Volunteers who access the registration form via an opportunity-specific link (e.g. `https://app.need4deed.org/de/forms/volunteer?id=123&title=Some+Opportunity`) see:

> Application error: a client-side exception has occurred while loading app.need4deed.org

Volunteers who go directly to `/forms/volunteer` with no query params are unaffected.

## Root cause

`BecomeVolunteer` calls `useSearchParams()` to read `?title` and conditionally renders a subtitle `<h6>`:

```tsx
{opportunity.title ? <h6>...</h6> : null}
```

During SSR, search params are not available → `opportunity.title = ""` → no `<h6>` rendered in the server HTML.

On the client, the URL has `?title=Some Opportunity` → React renders the `<h6>` → **hydration mismatch** → crash.

In Next.js 15 App Router, any component using `useSearchParams()` must be wrapped in a `<Suspense>` boundary so the component is skipped during SSR and rendered entirely client-side after hydration.

There is currently zero `<Suspense>` usage anywhere in the codebase.

## Fix

Wrap `<BecomeVolunteer />` in a `<Suspense>` boundary in `src/app/[lang]/forms/volunteer/page.tsx`.

## Impact

Affects all volunteers arriving via opportunity-specific registration links (links shared by centers to their volunteers). The form shows a full-page crash and they cannot register.

nadavosa:
## Summary
- Wraps `<BecomeVolunteer />` in `<Suspense>` in the page component
- One-line change: adds `import { Suspense }` and the wrapper

## Root cause
`BecomeVolunteer` calls `useSearchParams()` to read `?id` and `?title` from the URL and conditionally renders a subtitle when `title` is present. Without a `<Suspense>` boundary, Next.js 15 App Router renders the page on the server (where search params are unavailable → `title = ""`), then React hydrates on the client (where `?title=Some Opportunity` is present → subtitle rendered). This server/client mismatch causes:

> Application error: a client-side exception has occurred while loading app.need4deed.org

Only volunteers arriving via an **opportunity-specific link** (with `?title=` in the URL) were affected — those navigating within the app via client-side routing never hit SSR so no mismatch occurred.

## Test plan
- [ ] Visit `https://app.need4deed.org/de/forms/volunteer?id=123&title=TestOpportunity` directly (fresh load) → form loads without error, subtitle visible
- [ ] Visit `https://app.need4deed.org/de/forms/volunteer` directly → form loads without error, no subtitle
- [ ] Submit the form → works as before

Resolves https://github.com/need4deed-org/fe/issues/628

### DIFF
--- a/src/app/[lang]/forms/volunteer/page.tsx
+++ b/src/app/[lang]/forms/volunteer/page.tsx
@@ -1,5 +1,10 @@
+import { Suspense } from "react";
 import BecomeVolunteer from "@/components/forms/BecomeVolunteer/BecomeVolunteer";
 
 export default function VolunteerPage() {
-  return <BecomeVolunteer />;
+  return (
+    <Suspense>
+      <BecomeVolunteer />
+    </Suspense>
+  );
 }


DarrellRoberts (approved):
I tried to reproduce the bug in numerous ways but everytime never encountered the error (e.g. loading the url in different browsers, slow networks, incognito). 

All I can think of is, if you know roughly when these volunteers were having these issues, for whatever service you are using to deploy the `fe` , look at the terminal logs during the period on this service and it should indicate what happened.

Nonetheless, this looks good

PR #643, opened 2026-06-11
fix(#638): wrap opportunity card status tags within card boundary

### LINKED ISSUE #638: fix: opportunity card tags overflow beyond card boundary
## Description

Tags on opportunity cards i.e.- type can overflow beyond the card's visible boundary instead of being clipped or wrapping correctly. This breaks the card layout, especially when multiple tags are present.
 

## Expected behaviour

Tags stay within the card boundary. Overflow is either clipped, wrapped to the next line within the card, or truncated with a `+N` indicator (see related https://github.com/need4deed-org/fe/issues/597).

## Actual behaviour

Tags extend beyond the edge of the card, overlapping adjacent elements or spilling outside the card container.

Image

## Acceptance Criteria

- [ ] No tags extend beyond the card boundary in any viewport size
- [ ] Fix does not break the card layout for cards with few or no tags

thnndnly:
## Description

The status/type tag row on opportunity cards (`StatusTagsDiv`) is a flex row without `flex-wrap`, while the card has a fixed width (320px). With three tags (status + match status + type) the row extends past the card boundary — up to ~74px in English and on every tested combination in German, where even the shortest labels overflow.

Adding `flex-wrap: wrap` lets tags wrap to the next line inside the card. The card already uses `min-height`, so it grows naturally with the extra row. Cards with few or short tags keep a single row. Cards in the same grid row keep equal heights (the grid stretches them), so rows stay flush — only the vertical start of the content varies with the number of tag lines.

## Related Issues

Closes #638

## Scope

- The volunteer card has its own copy of the same styled components (`VolunteerCard.tsx`) with the same missing `flex-wrap` — left out here to keep the PR card-only; happy to follow up.

### DIFF
--- a/src/components/Dashboard/Opportunities/styles.ts
+++ b/src/components/Dashboard/Opportunities/styles.ts
@@ -43,6 +43,7 @@ export const Card = styled(BaseCard)`
 export const StatusTagsDiv = styled.div`
   display: flex;
   flex-direction: row;
+  flex-wrap: wrap;
   gap: var(--dashboard-volunteers-card-status-tags-div-gap);
   margin-top: var(--dashboard-volunteers-card-status-tags-div-margin-top);
 `;


DarrellRoberts (approved):
LGMT, thanks!

DarrellRoberts:
regarding the scope, are you able to do the same with VolunteerCards?

thnndnly:
@DarrellRoberts yes. `VolunteerCard` has its own copy of `StatusTagsDiv` (it isn't shared with the opportunity card) and it's missing the same `flex-wrap: wrap`, so it has the identical overflow. I'll open a follow-up PR adding it there.

Side note: the two `StatusTagsDiv` styles are duplicated across the two cards — I can fold them into one shared styled component in the same follow-up if you'd prefer, otherwise I'll keep it to the minimal one-line fix.

DarrellRoberts:
thanks @thnndnly ! I'd say just change the `flew-wrap` for now. I think the `VolunteerCard` is in need of refactoring anyway as it's a bit of a mess

PR #644, opened 2026-06-12
refactor: remove "as unknown as" type workarounds (#281)

### LINKED ISSUE #281: remove TS workarounds
remove `as unknown as`s
e.g.
[src/components/Dashboard/Profile/sections/OrganisationDetails/OrganisationDetails.tsx:38:3](https://github.com/need4deed-org/fe/pull/263/files/26019d481873b4c31b236e83bee3aef41285483d#diff-30ab1b01e36e8d128ab47296a513b81755267681c650566058f3e97ed7b44d98)

@HansKre

thnndnly:
Closes #281

Removes all 11 `as unknown as` double assertions from the codebase. A double cast
disables type checking entirely at the use site, so errors only surface at runtime.
Each occurrence fell into one of these categories:

## Changes

**Mistyped lookup maps** (`LanguageFieldRow.tsx`, `VolunteerProfile/constants.ts`,
`transformers.ts`)
The level maps were annotated as the `LanguageLevel` enum itself instead of
`Record<LanguageLevel, ...>`. Correcting the annotation removes the casts at both
definition and access sites and adds exhaustiveness checking: a new enum member now
forces a map entry. The access in `transformLanguagesToApi` needed a type predicate
on the existing filter so the compiler narrows `level` from `LanguageLevel | ""` to
`LanguageLevel` (same runtime logic as before).

**`getClearFilter` signature** (`helpers.ts` + 3 call sites)
The generic only appeared in the return type, so it could never be inferred and every
caller had to cast. Changed to `<T extends object>(filter: T): T`, which couples input
and output type. All three call sites (`Volunteers.tsx`, `Opportunities.tsx`,
`ClearAllFilters.tsx`) are now cast-free.

**Dead casts** (`VolunteerCard.tsx`, `AccordionVolunteer.tsx`)
`activities`/`skills` already are `string[]` via `getNormalizedVolunteer`, and
`VolunteerStateTypeType` is an alias of `ProfileVolunteeringType` in the current SDK.
These casts were leftovers and are removed without replacement.

**`statusMaps.ts`**
`Object.fromEntries` drops key types, which the two maps compensated with blanket
casts; the icon map's `as unknown as` meant even the values were unchecked. A small
typed `fromPairs<K, V>` helper now validates every pair at the definition site.

Two single `as T` assertions remain inside `getClearFilter` and `fromPairs`, where the
invariant (keys are preserved) is outside what the type system can express but locally
verifiable in two lines. No use-site assertions remain.

## Preventing regressions

The double-cast pattern can be banned via ESLint so it does not creep back in, e.g.
with `no-restricted-syntax`:

```json
{
  "no-restricted-syntax": ["error", {
    "selector": "TSAsExpression > TSAsExpression > TSUnknownKeyword",
    "message": "Avoid 'as unknown as' double assertions; fix the types instead."
  }]
}
```

(Stricter alternative: `@typescript-eslint/no-unsafe-type-assertion`, which flags all
unsafe assertions but requires type-aware linting.) Left out of this PR to keep it a
pure refactor; happy to add the rule here or in a follow-up if wanted.

## Verification

`yarn lint`, `yarn typecheck` and `yarn build` pass. Manually smoke-tested against the
local backend: clear-all filters on volunteers/opportunities, volunteer cards and
status badges, both profile pages, language level dropdown in the volunteer form, and
saving a language on the volunteer profile (PATCH payload carries the correct
`proficiency` value).

The diff is almost entirely type-level. The only runtime-visible edits are
behaviorally equivalent: a `Boolean(...)` wrapper in the filter predicate (required
for the type predicate; `filter` only evaluates truthiness) and the `fromPairs`
wrapper, which calls the same `Object.fromEntries` as before. No runtime behavior
change intended.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -11,6 +11,17 @@ const compat = new FlatCompat({
 
 const eslintConfig = [
   ...compat.extends("next/core-web-vitals", "next/typescript"),
+  {
+    rules: {
+      "no-restricted-syntax": [
+        "error",
+        {
+          selector: "TSAsExpression > TSAsExpression > TSUnknownKeyword",
+          message: "Avoid 'as unknown as' double assertions; fix the types instead.",
+        },
+      ],
+    },
+  },
 ];
 
 export default eslintConfig;
--- a/src/components/Dashboard/Opportunities/Opportunities.tsx
+++ b/src/components/Dashboard/Opportunities/Opportunities.tsx
@@ -69,7 +69,7 @@ export function Opportunities() {
   };
 
   const handleClearAllFilters = () => {
-    const cleared = getClearFilter(cardsFilter) as unknown as OpportunityCardsFilter;
+    const cleared = getClearFilter(cardsFilter);
     setCardsFilter(cleared);
     router.push(pathname + questionMark + serializeOpportunityFilters(cleared, searchParams));
   };
--- a/src/components/Dashboard/Profile/common/statusMaps.ts
+++ b/src/components/Dashboard/Profile/common/statusMaps.ts
@@ -47,10 +47,13 @@ type IconComponent = React.ComponentType<{ size?: number; color?: string }>;
 
 // Several SDK enums share the same underlying string values (e.g. "new", "active",
 // "pending-match"). Object literals with computed duplicate keys are a TS error, so
-// the SDK entries are built with Object.fromEntries — arrays have no such restriction and
+// the SDK entries are built from pairs — arrays have no such restriction and
 // the last entry for a given key wins, matching the original intended behaviour.
 // OpportunityManualStatusType values are unique ("opp-*") and use a typed Record
 // to preserve exhaustiveness checking.
+
+const fromPairs = <K extends string, V>(pairs: readonly (readonly [K, V])[]): Record<K, V> =>
+  Object.fromEntries(pairs) as Record<K, V>;
 
 const manualStatusColorMap: Record<OpportunityManualStatusType, string> = {
   [OpportunityManualStatusType.NEW]: "var(--color-violet-100)",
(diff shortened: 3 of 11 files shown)


nadavosa (commented):
1. **`as ApiLanguage` still in `transformLanguagesToApi`** — with `LEVEL_TO_PROFICIENCY` now typed as `Record<LanguageLevel, LangProficiency>` and `level` narrowed by the type predicate, check whether the mapped object fully satisfies `ApiLanguage` structurally. If it does, this cast can go too and the file would be fully clean.

2. **Add the ESLint rule now, not in a follow-up** — the PR description already has the exact config to ban `as unknown as`. Since this PR proves every occurrence was fixable, add the `no-restricted-syntax` rule here so the pattern can't creep back in immediately after merge.

thnndnly:
Thanks for the review. Both points addressed in 5c89eb7:

**1. `as ApiLanguage` in `transformLanguagesToApi`** — removed. With `level` narrowed to `LanguageLevel` by the type predicate and `LEVEL_TO_PROFICIENCY` typed as `Record<LanguageLevel, LangProficiency>`, the mapped literal already satisfies `ApiLanguage` structurally (`id: number`, `title: string`, `proficiency?: LangProficiency`, `purpose?: LangPurpose` — all present and assignable). I replaced the cast with an explicit `ApiLanguage[]` return type so the compiler verifies the shape instead of taking the assertion on trust. `yarn typecheck` passes, and the file now has zero use-site assertions.

**2. ESLint rule** — added to `eslint.config.mjs`:

```js
"no-restricted-syntax": ["error", {
  selector: "TSAsExpression > TSAsExpression > TSUnknownKeyword",
  message: "Avoid 'as unknown as' double assertions; fix the types instead.",
}]
```

Verified it targets exactly the double-cast pattern: `x as unknown as number` errors, while a legitimate single `x as string` does not.

PR #660, opened 2026-06-14
#569: NGO view:  restrict contact columns on volunteer list

### LINKED ISSUE #569: NGO view: Volunteer list - restrict visible columns
## Parent Epic
https://github.com/need4deed-org/fe/issues/568

## Description

When a user is authenticated as an NGO/agent, the volunteer list view must only show a restricted set of columns.

## Visible columns (NGO role)

| Column | Visible |
|---|---|
| First name | ✅ |
| Last name | ❌ |
| Volunteer type | ✅ |
| Engagement status | ✅ |
| Language | ✅ |
| Location | ✅ |
| Activities | ✅ |
| Schedule | ✅ |
| Email | ❌ |
| Phone | ❌ |
| Any other contact fields | ❌ |

## Acceptance Criteria

- [ ] Only the listed columns appear in the volunteer list for NGO-role users
- [ ] The restriction does not affect coordinator/admin views
- [ ] No contact details leak through tooltips, expanded rows, or exports

ivannissimrch:
## Description
Restricts the volunteer list for NGO/agent and volunteer roles to hide contact data. The email column and both copy actions (copy-all header and per-row) are gated behind an allowlist (`canSeeContactColumns = role === COORDINATOR || ADMIN`), so non-staff can't see or export volunteer emails. Last name is already hidden (first name only). Coordinator/admin views are unchanged.

## Related Issues
Closes #569 

## Changes
 - Hide email column for NGO/agent and volunteer roles
 - Remaining columns rebalance to restricted widths when email is dropped
 - Coordinator/admin views unchanged

**Open questions for @nadavosa** (raised in the issue, surfacing here):
- **Matching column**  currently still shown. Your spec lists only specific columns, and Matching isn't one. Hide it for the NGO, or keep it?
- **Activities and Schedule**  in your spec, but the table doesn't have that data (BE only sends it for the card view). Keep them out for now, or should they be added?

## Screenshots / Demos
Screenshot from 2026-06-14 07-51-16

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes

### DIFF
--- a/src/components/Dashboard/Volunteers/VolunteerListController.tsx
+++ b/src/components/Dashboard/Volunteers/VolunteerListController.tsx
@@ -3,13 +3,14 @@ import { useEffect } from "react";
 import { DashboardListLoading } from "@/components/Dashboard/common/DashboardListLoading";
 import { apiPathVolunteer, cacheTTL, CARD_COLUMNS, CARD_LIMIT, CARD_ROWS, TABLE_LIMIT } from "@/config/constants";
 import { useGetQuery, usePageParam } from "@/hooks";
-import { ApiOptionLists, ApiVolunteerGetList, SortOrder } from "need4deed-sdk";
+import { ApiOptionLists, ApiVolunteerGetList, SortOrder, UserRole } from "need4deed-sdk";
 import { CardsFilter } from "./Filters/types";
 import { serializeFilters } from "./helpers";
 import { VolunteerCardList } from "./VolunteerCardList";
 import { VolunteerTableList } from "./VolunteerTableList";
 import { ViewMode } from "../common/types";
 import { useCopyVolunteerEmails } from "@/hooks/useCopyVolunteerEmails";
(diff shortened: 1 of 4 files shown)


thnndnly (approved):
Frontend side looks correct. The email column and both copy actions are gated on `canSeeContactColumns` (`COORDINATOR || ADMIN`), fail-closed, and the header columns and row cells gate email consistently so the table stays aligned. This sits cleanly on top of the server-side PII masking (be#667 / be#671), so it's the right view-layer complement rather than the security boundary itself.

One coordination point to flag before #569 can close as intended: `Person` (including the first name) is masked server-side for non-staff (be#667), but the issue explicitly asks for the first name to stay unmasked in the NGO view. There's no backend issue tracking that yet, so after this merges an NGO/agent will see masked first names plus no email in the list. The email part is done here; the first-name visibility depends on a backend unmask that should be filed and linked.

PR #668, opened 2026-06-14
feat: add show/hide toggle to password fields

### LINKED ISSUE #663: feat: add show/hide toggle to password fields
### Summary
There is no way to reveal an entered password before submitting, so users cannot
verify what they typed.

### Affected fields
- Login (`src/components/Login/LoginForm.tsx`)
- Agent registration, password + confirm password
  (`src/components/AgentRegistration/steps/AccountStep.tsx`)

All three use the shared `src/components/core/common/FormInput.tsx`.

### Proposed fix (central, covers all fields)
Add a `showPassword` state + an Eye/EyeSlash toggle (icons already available via
`@phosphor-icons/react`), rendered only when `type === "password"`. Effective
type becomes `type === "password" && showPassword ? "text" : type`. The toggle
button must use `type="button"` (so it does not submit the form) and an i18n
`aria-label`.

thnndnly:
## Summary
Password fields had no way to reveal the entered value before submitting.

Adds an eye / eye-slash toggle inside the shared `FormInput`, rendered only
for `type="password"`. Covers the login form and agent registration (password
and confirm password). The toggle is a `type="button"` (so it never submits the
form) with an i18n `aria-label` (new `formInput.showPassword` / `hidePassword`
keys in en and de).

Closes #663

### DIFF
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -1950,5 +1950,9 @@
       "title": "Beitrittsanfrage gesendet",
       "description": "Ein Administrator prüft deine Anfrage, dieser Organisation beizutreten. Nach der Freigabe erhältst du Zugriff."
     }
+  },
+  "formInput": {
+    "showPassword": "Passwort anzeigen",
+    "hidePassword": "Passwort verbergen"
   }
 }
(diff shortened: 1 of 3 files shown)


nadavosa (approved):
Good implementation. `type="button"` on the toggle prevents accidental form submission, `aria-label` covers accessibility, the extra right padding on the input avoids text running under the icon, and both EN and DE translation strings are included. Nothing missing here.

PR #678, opened 2026-06-14
fix(#673): pre-fill matched org address on confirm

### LINKED ISSUE #673: bug: registration — selected org address not pre-populated, shows partial typed text instead
Part of #672.

## What happens

On /register/agent/complete (Step 1 — address), the user types a partial street like "elen". The lookup finds a matching org (e.g. "Elsenstr. 123") and shows a banner with "Use this org". The user clicks it.

After clicking, the confirmation view appears — but the address field still holds "elen" (what the user typed) instead of the full matched address or title. The user never sees what was actually selected pre-populated in the field.

Note: agents sometimes have the address stored only in the agent title field, not in the address entity. The lookup searches both. So the matched value may be a full street ("Elsenstr. 123") or a full org title — either way the whole thing should be shown, not the partial text the user started typing.

## Fix

**File:** src/components/AgentRegistration/ProfileCompletion/useAgentAddressLookup.ts

In confirmMatch(), after setting selectedAgent, also update the address field to the full matched value:

The hook currently does:
```ts
const confirmMatch = () => {
  if (matched) setSelectedAgent(matched);
};
```

confirmMatch needs to also surface the full matched title so ProfileCompletion can call update({ addressStreet: matched.title }). Either:
- Return a callback that ProfileCompletion can use, or
- Accept an onConfirm callback in the hook that ProfileCompletion passes in

**File:** src/components/AgentRegistration/ProfileCompletion/ProfileCompletion.tsx

Pass an onConfirm to useAgentAddressLookup (or handle in the confirmMatch call site) that calls:
```ts
update({ addressStreet: matched.title })
```

so the address field is replaced with the full matched value on selection.

## No BE changes needed

thnndnly:
Closes #673

## Problem
On `/register/agent/complete` (Step 1), typing a partial street (e.g. "elen")
surfaces a matching org banner. After clicking "Yes, use this", the address
field still held the partial typed text instead of the full matched value, so
the user never saw what was actually selected.

## Fix
- `useAgentAddressLookup`: `confirmMatch` now invokes an optional `onConfirm`
  callback with the full match.
- `ProfileCompletion`: passes `(m) => update({ addressStreet: m.title })` so the
  address field is replaced with the matched org's full title on selection.

The matched value may be a full street or a full org title (the lookup searches
both); either way the whole value is now surfaced.

No BE changes.

## Test plan
- [x] `yarn typecheck`, `yarn lint` green
- [ ] Reviewer: type a partial street, confirm match, verify field shows the
      full matched value

### DIFF
--- a/src/components/AgentRegistration/ProfileCompletion/ProfileCompletion.tsx
+++ b/src/components/AgentRegistration/ProfileCompletion/ProfileCompletion.tsx
@@ -78,6 +78,7 @@ export function ProfileCompletion() {
   const { matched, selectedAgent, showBanner, confirmMatch, dismissMatch } = useAgentAddressLookup(
     formData.addressStreet,
     token,
+    (m) => update({ addressStreet: m.title }),
   );
 
   const update = (fields: Partial<ProfileCompletionData>) => {
--- a/src/components/AgentRegistration/ProfileCompletion/useAgentAddressLookup.ts
+++ b/src/components/AgentRegistration/ProfileCompletion/useAgentAddressLookup.ts
@@ -9,7 +9,11 @@ type AgentSearchMatch = { id: number; title: string };
 // existing agents at a matching address so they can JOIN their org instead of
 // creating a duplicate. Backed by the token-gated GET /agent/register/search
 // (the COORDINATOR-only GET /agent is not available to a registrant).
-export function useAgentAddressLookup(addressStreet: string, token: string | null) {
+export function useAgentAddressLookup(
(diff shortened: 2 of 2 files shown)


nadavosa (approved):
Minimal and correct. Adding `onConfirm` as an optional callback keeps `useAgentAddressLookup` generic, and wiring it to pre-fill `addressStreet` from the matched org title is exactly the right UX. The early return guard added to `confirmMatch` is also a nice defensive improvement.

PR #679, opened 2026-06-14
fix: surface registration token errors and block form on bad token (#674)

### LINKED ISSUE #674: bug: registration — Request to Join button gives no feedback, silent failure on bad token
Part of #672. Related BE context: need4deed-org/be#651 (token expiry).

## What happens

On /register/agent/complete, after selecting a matched org the user sees the join confirmation view with a "Request to Join" button. Clicking it appears to do nothing — no loading state, no visible error.

What is actually happening: the button calls submit({ agentId }) which requires a valid token from the URL query string. If the token is null or has expired (BE #651 is adding expiry), setSubmitError() is called — but the ErrorBanner renders at the top of the card, out of view, so the user never sees it. The button itself gives no feedback.

## Fix

**File:** src/components/AgentRegistration/ProfileCompletion/ProfileCompletion.tsx

- On component mount, if token is null show the error banner immediately — do not wait for the button click
- After a failed submit, scroll the ErrorBanner into view (ref + scrollIntoView) or render a second inline error message directly below the button
- Verify the button correctly re-enables after a failed submit (check the finally block in submit())

## No BE changes needed for the FE feedback fixes
The token expiry logic itself is BE #651 — but the silent failure is a FE UX problem independent of that.

thnndnly:
Closes #674

## Problem

On `/register/agent/complete`, a null or expired token made every submit fail silently: `setSubmitError()` fired but the ErrorBanner sits at the top of the card, often out of view, so the "Request to Join" button looked dead. The user could also fill in the whole form and only hit the failure at the final submit.

## Changes

`src/components/AgentRegistration/ProfileCompletion/ProfileCompletion.tsx`

- Show the missing-token error on mount, not just after a click.
- Add `showSubmitError`, which sets the error and scrolls the banner into view via `requestAnimationFrame` on every error, so a repeat click with an identical message still re-scrolls.
- Disable "Next" and "Submit / Request to Join" while the token is missing, so the form cannot be worked through only to fail at the end.
- The `finally` block in `submit()` already re-enables the button on every failure path; verified, left unchanged.

Uses the existing `agentRegistration.errors.missingToken` and `message.errorGeneric` translation keys. No new strings, no BE changes.

### DIFF
--- a/src/components/AgentRegistration/ProfileCompletion/ProfileCompletion.tsx
+++ b/src/components/AgentRegistration/ProfileCompletion/ProfileCompletion.tsx
@@ -13,7 +13,7 @@ import {
   ApiOptionLists,
 } from "need4deed-sdk";
 import { useRouter, useSearchParams } from "next/navigation";
-import { useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { validateCompletionStep } from "../helpers";
 import { ProgressBar } from "../ProgressBar";
@@ -69,6 +69,7 @@ export function ProfileCompletion() {
   // When a CREATE collides with an existing title, the API returns its id so we
   // can offer to JOIN it instead.
   const [titleConflictAgentId, setTitleConflictAgentId] = useState<number | null>(null);
+  const errorBannerRef = useRef<HTMLDivElement>(null);
 
   const { data: optionLists } = useGetQuery<ApiOptionLists>({
     queryKey: ["options"],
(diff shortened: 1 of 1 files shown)


nadavosa (approved):
Clean fix. The `showSubmitError` helper correctly consolidates `setSubmitError` + scroll-into-view, and the `useEffect` that fires on mount when the token is missing means the user sees the error immediately rather than only after an attempted submit. Disabling both the Next and Submit buttons when `tokenMissing` is the right call — no point letting the user fill out the form if the registration URL is broken.

PR #641, opened 2026-06-11
571: Add NGO/Volunteer view to Agent, Opportunity  lists

DarrellRoberts:
## Description
Adds NGO/Volunteer view to Agent & Opportunity lists (only `title`, `type` and `district` are visible and are `non-clickable`)

## Related Issues
closes #571 , cooicides with [PR ](https://github.com/need4deed-org/fe/pull/680) which adds a redirect on Agent profile

## Changes
- Bullet list of meaningful changes (optional)

#### Opportunities
image
image

#### Agents
image
image

## Checklist
- [ ] WITHIN THE SCOPE OF AN ISSUE; No unnecessary files included
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

PR #680, opened 2026-06-15
Redirect on Agent Profile for NGO and Volunteers

### LINKED ISSUE #576: NGO + Volunteer view: Agent profiles — no access, list only
## Parent Epic
https://github.com/need4deed-org/fe/issues/568
https://github.com/need4deed-org/fe/issues/631

## Description

NGO and volunteer users cannot open agent/organisation profiles at all. The agent list is the only view available to these roles, and it shows a restricted set of columns (see https://github.com/need4deed-org/fe/issues/571).

## Access rules

| Action | NGO role | Volunteer role |
|---|---|---|
| View agent list | ✅ (restricted columns) | ✅ (restricted columns) |
| Open / navigate to an agent profile | ❌ | ❌ |

## Acceptance Criteria

- [ ] Agent list rows are not clickable / have no link to the profile for NGO and volunteer roles
- [ ] Navigating directly to an agent profile URL while logged in as NGO or volunteer returns a 403 / redirects away
- [ ] Coordinator/admin view retains full profile access

DarrellRoberts:
## Description
Adds a redirect on `/agent/[id]` for users who aren't `ADMIN` or `COORDINATOR`

## Related Issues
Closes #576 , cooicides with [PR 641](https://github.com/need4deed-org/fe/pull/641) which adds unclickable links and restricted-list views

## Changes
To reproduce:

1. Login as `AGENT` or `VOLUNTEER`
2. Navigate to `.../dashbaord/agents/1`
3. You should be redirected to `.../dashboard/agents`
4. Change role as either `ADMIN` or `COORDINATOR`
5. Navigate to `.../dashbaord/agents/1`
6. You should not be redirected and will see profile

## Screenshots / Demos

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/src/app/[lang]/dashboard/agents/[id]/page.tsx
+++ b/src/app/[lang]/dashboard/agents/[id]/page.tsx
@@ -1,7 +1,19 @@
 import ProfileLayout from "@/components/Dashboard/Profile/ProfileLayout";
+import { getServerUserRole } from "@/hooks/api/getUserRole";
 import { RouteParams } from "@/types";
+import { UserRole } from "need4deed-sdk";
+import { cookies } from "next/headers";
+import { redirect } from "next/navigation";
 
 export default async function DashboardAgentPage({ params }: RouteParams) {
   const { id } = await params;
+
+  const cookieStore = await cookies();
+  const cookieHeader = cookieStore.toString();
+  const userRole = await getServerUserRole(cookieHeader);
+
+  if (!userRole || (userRole !== UserRole.COORDINATOR && userRole !== UserRole.ADMIN)) {
+    redirect(`/dashboard/agents`);
+  }
   return <ProfileLayout entityId={id} entityType="agent" />;
 }
--- a/src/hooks/api/getUserRole.ts
+++ b/src/hooks/api/getUserRole.ts
@@ -1,0 +1,27 @@
+import { fetchFn } from "@/hooks/api/utils";
+import { UserRole } from "need4deed-sdk";
+import { apiPathMe } from "@/config/constants";
+
+export interface ApiResponse<T> {
+  message: string;
+  data: T;
+  count: number;
+}
+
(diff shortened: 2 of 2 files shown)


nadavosa (commented):
Logic is correct and the server-side auth check is the right approach for the app router. One issue though: the redirect loses the language prefix.

```ts
redirect(`/dashboard/agents`);
```

Since this page is under `[lang]`, it should be:

```ts
const { id, lang } = await params;
// ...
redirect(`/${lang}/dashboard/agents`);
```

Otherwise users hitting an agent profile URL in German (`/de/dashboard/agents/123`) get redirected to `/dashboard/agents` which would either 404 or land on the wrong language.

nadavosa (commented):
Logic is correct and the server-side auth check is the right approach for the app router. One issue though: the redirect loses the language prefix.

```ts
redirect(`/dashboard/agents`);
```

Since this page is under `[lang]`, it should be:

```ts
const { id, lang } = await params;
// ...
redirect(`/${lang}/dashboard/agents`);
```

Otherwise users hitting an agent profile URL in German (`/de/dashboard/agents/123`) get redirected to `/dashboard/agents` which would either 404 or land on the wrong language.

DarrellRoberts:
thanks @nadavosa ! 

I'm going to have to argue with the AI I'm afraid. 

In the `middleware` we are tracking the locale via a cookie which automatically inserts the chosen locale if absent. Therefore, when you redirect, the `middleware` intercepts the request and inserts the preferred locale based on the cookie value (it changes when the user changes the language).

Adding the `lang` to the redirect actually causes a 404 error as it redirects to something like: `/de/dashboard/agents/de/dashboard/agents`  (because of the middleware).

If you've experienced issues when manually testing it though, let me know

Issue #675, opened 2026-06-14
Epic: NGO registration → profile → opportunity creation & management


nadavosa:
## Overview

End-to-end flow for an NGO user: register, verify email, complete profile, see their own NGO profile, create and manage opportunities. Currently broken at multiple points.

---

### Registration (no BE dependency — do first)
- #673 — Bug: selected org address not pre-populated after match, shows partial typed text
- #674 — Bug: Request to Join button gives no feedback on failure
- #676 — feat: add phone number field to registration form (blocked on need4deed-org/be#685)

### Post-login profile
- #677 — feat: /dashboard/profile page — NGO user sees their own organisation profile (blocked on need4deed-org/be#684)
- #657 — fix: agent profile address not editable, shows "Berlin" fallback (blocked on need4deed-org/be#683)

### Opportunity creation & management
- #445 — feat: NGO opportunity creation — fix pre-fill, fix submit crash, wire up full form

---

## BE dependencies

- need4deed-org/be#684 — GET /agent/me (unblocks #677 and opportunity pre-fill in #445)
- need4deed-org/be#685 — phone in registration schema (unblocks #676)
- need4deed-org/be#683 — addressStreet/addressPostcode in agent GET response (unblocks #657)
- need4deed-org/be#651 — registration token expiry (context for #674)

---

## Full user journey when all sub-issues are done

1. NGO visits /register/agent — enters name, email, password, phone
2. Confirms email → lands on /register/agent/complete with token
3. Types street address → system finds matching org → full org name/address pre-populated → clicks "Request to Join" → clear loading + success/pending feedback
4. Logs in → clicks Profile in sidebar → sees own NGO profile at /dashboard/profile
5. NGO profile shows real address (not "Berlin"), address is editable
6. Clicks "Create New Opportunity +" on home → form pre-populated with name, email, phone from profile → submits successfully → redirected to opportunity profile
7. Edits and manages their opportunity from the opportunity profile

Supersedes: #672

PR #683, opened 2026-06-15
fix(#682): stale is_logged_in cookie causes /me 401 to /refresh 400 loop

### LINKED ISSUE #682: fix: stale is_logged_in cookie causes /user/me 401 → /auth/refresh 400 loop after pending registration
Part of #675. Related: #328/#334 (same /me→refresh→redirect mechanism, fixed for
public pages via query-gating — does not cover the cookie case), #661 (no logout
path in UI), #674 (same component).

## What happens

After an agent registers and requests to join an org that lands PENDING, navigating
to any page that reads the auth hint produces an endless error loop in the console:

```
GET  /api/user/me?language=en   401 (Unauthorized)
POST /api/auth/refresh          400 (Bad Request)
```

One pair per `refetchOnWindowFocus`, i.e. every time the window regains focus.

## Root cause

The `is_logged_in` cookie is used as the auth hint that gates the `/user/me` and
`/agent/me` queries:

```ts
// useCurrentUser.ts:7, useGetCurrentAgent.ts:8
const hasAuthHint = getCookie("is_logged_in") === "true";
```

Two setup bugs make this hint lie:

1. **Set too early.** `AgentRegistration.tsx:74` sets `LOGGED_IN_COOKIE` right after
   `POST /user` (account created, still unverified/pending) — before any session
   exists. On a PENDING membership (`ProfileCompletion.tsx:121`) the backend issues
   no tokens, but the cookie already says `is_logged_in=true`.

2. **Never cleared.** `is_logged_in` is set in exactly one place and removed nowhere
   (no logout flow exists — see #661). So the false hint survives for its full
   `max-age=6000`.

Result: the hint keeps firing `/user/me` → 401 → interceptor refresh → 400, and
`refetchOnWindowFocus` retriggers it indefinitely. The interceptor's redirect to
`/login` (which would break the loop) is intentionally suppressed on `login`/`forms`
paths (`axios.ts:71`), so on those routes nothing ever clears the bad state.

## Regression origin

PR #563 (#444) set `is_logged_in` only AFTER `POST /auth/login` established a real
session. PR #582 (#580) dropped the auto-login step (email verification now sits in
between) but kept setting the cookie right after `POST /user` — so the hint is now
set without a session. This issue fixes that.

## Fix

**`src/config/axios.ts`** — in the `catch (refreshError)` block, before the redirect,
clear the auth hint so it self-heals after a failed refresh:

```ts
document.cookie = "is_logged_in=; path=/; max-age=0; SameSite=Lax; Secure";
```

**`src/components/AgentRegistration/AgentRegistration.tsx:74`** — do not set
`LOGGED_IN_COOKIE` on account creation. The hint belongs only where the backend
actually establishes a session (as `LoginController.tsx:25` does on `onLoginSuccess`).

Suggested: extract a `clearAuthHint()` / `setAuthHint()` helper in `utils/helpers`
so axios, login and a future logout (#661) share one definition.

## Acceptance

- After a PENDING registration, no repeating `/user/me` 401 / `/auth/refresh` 400 in
  the console on window focus.
- A failed token refresh clears `is_logged_in` so queries stop refiring.
- `is_logged_in=true` is set only after a real login / when tokens exist.

thnndnly:
Fixes #682. Part of #675.

## Problem
After an agent registers and the org membership lands PENDING, the
`is_logged_in` auth hint lies. It gates `/user/me` and `/agent/me`, but was
set right after `POST /user` (account created, no session yet) and was never
cleared anywhere. The false hint keeps firing `/user/me` 401, the interceptor
runs a refresh that returns 400, and `refetchOnWindowFocus` retriggers the
pair on every window focus. On `login`/`forms` paths the interceptor redirect
is suppressed by design, so nothing ever breaks the loop.

## Fix
Two independent layers, one source of truth for the cookie.

- `config/constants.ts`: replace the single `LOGGED_IN_COOKIE` string with
  shared building blocks (`AUTH_HINT_COOKIE_NAME`, `AUTH_HINT_COOKIE_ATTRS`,
  `AUTH_HINT_MAX_AGE`) so set and clear cannot drift apart. Deleting a cookie
  requires the same name and attributes as setting it, otherwise the clear is
  a no-op.
- `utils/helpers.ts`: add `setAuthHint()` / `clearAuthHint()` built from those
  constants. One definition shared by axios and login.
- `AgentRegistration.tsx`: stop setting the hint on account creation. No
  session exists at that point, this was the root cause.
- `LoginController.tsx` / `ProfileCompletion.tsx`: set the hint via
  `setAuthHint()` only where the backend established a session (the
  ProfileCompletion call sits after the PENDING early return).
- `config/axios.ts`: call `clearAuthHint()` in `catch (refreshError)` before
  the redirect, so a failed refresh self-heals and queries stop refiring.

## Known limitation
The `catch (refreshError)` block catches every refresh failure, not just an
auth rejection. So a transient `/auth/refresh` 500 or a network error also
clears the hint and logs the user out, same as a real 400. This is acceptable
for now only because all refresh failures funnel through this one catch-all
handler. Distinguishing transient server errors from real auth failures is a
separate follow-up, out of scope here.

### DIFF
--- a/src/components/AgentRegistration/AgentRegistration.tsx
+++ b/src/components/AgentRegistration/AgentRegistration.tsx
@@ -1,6 +1,6 @@
 "use client";
 import { Button } from "@/components/core/button";
-import { apiPathUser, LOGGED_IN_COOKIE } from "@/config/constants";
+import { apiPathUser } from "@/config/constants";
 import axios from "axios";
 import { UserRole } from "need4deed-sdk";
 import { useState } from "react";
@@ -71,7 +71,6 @@ export function AgentRegistration() {
         },
       });
 
-      document.cookie = LOGGED_IN_COOKIE;
       document.cookie = PENDING_ROLE_COOKIE;
       setIsSuccess(true);
     } catch (err) {
--- a/src/components/AgentRegistration/ProfileCompletion/ProfileCompletion.tsx
+++ b/src/components/AgentRegistration/ProfileCompletion/ProfileCompletion.tsx
@@ -1,7 +1,7 @@
 "use client";
 import { Button } from "@/components/core/button";
 import { FormInput } from "@/components/core/common";
-import { apiPathAgentRegister, apiPathOption, LOGGED_IN_COOKIE } from "@/config/constants";
+import { apiPathAgentRegister, apiPathOption } from "@/config/constants";
 import { useGetQuery } from "@/hooks";
 import axios from "axios";
 import i18next from "i18next";
@@ -38,6 +38,7 @@ import {
 import { defaultProfileCompletionData, ProfileCompletionData, TOTAL_COMPLETION_STEPS } from "../types";
 import { CheckMark, MatchBanner, MatchActions, SmallButton } from "./styled";
 import { useAgentAddressLookup } from "./useAgentAddressLookup";
+import { setAuthHint } from "@/utils/helpers";
 
 function buildNewAgent(formData: ProfileCompletionData): ApiAgentRegisterNew {
   return {
@@ -122,7 +123,7 @@ export function ProfileCompletion() {
         setPending(true);
         return;
       }
-      document.cookie = LOGGED_IN_COOKIE;
+      setAuthHint();
       router.push(`/${i18next.language}/dashboard`);
     } catch (err) {
       if (axios.isAxiosError(err) && err.response?.status === 409) {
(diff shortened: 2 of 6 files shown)


nadavosa (approved):
Clean fix for the auth loop. Splitting the monolithic cookie string into `setAuthHint`/`clearAuthHint` helpers is the right abstraction, and calling `clearAuthHint()` in the axios interceptor before the login redirect is exactly what breaks the stale-cookie loop. The `AgentRegistration` component dropping `setAuthHint` entirely also makes sense — the agent isn't approved yet so setting `is_logged_in` there would be premature.

PR #689, opened 2026-06-16
fix: restore agent table title column width so header and rows align

ivannissimrch:
## Description
The admin agent table's title column was missing a width, so the header (which flexes when it has no width) and the rows (fixed width) didn't line up. Restored width: AGENT_COL_WIDTHS.title to align them. 
Is isolated to the admin view; the NGO read-only view uses separate widths and is unaffected.

## Changes
- Add width: AGENT_COL_WIDTHS.title to the admin agent table's title column header

## Screenshots / Demos
before:
before

after: 
after
Screenshot from 2026-06-16 18-32-34

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes

### DIFF
--- a/src/components/Dashboard/Agents/agentsTableColumns.ts
+++ b/src/components/Dashboard/Agents/agentsTableColumns.ts
@@ -20,7 +20,7 @@ export const AGENT_READ_ONLY_COL_WIDTHS = {
 };
 
 export const createAgentTableColumns = (t: TFunction, copyButton: ReactNode): Column[] => [
-  { key: "title", label: t("dashboard.agents.table.title") },
+  { key: "title", label: t("dashboard.agents.table.title"), width: AGENT_COL_WIDTHS.title },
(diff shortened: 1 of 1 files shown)


nadavosa (approved):
Correct fix. The admin agent table passes `noFixedWidth={false}` (authorized path), so `col.width` in `EntityTableList` is applied to the `TableHeaderCell` — but `createAgentTableColumns` was missing `width` on the title column while `AgentTableRow` already had `$width={AGENT_COL_WIDTHS.title}` on the corresponding cell. Adding `width: AGENT_COL_WIDTHS.title` aligns the header with the rows.

One pre-existing note (not introduced here): the `numOpportunities` cell in `AgentTableRow` always renders `"—"` — the API field isn't consumed yet. Nothing to address in this PR, just flagging for tracking.

ivannissimrch:
I created an issue to keep track of the numOpportunities cell displaying "-" instead of the number of opportunities. 
issue #712

PR #693, opened 2026-06-17
feat: rework new opportunity page to match profile edit layout

### LINKED ISSUE #445: feat: agent can publish an opportunity within the app
Sub-issue of #675.

## Context

The opportunity creation form at /dashboard/opportunities/new exists but is broken in two ways:

1. Contact fields (name, email, phone) never pre-populate because GET /agent/me does not exist on the BE yet (need4deed-org/be#684). The useGetCurrentAgent() hook always returns undefined.

2. The form submission crashes — POST /api/opportunity/ does not exist on the BE. The only opportunity creation route is the public legacy form. A dashboard-specific endpoint needs to be added on the BE side.

## Blocked on

- need4deed-org/be#684 — GET /agent/me (pre-fill)
- BE issue to be opened — POST /opportunity/ dashboard endpoint (developer must check network tab, confirm the failing request, and open the BE issue)

## Fix

**Pre-fill (src/components/Dashboard/NewOpportunity/NewOpportunity.tsx):**
- Once BE #684 is merged, verify the useEffect pre-fill (lines 49-55) populates contactName, contactEmail, contactPhone correctly
- Add a loading skeleton while agentLoading is true
- Add a clear error state if agent is null after loading

**Submit crash:**
- Reproduce as an NGO user: open the page, fill the form, submit, check the network tab
- Identify the exact failing endpoint
- BE issue 236 for the missing POST /opportunity/ dashboard route
- Once the BE endpoint exists, wire it up and verify end-to-end

**After successful submit:**
- Verify redirect to the new opportunity profile page works
- Verify the NGO can see and edit their opportunity at /dashboard/opportunities/:id

nadavosa:
## Summary

- Replaces the simple card form with the same `PageContainer` + `SectionCard` layout used by the opportunity profile page
- Title is presented in a Wrench section card (same as Opportunity Details on the profile)
- Contact details are in a ChatsCircle section card, always in edit mode, pre-filled from the current agent's representative
- Validation uses zod + react-hook-form with the same phone regex and email rules as `opportunityContactDetailsSchema`
- Save POSTs `{ title, contact: { name, phone, email }, agentId }` to `POST /api/opportunity/` and redirects to the new opportunity's profile on success

Closes https://github.com/need4deed-org/fe/issues/445

## Test plan
- [ ] Log in as an agent; click "New Opportunity" from the dashboard
- [ ] Verify the page uses the profile-style section card layout (not the old card form)
- [ ] Verify contact name/phone/email are pre-filled from the logged-in agent's representative
- [ ] Verify title is empty and required; submit without it shows validation error
- [ ] Verify phone/email validation works the same as on the opportunity profile contact section
- [ ] Fill in all fields and submit; confirm redirect to the new opportunity's profile page

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "date-fns": "^4.1.0",
     "email-validator": "^2.0.4",
     "i18next": "^25.3.2",
-    "need4deed-sdk": "0.0.108",
+    "need4deed-sdk": "0.0.109",
     "next": "15.3.8",
     "react": "^19.0.0",
     "react-day-picker": "^9.13.0",
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "date-fns": "^4.1.0",
     "email-validator": "^2.0.4",
     "i18next": "^25.3.2",
-    "need4deed-sdk": "0.0.109",
+    "need4deed-sdk": "0.0.110",
     "next": "15.3.8",
     "react": "^19.0.0",
     "react-day-picker": "^9.13.0",
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "date-fns": "^4.1.0",
     "email-validator": "^2.0.4",
     "i18next": "^25.3.2",
-    "need4deed-sdk": "0.0.107",
+    "need4deed-sdk": "0.0.108",
     "next": "15.3.8",
     "react": "^19.0.0",
     "react-day-picker": "^9.13.0",
--- a/src/components/Dashboard/Home/AgentOpportunityCards.tsx
+++ b/src/components/Dashboard/Home/AgentOpportunityCards.tsx
@@ -1,0 +1,42 @@
+"use client";
+import { apiPathAgent, apiPathOption, cacheTTL } from "@/config/constants";
+import { useGetCurrentAgent } from "@/hooks/useGetCurrentAgent";
+import { useGetQuery } from "@/hooks";
+import { ApiOptionLists, ApiOpportunityGetList, ApiVolunteerOpportunityGetList } from "need4deed-sdk";
+import { OpportunityCard } from "../Opportunities/OpportunityCard";
+import { useTranslation } from "react-i18next";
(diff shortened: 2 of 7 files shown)


nadavosa (commented):
## Review

Overall the approach is solid — three-form aggregation, profile-style layout, role-aware home page, correct SDK upgrade path. A few things worth addressing before merge:

**Issues:**

1. **`AccompanyingDetailsEdit` no-op handlers** (`NewOpportunity.tsx` ~line 823): `onCancel={() => {}}` and `onSubmit={() => {}}` mean the Cancel and Submit buttons inside that component render and are clickable but do nothing. The agent will click them expecting something to happen. Either suppress those buttons (if `AccompanyingDetailsEdit` supports that) or document that submission is driven by the outer Create button only.

2. **Empty state missing in `AgentOpportunityCards`**: When `opportunities` is an empty array (a freshly registered agent with no opportunities yet), the component renders nothing — no message. Add a simple empty-state string so the home page isn't just a button with a blank area below it.

3. **`category` / `category_id` sent as `""`** (`buildCreatePayload`): These are required fields in `OpportunityFormDataWithAgentSubmitter` and we're sending empty strings. If the BE rejects blank values here a creation attempt will silently fail. Worth confirming with BE what the correct sentinel is (`null`? omit entirely? a real default?).

**Minor notes:**

- `keyToLabel` / `labelToKey` maps are rebuilt on every render in the component body — should be `useMemo(() => ..., [apiLanguages])` since they only change when `apiLanguages` changes.
- `opportunity_type: isAccompanying ? "accompanying" : "volunteering"` — events map to `"volunteering"`. The SDK type only has two values so this is correct, but it's worth a short comment that the BE distinguishes events from regular via `onetime_date_time`.
- `useGetCurrentAgent` in `AgentOpportunityCards` fires two requests (`/me` → `/agent/{id}`) when we only need the `agentId` from `/me`. Not a blocker but a minor over-fetch.

PR #694, opened 2026-06-17
removes agent redirect for now

DarrellRoberts:
## Description
Removes redirect when accessing `/agents/[id]`

## Related Issues
Closes #ISSUE_ID (mandatory if applicable)

## Changes

## Screenshots / Demos

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/src/app/[lang]/dashboard/agents/[id]/page.tsx
+++ b/src/app/[lang]/dashboard/agents/[id]/page.tsx
@@ -1,19 +1,8 @@
 import ProfileLayout from "@/components/Dashboard/Profile/ProfileLayout";
-import { getServerUserRole } from "@/hooks/api/getUserRole";
 import { RouteParams } from "@/types";
-import { UserRole } from "need4deed-sdk";
-import { cookies } from "next/headers";
-import { redirect } from "next/navigation";
 
 export default async function DashboardAgentPage({ params }: RouteParams) {
   const { id } = await params;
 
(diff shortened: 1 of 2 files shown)


nadavosa (commented):
## Review

Two sensible changes: removing the server-side role guard (client-side auth in profile sections handles visibility) and making the read-only table row navigable for NGO agents.

**One issue to fix before merge:**

`AgentReadOnlyTableRow` now has an `onClick` handler but `$cursor` is still `"auto"`. This means no pointer cursor on hover, which makes the row feel non-interactive. Change to `$cursor="pointer"`:

```tsx
- <ClickableRow $isLast={isLast} $cursor={"auto"} data-testid={`agent-row-${id}`} onClick={handleGoToProfile}>
+ <ClickableRow $isLast={isLast} $cursor="pointer" data-testid={`agent-row-${id}`} onClick={handleGoToProfile}>
```

**Minor notes:**
- PR description still has the placeholder `#ISSUE_ID` — worth linking the actual issue.
- The "for now" in the title implies the redirect will come back. If the plan is to replace it with client-side auth (which the profile sections already handle via `isAuthorized`/`isOwnProfile`), consider noting that in the description so reviewers aren't left wondering.

DarrellRoberts:
done, I've removed the Clickhandler like it was before. This prevents NGOs from clicking on other NGOs profiles

PR #696, opened 2026-06-17
🐛 fix: align POST /opportunity body with OpportunityFormDataWithAgentSubmitter

arturasmckwcz:
## Problem

`NewOpportunity` (the agent-authenticated "create opportunity" screen) posted to `POST /opportunity/` with a body that didn't match the SDK contract `OpportunityFormDataWithAgentSubmitter`:

- The agent was sent as `agentId` (camelCase). The endpoint expects **`agent_id`** (snake_case), confirmed against both the SDK type (bumped to `need4deed-sdk@0.0.110`) and the runtime swagger. Since the schema is `additionalProperties: true`, the backend silently dropped the unknown key — so **opportunities were never linked to their agent**.
- The body also included a `contact` object that isn't part of the create contract, so the agent's contact details were silently dropped too.

## Fix

- Rename `agentId` → `agent_id` so the agent link actually binds.
- Keep the create body to contract fields only (`title`, `agent_id`).
- Persist contact via a follow-up `PATCH /opportunity/:id` (`ApiOpportunityPatch.contact`), chained inside a single `mutationFn` so there's one toast and one navigation.
- The contact PATCH is **best-effort**: if it fails, the create still succeeds and the agent lands on the opportunity profile where contact is editable.
- Bump `need4deed-sdk` to `0.0.110`.

## Testing

- `yarn typecheck` ✅
- `yarn lint` ✅ (only pre-existing warnings in unrelated files)

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "date-fns": "^4.1.0",
     "email-validator": "^2.0.4",
     "i18next": "^25.3.2",
-    "need4deed-sdk": "0.0.107",
+    "need4deed-sdk": "0.0.110",
     "next": "15.3.8",
     "react": "^19.0.0",
     "react-day-picker": "^9.13.0",
--- a/src/components/Dashboard/NewOpportunity/NewOpportunity.tsx
+++ b/src/components/Dashboard/NewOpportunity/NewOpportunity.tsx
@@ -104,9 +104,11 @@ export function NewOpportunity() {
 
   const handleSubmit = () => {
     if (!validate()) return;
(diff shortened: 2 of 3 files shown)


arturasmckwcz:
Superseded by #693, which rewrote `NewOpportunity.tsx` (new profile-layout form) and already bumped need4deed-sdk to 0.0.110. The agent-link fix is still needed on the new code (`buildCreatePayload` hardcodes `agent_id: null`) and will be addressed in a fresh PR against the rewritten component.

PR #697, opened 2026-06-17
🐛 fix: link created opportunity to the current agent

arturasmckwcz:
## Problem

The rewritten `NewOpportunity` form (from #693) builds its payload with `buildCreatePayload`, which hardcodes:

```ts
agent_id: null,
submitted_by_id: null,
```

and the component never fetches the current agent. So every opportunity created through this screen is sent to `POST /opportunity/` **unlinked to any agent**, even though the contract (`OpportunityFormDataWithAgentSubmitter`) carries `agent_id`.

This supersedes #696 (closed), which fixed the same class of bug on the pre-rewrite version of this file.

## Fix

- Fetch the logged-in agent via `useGetCurrentAgent` (`GET /agent/me`).
- Pass `agent.id` into `buildCreatePayload`, setting `agent_id` on the payload.
- Gate submit until the agent is loaded (`disabled={isPending || agentLoading || !agent}`) and bail in `onSubmit` if there's no agent id, so we never create an unlinked opportunity under a race.

`submitted_by_id` is left as `null` — its source/semantics aren't established here and the backend may derive it from the authenticated session. Happy to wire it up in a follow-up if it should come from the agent representative.

## Testing

- `yarn typecheck` ✅
- `yarn lint` ✅ (no new warnings)

### DIFF
--- a/src/components/Dashboard/NewOpportunity/NewOpportunity.tsx
+++ b/src/components/Dashboard/NewOpportunity/NewOpportunity.tsx
@@ -392,7 +392,7 @@ export function NewOpportunity() {
   const lang = i18n.language;
   const locale = lang === "de" ? de : enUS;
   const router = useRouter();
-  const { agent, isLoading: agentLoading } = useGetCurrentAgent();
+  const { agentId } = useGetCurrentAgent();
   const volunteerTypeLabelMap = createVolunteerTypeLabelMap(t);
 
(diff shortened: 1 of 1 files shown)


need4deed (approved):
Approving: agent-link fix reviewed (uses agentId from /me; submit gated on agentId). LGTM.

PR #699, opened 2026-06-18
695: Fixes redirect on /agent/[id]

### LINKED ISSUE #695: fix: redirect on /agent/[id]
### Summary

On production and staging when the `agent/:id` redirect was live, the user even if they were `Coordinator` or `Admin` would be redirected and unable to access the profile.

### Steps to Reproduce

1. Paste the following in `src/app/[lang]/dashboard/agents/[id]/page.tsx`
```typescript
import ProfileLayout from "@/components/Dashboard/Profile/ProfileLayout";
import { getServerUserRole } from "@/hooks/api/getUserRole";
import { RouteParams } from "@/types";
import { UserRole } from "need4deed-sdk";
import { cookies } from "next/headers";
import { redirect } from "next/navigation";

export default async function DashboardAgentPage({ params }: RouteParams) {
  const { id } = await params;

  const cookieStore = await cookies();
  const cookieHeader = cookieStore.toString();
  const userRole = await getServerUserRole(cookieHeader);

  if (!userRole || (userRole !== UserRole.COORDINATOR && userRole !== UserRole.ADMIN)) {
    redirect(`/dashboard/agents`);
  }
  return <ProfileLayout entityId={id} entityType="agent" />;
}
```
2. Change role of account to either `COORDINATOR` or `ADMIN` (if not already)
3. Navigate to an agent profile
4. Observe redirect

### Expected Behavior

If user is `COORDINATOR` or `ADMIN`, there should be no redirect

If user is `AGENT` or `VOLUNTEER`, there should be a redirect

### local dev
If user is `COORDINATOR` or `ADMIN`, there is no redirect

If user is `AGENT` or `VOLUNTEER`, there is a redirect

### staging
If user is `COORDINATOR` or `ADMIN`, there is a redirect

If user is `AGENT` or `VOLUNTEER`, there is a redirect

### production
If user is `COORDINATOR` or `ADMIN`, there is a redirect

If user is `AGENT` or `VOLUNTEER`, there is a redirect

### Environment

Incorrect redirect was recorded on:

- Firefox, ubuntu

DarrellRoberts:
## Description
Redirects user is they land on `/agent/:id` without sufficient permissions.

The authorized accounts/permissions are:

- Admin
- Coordinator
- Agent with matching `personId`

## Related Issues
Closes #695 

## Changes

- Used a two layer approach:

1. Adds `role` check on `middleware` which only redirects if user is on path `/agent/[id]` and is neither `admin` , `coordinator` nor `agent`
2. Adds serverside functions on `/agent/[id]` with a conditional check and only redirects if the user is an `agent` but their `personId` does not match the agent's representative `personId`

To test:

There should be no redirect for:
- Use `admin` account and land on `/dashboard/agent/[id]`. 
- Use `coordinator` account and land  `/dashboard/agent/[id]`.
- Use `agent` account and land on the `agentId` to which you are the representative for. 

There should be a redirect for:
- Use `agent` account and land on any `agentId` to which you are not the represntative for 
- Use `volunteer` account and land on `/dashboard/agent/[id]`
- Use `user` account and land on `/dashboard/agent/[id]`
- Don't login and land on `/dashboard/agent/[id]`

## Screenshots / Demos

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes


nadavosa:
## Review findings

**1. `src/middleware.ts:24-25` — unguarded null dereference can crash every page**
`decodeJwtPayload` returns `null` on any decode failure (see catch block in `helpers.ts`), but line 25 immediately does `userObject.role === UserRole.ADMIN` with no null check. Since the middleware `matcher` runs on virtually every route, any request carrying a `refresh` cookie that isn't a valid 3-segment JWT (corrupted, truncated, stale format) throws a `TypeError` and 500s the whole site for that user — not just `/dashboard/agents/[id]`.

**2. `src/app/[lang]/dashboard/agents/[id]/page.tsx:15-18` — fails open if `getServerUser`/`getServerAgent` errors**
Both helpers swallow fetch errors and return `null`. If `getServerUser` returns `null` due to a transient backend/network failure, `user?.role === UserRole.AGENT` evaluates to `false` and the redirect is skipped entirely — an agent could view another agent's profile during exactly the kind of failure this guard exists to handle.

**3. `src/hooks/api/getUser.ts` — duplicates existing `getServerUserRole`**
`src/hooks/api/getUserRole.ts` already fetches `/user/me` with the same cookie-header pattern and returns the role; `ApiResponse<T>` is now defined identically in four places. Consider reusing/extending the existing helper instead of re-implementing the same fetch+try/catch+type again.

**4. `src/hooks/api/getAgent.ts:25` — copy-pasted error message**
The catch block logs `"Failed to fetch server user role:"` even though this function fetches an agent, not a user role — likely a leftover from copy-pasting `getUser.ts`.

**5. `src/app/[lang]/dashboard/agents/[id]/page.tsx:15-16` — independent fetches awaited sequentially**
`getServerUser` and `getServerAgent` don't depend on each other but are awaited one after another, doubling this page's SSR latency vs. `Promise.all([...])`.

**6. `src/middleware.ts:14` — misleading config key**
`authorizedRoutes.NGO` is keyed `"NGO"` but there is no `NGO` role in `UserRole` (`user`, `coordinator`, `agent`, `volunteer`, `admin`) — it actually guards the agent-detail route. Harmless today but will confuse whoever extends this map next.

src/hooks/api/getUser.ts
```diff
@@ -0,0 +1,30 @@
...
+import { ApiUserGet } from "need4deed-sdk";
+
+export interface ApiResponse<T> {
+  message: string;
+  data: T;
+  count: number;
+}
+
+type ApiUserMe = ApiUserGet & { agentId: string | number };
+
```
DarrellRoberts, line 12: adding `agentId` as expecting it on `/user/me` when `sdk` is updated

DarrellRoberts:
thanks @nadavosa , I've updated the PR. 
We should now have `agentId` on the `user` response from `be` which actually makes this a lot simpler so I remove some unneeded files.

The failing Lint & Typecheck below is unrelated (regards unrelated files inherited from merging with `develop`)

PR #709, opened 2026-06-19
704: bug, fix district filter checkbox on table and cards

### LINKED ISSUE #704: bug: district (Bezirk) filter checkbox has no effect — table still shows all districts
## Description

On the opportunities table (and likely volunteers table), selecting a district in the Bezirk filter panel does not filter the rows. Checking e.g. \"Charlottenburg-Wilmersdorf\" leaves rows with \"Mitte\" and other districts still visible in the table.

## Steps to reproduce

1. Go to the dashboard opportunities or volunteers list
2. Open the filter panel
3. Check any district under **Bezirk**
4. Observe: table rows are unchanged — other districts remain visible

## Expected behaviour

Checking one or more districts should filter the table to show only rows whose Bezirk column matches a checked district.

## Observed behaviour

The checkbox is visually checked but the table data does not change. All districts continue to appear regardless of which filters are selected.

## Likely location

The filter state is probably applied correctly to the UI (checkbox renders as checked) but not passed to the query or not used to filter the displayed data. Likely area to investigate:

- Filter state management for the Bezirk field in the opportunities/volunteers filter component
- Whether the selected district IDs are forwarded to the API query params or client-side filter predicate
- `src/components/Dashboard/Opportunities/` or `src/components/Dashboard/Volunteers/` — filter hooks or filter panel components

## Notes

- The Typ and Status filters may or may not have the same issue — worth checking during the fix
- This is a frontend bug; the API likely supports district filtering already via query params

DarrellRoberts:
## Description
Changes  district to location on display which is what it is actually filtering

## Related Issues
Closes #704 

## Changes

## Screenshots / Demos
image

image

image

image

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/src/components/Dashboard/Opportunities/OpportunityCard.tsx
+++ b/src/components/Dashboard/Opportunities/OpportunityCard.tsx
@@ -38,7 +38,7 @@ export function OpportunityCard({ opportunity, volunteerId, activitiesList, dist
     availability,
     accompanyingDetails,
     statusMatch,
-    district,
+    location,
   } = opportunity as ApiVolunteerOpportunityGetList & {
     accompanyingDetails?: { appointmentDate?: string; appointmentTime?: string };
     statusMatch?: string;
@@ -48,7 +48,7 @@ export function OpportunityCard({ opportunity, volunteerId, activitiesList, dist
   const mainCommunication = getLanguagesByPurpose(languages, LangPurpose.GENERAL);
   const recipientLanguage = getLanguagesByPurpose(languages, LangPurpose.RECIPIENT);
   const activityTitles = getActivityTitles(activities, activitiesList);
-  const districtTitle = district?.id ? (districtsList?.find((d) => d.id === district.id)?.title ?? null) : null;
+  const districtTitle = location[0]?.id ? (districtsList?.find((d) => d.id === location[0].id)?.title ?? null) : null;
 
   const isAccompanying = volunteerType === ProfileVolunteeringType.ACCOMPANYING;
   const scheduleText = isAccompanying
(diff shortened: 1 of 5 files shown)


ivannissimrch (commented):
Running this branch throws Cannot read properties of undefined (reading '0') in OpportunityCard.tsx:51 when an opportunity has no location; the BE omits it sometimes.  After editing  adding `"?"` after `location` on  `const districtTitle = location?.[0]?.id ? (districtsList?.find((d) => d.id === location[0].id)?.title ?? null) : null;`  I tested: opportunities filter correctly now, but the volunteer district filter is still broken; checking a district doesn't filter the rows, exactly as the issue describes. Should this PR also cover volunteers?

ivannissimrch:
Update: Discard the "Cannot read properties" error. I rebuilt my local DB, and the crash is gone. It was triggered by stale local data (old opportunities with no location), not the live BE

DarrellRoberts:
thanks for reviewing @ivannissimrch ! So can I confirm, are you still experiencing issues with filtering `opportunities` ?

DarrellRoberts:
as for `volunteers` , I'm not 100% sure if there is an error as with `volunteers` they have so many districts that it's possible that they are matching, it's just not immediately visible until you go into the volunteer profile. e.g. a Volunteer could have 10 districts on their profile, one of which is say Mitte, so they match for Mitte but on the Table and Cards, there are the 9 other districts as well. If you are seeing an issue with it though, I can create a seperate issue & PR

ivannissimrch:
Filtering opportunities is working correctly.  Oh, I see what you say about volunteers, yes, I could be that.

ivannissimrch (approved):
Opportunities filter correctly, and the volunteer behavior makes sense per your explanation.

PR #711, opened 2026-06-19
feat: activity log (date and hours per session) for active matches

### LINKED ISSUE #577: Volunteer & Opp profiles: Activity time tracking - date + hours log in Active matching tab
## What this is

An activity log — date + hours per volunteering session — tied to a specific volunteer↔opportunity match. Appears in both profiles but is only editable from the opportunity profile. Shown when the match status is **Active** (editable) or past (read-only, to preserve the historical record and feed the future appreciation tier).

---

### Opportunity profile — editable (source of truth)
Inside the **Active** accordion row in the OpportunityVolunteers section (currently `VolunteerDetail.tsx`), when `currentStatus === ACTIVE`:

- Section renders below the volunteer's detail info and above the action buttons
- Table with two columns: **Date** | **Hours** (e.g. 3, 2.5)
- Each row has edit + delete action buttons (same pattern as Appreciation / CommunicationTracker)
- **"+ Add entry"** button opens an inline form or dialog: date picker + hours input (decimals allowed)
- Footer row: **Total: X hours**
- Editable by: Coordinator and NGO roles. Read-only for all others.

### Volunteer profile — read-only mirror
Inside the accordion row in the VolunteerOpportunities section (currently `OpportunityDetail.tsx`), when `currentStatus === ACTIVE` **or past**:

- Same table: **Date** | **Hours** — no edit/delete buttons
- Footer row: **Total: X hours**
- Data is the same records — no separate storage

---

## Future (out of scope now, but design for it)
The Appreciation section currently requires manual entry of t-shirt / bag / certificate. In the future it should derive from the total volunteering hours (e.g. 10h → t-shirt, 25h → bag, 50h → certificate). The hours total from this section is the input. Note this in code but don't implement yet.

---

## FE tasks

- [ ] Create shared `ActivityLog` component (table + add/edit form + total row)
- [ ] Render read-only `ActivityLog` in `OpportunityDetail.tsx` when `currentStatus === ACTIVE` or past
- [ ] Render editable `ActivityLog` in `VolunteerDetail.tsx` when `currentStatus === ACTIVE`
- [ ] Hook: `useActivityLog(opportunityVolunteerId)` — GET list, POST, PATCH, DELETE
- [ ] Add/edit dialog: date picker (reuse existing `react-day-picker`) + hours number input
- [ ] Display total hours sum below the table
- [ ] NGO/volunteer read-only enforcement (follow #568 auth pattern)
- [ ] Translations: en + de

## Depends on
- need4deed-org/be#679 — ActivityLog stored on the intermediate join entity (OpportunityVolunteer / Deal)
- #568 (NGO visibility rules)

thnndnly:
Closes #577

## What
Adds an activity log (date and hours per volunteering session) tied to a volunteer↔opportunity match, per #577. It appears in both profiles but is editable only from the opportunity profile, and is shown only while the match status is `active`, following the same conditional pattern as `AccompanyingDetails`.

## Where it appears
- Opportunity profile, OpportunityVolunteers section, active match accordion: editable source of truth (table, add/edit/delete dialog, total row).
- Volunteer profile, VolunteerOpportunities section, active match accordion: read only mirror of the same data.

## Implementation
- New shared `ActivityLog` component (`sections/ActivityLog/`): table with Date and Hours, inline add/edit dialog (date picker plus hours input with decimal support and validation), per row edit and delete, total row aligned under the Hours column.
- `useActivityLog(opportunityId, volunteerId)` hook wrapping `useGetQuery` and `useMutationQuery`, matching the `useAppreciationTracker` pattern. GET and POST are nested under the match, PATCH and DELETE address an entry by id.
- Edit is gated to Coordinator and Admin via `useAuth`. Other roles get the read only view.
- i18n keys under `dashboard.activityLog` for en and de.
- Accessibility: icon buttons carry `aria-label` and `type="button"`, the save button is disabled while the form is invalid, the hours field sets `aria-invalid`.

## Backend dependency (need4deed-org/be#679)
Built swap ready against the contract in need4deed-org/be#679 (https://github.com/need4deed-org/be/issues/679). Interim local types mirror `ApiActivityLog{Get,Post,Patch}` exactly, so finalizing is a single change in `types.ts` once the SDK ships them. The CRUD endpoints and SDK types are not in place yet, so the live table stays empty until the backend lands.


nadavosa:
## Review

 From me and claude (so take it with a dash of salt):
 
**1. Edit dialog corrupts the date for any user west of UTC (confirmed — data loss)**
`new Date(initialData.date)` in `ActivityLogDialog.tsx:31` parses a `yyyy-MM-dd` string as midnight UTC. A coordinator in UTC-5 opening an edit for `"2025-01-15"` gets `2025-01-14T19:00` local — the date picker pre-selects the 14th. Saving without re-picking the date PATCHes `"2025-01-14"` permanently. Fix: use `parseISO(initialData.date)` from `date-fns` (already a dependency), which parses date-only strings as local midnight.

**2. All log table dates display one day early in UTC- browsers (same root cause)**
`formatDate(entry.date)` in the table goes through `shared/utils/formatDate.ts` which also uses `new Date(string)`. Every date in the log renders as the day before what's stored for users outside UTC+. Display-only, but the entire activity history looks wrong. Same fix: use `parseISO` in `formatDate`.

**3. `isLoading` unused — "no hours logged" flashes on every accordion open**
`useActivityLog` returns `isLoading` but `ActivityLog.tsx` doesn't use it. While the GET is in-flight, `entries` defaults to `[]` and the empty state message shows. A coordinator who opens the accordion quickly may hit "Add entry" before existing data appears. Fix:
```tsx
{!isLoading && entries.length === 0 ? <SectionEmptyState>...</SectionEmptyState> : ...}
```

**4. Save and Delete not disabled during in-flight mutations**
`isCreating`, `isUpdating`, `isDeleting` are returned from `useActivityLog` but unused in the component. Double-clicking Save fires two identical POSTs; double-clicking Delete fires two DELETEs (the second will 404 with an error toast). Fix: pass `disabled={isCreating || isUpdating}` to the Save button and guard the delete handler with `if (isDeleting) return`.

thnndnly:
Thanks, all four are valid. Fixed in `5d827c28` (now on the branch as `f1a01fd6` after rebasing onto the latest develop merges):

1. **Edit dialog date corruption** — dialog now parses `initialData.date` with `parseISO` instead of `new Date()`, so the picker pre-selects the stored day in local time.
2. **Table dates one day early** — `shared/utils/formatDate.ts` switched to `parseISO` for the same reason; date-only strings are no longer read as UTC midnight.
3. **Empty state flashing** — the empty state is now gated on `!isLoading && entries.length === 0`, so it no longer shows while the initial GET is in flight.
4. **Double submit** — Save is `disabled` while a create/update is in flight (`isSaving={isCreating || isUpdating}`), and `confirmDelete` returns early when `isDeleting`, so a second POST/DELETE can't fire.

typecheck and lint are green.

ivannissimrch:
## I tested using an admin account:
- The functionality for adding, editing, and deleting entries is working correctly; the Total row accurately sums the hours.
- Date handling has been validated; when entries are added and the page is reloaded, the dates remain consistent without any timezone shifts. Additionally, reopening the edit dialog correctly pre-selects the stored date.  
- The empty state renders cleanly when the page loads, with no flickering before the data is displayed.  
 Hours validation is working properly: valid decimal values are accepted for both 2.5 and 2,5, while invalid inputs such as 'abc', '0', or '-1' trigger an error message and disable the Save button.
 - Double-submit on Save is prevented; the Save button disables on the first click, so a second submit can't fire. For Delete, the confirm dialog closes immediately; ConfirmDelete also guards against it in code, returning early when a delete is already in flight.
- The read-only mirror feature displays the same records on the volunteer profile without any edit, delete, or add buttons.

PR #713, opened 2026-06-21
required-field validation for accompanying opportunity creation

### LINKED ISSUE #698: fix: accompanying opportunity creation — required field validation and submission bugs
## Problem

When an agent creates a new opportunity with type **Accompanying**, the accompanying details section has no required field validation. The schema marks all fields as optional, so the form can be submitted with empty critical fields, sending null values to the BE.

## Affected file

`src/components/Dashboard/Profile/sections/AccompanyingDetails/accompanyingDetailsSchema.ts`

## Bugs to fix

1. **No required validation on key fields** — `appointmentAddress`, `appointmentPostcode`, `appointmentDate`, `appointmentTime`, `refugeeName` are all `.optional()` in the schema. For accompanying type these should be required (or at minimum address + date).
2. **`accompanyingDetailsSchema` is shared** between the creation form (`NewOpportunity`) and the profile edit section (`AccompanyingDetails`). The creation form should enforce stricter validation than the edit form (where partial saves are acceptable). Consider a separate schema or a stricter variant for creation.

## Expected behaviour

When opportunity type is **Accompanying** and the agent clicks Create, form submission should be blocked until the required accompanying fields are filled in, with inline error messages shown.

## Related

- PR https://github.com/need4deed-org/fe/pull/693 (new opportunity creation page)

ivannissimrch:
## Description
Adds required-field validation to the accompanying create form, so it can't be submitted with the important appointment fields left blank.

## Related Issues
Closes #698

## Changes
- Turned the accompanying schema into a function `createAccompanyingDetailsSchema(t, required)`  so the same schema can be strict or loose:
- Create form uses `required=true.` 
- Edit page uses `required=false.` 
- Required on create: appointment address, postcode, date, time, refugee name, and appointment language (each shows an inline error, en, de).
- The create button now validates the accompanying section before submitting

## Questions for review
1. **Clearing a field on edit doesn't stick.** If you clear the postcode, date,
     or time and save, it goes back to the old value. Changing it to a *new* value
     works fine, only clearing to empty reverts. (The empty value becomes `undefined`,
     which gets dropped from the PATCH request, so the backend never updates it.)
     Is this expected, or should I fix it as a separate bug?

 2. **Why appointment language is required.** This goes beyond the five fields of the
     ticket listed, but the reason: if no language is picked, the backend rejects the request, 
    Because the value has to be one of the allowed enum options. 

  3. **How strict should postcode/phone be?** I kept them as "just not empty," but
     I noticed `isValidPLZ` exists in `forms/utils.ts`  but isn't used anywhere,
     And `PHONE_NUMBER_REGEX` is used in the contact-detail schemas but not here.
     Keep them loose, or add format validation?

## Screenshots / Demos
Screenshot from 2026-06-21 07-59-36

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes


arturasmckwcz:
## Code Review

### Overview

Adds required-field validation to the accompanying opportunity creation form by refactoring `accompanyingDetailsSchema` into a factory function `createAccompanyingDetailsSchema(t, required)`. Edit form uses `required=false` (loose), create form uses `required=true` (strict). Clean, well-scoped approach.

---

### Correctness

**`??` → `||` change is correct:**
```ts
// Before
accomp_translation: isAccompanying ? (accompData?.appointmentLanguage ?? null) : null,
// After
accomp_translation: isAccompanying ? accompData?.appointmentLanguage || null : null,
```
`||` is intentionally stricter here — it prevents empty string `""` (a possible schema value) from reaching the API. Good catch.

**Date required validation is unusual:**
```ts
z.date().nullable().refine((v) => v !== null, { message: t(...) })
```
`.nullable()` followed by a refine rejecting `null` works, but `z.date({ required_error: t(...) })` without `.nullable()` is more idiomatic Zod.

**`appointmentLanguage` required chain is over-engineered:**
```ts
z.nativeEnum(TranslatedIntoType).or(z.literal("")).optional().refine((v) => !!v, {...})
```
The `.optional()` before `.refine()` means `undefined` passes to refine and gets rejected — which is the intent — but this could be simplified. Consider a comment explaining the `or(z.literal(""))` is for the unselected state.

---

### I18n regression (minor)

The old schema used message keys (`"invalidTimeFormat"`) that were translated at render time via `t(error.message)`. The new schema pre-translates at schema creation time. `useForm`'s `resolver` is not reactive — it only reads the options once. If a user switches language at runtime, the validation messages won't update until the component remounts. Probably acceptable for this app, but worth noting.

---

### Dead code

`ErrorText` was removed from the import in `AccompanyingDetailsEdit.tsx` but its definition is almost certainly still in `styles.ts` (not in the diff). Should be removed from styles to avoid confusion.

---

### Question answers

**Q1 (clearing fields reverts):** Out of scope here; the empty string → `undefined` → omitted from PATCH is a real bug but a separate one. Worth filing as a follow-up issue.

**Q2 (appointmentLanguage required):** Correct call. The backend enum constraint means `""` or missing would be a runtime error. Making it required at the form level is the right fix.

**Q3 (postcode/phone strictness):** `isValidPLZ` already exists in `forms/utils.ts` and is unused anywhere — this PR is the right place to wire it in for postcode. Phone is lower-priority. Recommend adding:
```ts
appointmentPostcode: required
  ? z.string().min(1, t(`${i18nPrefix}.postcodeRequired`)).refine(isValidPLZ, t(...))
  : z.string().optional()
```

---

### Suggestions

1. **Remove `ErrorText` from `styles.ts`** — it's now dead code.
2. **Use `z.date()` without `.nullable()`** for the required date field, or at minimum add a comment on the nullable+refine pattern.
3. **Wire in `isValidPLZ`** for the postcode field while the schema is being touched here.
4. **File a follow-up issue** for the clear-field-reverts-on-edit bug (Q1).

---

Solid, well-structured PR. The factory-function approach is the right abstraction. No blocking issues — the `isValidPLZ` wiring and dead-code cleanup are nice-to-haves but not merge blockers.

ivannissimrch:
Thanks for the thorough review. I added the changes:
 - Removed the dead ErrorText from styles.ts
 - Simplified the date field to z.date({ message }) and appointment-language to z.nativeEnum(E, { message })
 - Wired isValidPLZ into the postcode field,  reused the existing form.error.postcode message (same one BecomeVolunteer uses for this check, so no new key). Small note: isValidPLZ isn't actually unused,it's already used in BecomeVolunteer and AddOpportunity. Since it's now shared across forms and the dashboard, it could move to @/utils, but I kept that out of this PR to keep it scoped. Happy to do it as a follow-up or here if you want to.
- i18n: acknowledged, thank you.
- Filed the edit clear-field-reverts bug as a follow-up: #715
-Also guarded OpportunityCard's location?.[0] To stop a crash on the agent dashboard home. It hits when an agent that is active and has opportunities assigned, the home renders cards from /agent/:id/opportunity-linked, whose data has no location field, but the card reads location[0], and it shows Cannot read properties of undefined. An admin or an agent with no opportunities doesn't hit it. The guard prevents the crash.

src/components/Dashboard/NewOpportunity/NewOpportunity.tsx
```diff
@@ -612,7 +613,11 @@ export function NewOpportunity() {
           text={t("dashboard.newOpportunity.submit")}
           backgroundcolor="var(--color-aubergine)"
           textColor="var(--color-white)"
-          onClick={handleHeaderSubmit(onSubmit)}
+          onClick={async () => {
+            const accompValid = !isAccompanying || (await accompanyingMethods.trigger());
+            if (!accompValid) return;
```
arturasmckwcz: this early return reveals accompanying errors only masking other details errors if any.

ivannissimrch: I removed the early return and pulled the submit into a single handleCreate function that validates all three forms (header, details, accompanying) before submitting, so no section's errors get masked.

need4deed:
@ivannissimrch any idea why lint fails?

ivannissimrch:
@nadavosa  develop is broken, #735 merged with unresolved conflict markers in package.json (need4deed-sdk 0.0.115 vs ../sdk). It's invalid JSON,  Opening a hotfix to keep 0.0.115.  #742

nadavosa:
@ivannissimrch ok i didnt realize 
Thanks

ivannissimrch:
I merged the latest develop, lint is green now.

PR #714, opened 2026-06-22
fix: open avatar dropdown with logout on caret click

### LINKED ISSUE #661: fix: header avatar caret opens nothing (no dropdown menu)
### Summary
When logged in, the caret next to the user avatar (top-right header) toggles its
direction on click but never opens a menu.

### Steps to Reproduce
1. Log in.
2. Click the arrow next to the avatar.

### Expected Behavior
A menu opens (e.g. Profile, Logout). Note there is currently no logout path in
the UI at all.

### Actual Behavior
Only the arrow flips down→up; nothing else happens.

### Technical notes
`src/components/Header/UserProfile.tsx`: the `onClick` only flips `isCaretDown`.
There is no dropdown state, no menu, no `useClickOutside`.

Implement the dropdown to logout

Cy-fox:
Closes #661

When clicking the avatar caret in the dashboard header, only the icon flipped direction but no menu appeared. There was also no way to log out at all.

Changes:
- UserProfile.tsx: added dropdown with useRef and useClickOutside, opens on caret click and closes on outside click
- Logout item clears the is_logged_in cookie and redirects to /login
- Added dashboard.header.button.logout translation key to en and de locales

### DIFF
--- a/src/components/Header/UserProfile.tsx
+++ b/src/components/Header/UserProfile.tsx
@@ -1,8 +1,17 @@
+"use client";
+import { useClickOutside } from "@/hooks";
 import { useCurrentUser } from "@/hooks/useCurrentUser";
-import { CaretDownIcon, CaretUpIcon } from "@phosphor-icons/react";
-import { useState } from "react";
+import { CaretDownIcon, CaretUpIcon, SignOutIcon } from "@phosphor-icons/react";
+import { useRouter } from "next/navigation";
+import { useRef, useState } from "react";
+import { useTranslation } from "react-i18next";
 import styled from "styled-components";
 import { Paragraph } from "../styled/text";
+
+const UserProfileWrapper = styled.div`
+  position: relative;
+  display: inline-flex;
+`;
 
 const UserProfileContainer = styled.div`
   display: flex;
@@ -22,6 +31,35 @@ const UserIconDiv = styled.div`
   border-radius: var(--dashboard-cards-header-user-icon-border-radius);
 `;
 
+const DropdownMenu = styled.ul`
+  position: absolute;
+  top: calc(100% + 8px);
+  right: 0;
+  z-index: 100;
+  margin: 0;
+  padding: var(--spacing-8) 0;
+  list-style: none;
+  background: var(--color-white);
+  border-radius: var(--border-radius-xs);
+  box-shadow: var(--dropdown-box-shadow);
+  min-width: 140px;
+`;
+
+const DropdownItem = styled.li`
+  display: flex;
+  align-items: center;
+  gap: var(--spacing-8);
+  padding: var(--spacing-8) var(--spacing-16);
+  font-size: var(--font-size-lg);
+  color: var(--color-midnight);
+  cursor: pointer;
+  white-space: nowrap;
+
+  &:hover {
+    background-color: var(--color-grey-50);
+  }
+`;
+
 const getInitials = (fullName: string | undefined): string => {
   if (!fullName) return "";
   return fullName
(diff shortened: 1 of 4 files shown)


src/components/Header/UserProfile.tsx
```diff
@@ -34,23 +72,51 @@ const getInitials = (fullName: string | undefined): string => {
...
   const user = useCurrentUser();
+  const router = useRouter();
+  const { t, i18n } = useTranslation();
+
+  useClickOutside(ref, () => setIsOpen(false));
+
   const userName = getInitials(user?.fullName) || user?.firstName?.[0]?.toUpperCase() || "";
 
+  const handleLogout = () => {
+    document.cookie = "is_logged_in=; path=/; max-age=0; SameSite=Lax; Secure";
```
DarrellRoberts: You can use `clearAuthHint()` here from `@/utils` as it does the same thing

src/components/Header/UserProfile.tsx
```diff
@@ -34,23 +72,51 @@ const getInitials = (fullName: string | undefined): string => {
...
+    router.push(`/${i18n.language}/login`);
```
DarrellRoberts: Next router uses caching which can cause issues when there's been an auth change. I think this would also conflict with the `middleware`
`window.location.href` would be better here as it forces the browser to do a complete refresh

DarrellRoberts (changes requested):
we've needed this for a while and good work

Along with some comments below, you also need to remove the HTTP only cookies. You can do this on the `middleware` . Otherwise, although it may appear the user is logged out, the tokens are still there and you can still view the authorised content by just adding the `is_logged_in` cookie in the browser.

Can be a bit fiddly on the `middleware` but if you captured and stored both the access token and `is_logged_in` cookie, just could do something like this:

`/middleware.ts`
```typescript
if (cookieToken && !cookieAuth) {
  const isProd = process.env.NODE_ENV === "production";

  const baseOptions = {
    path: "/",
    secure: true,
    sameSite: "lax" as const,
    domain: isProd ? "app.need4deed.org" : undefined,
  };

  response.cookies.delete({ name: "access", ...baseOptions });
  response.cookies.delete({ name: "refresh", ...baseOptions });
}
```

That way, once the user is logged out, the only way to capture the access token again is to create another instance aka login.

What do you think?

src/components/Header/UserProfile.tsx
Cy-fox: Thanks, didn't know that existed. Will use it.

src/components/Header/UserProfile.tsx
Cy-fox: I agreed, but `window.location.href` is better here to force a full reload and let the middleware run fresh.

Cy-fox:
> we've needed this for a while and good work
> 
> Along with some comments below, you also need to remove the HTTP only cookies. You can do this on the `middleware` . Otherwise, although it may appear the user is logged out, the tokens are still there and you can still view the authorised content by just adding the `is_logged_in` cookie in the browser.
> 
> Can be a bit fiddly on the `middleware` but if you captured and stored both the access token and `is_logged_in` cookie, just could do something like this:
> 
> `/middleware.ts`
> 
> ```ts
> if (cookieToken && !cookieAuth) {
>   const isProd = process.env.NODE_ENV === "production";
> 
>   const baseOptions = {
>     path: "/",
>     secure: true,
>     sameSite: "lax" as const,
>     domain: isProd ? "app.need4deed.org" : undefined,
>   };
> 
>   response.cookies.delete({ name: "access", ...baseOptions });
>   response.cookies.delete({ name: "refresh", ...baseOptions });
> }
> ```
> 
> That way, once the user is logged out, the only way to capture the access token again is to create another instance aka login.
> 
> What do you think?



Updated the middleware to clear the access and refresh cookies when the `is_logged_in` hint isn't present but the access token still exists. Also followed the same options pattern you suggested.

PR #716, opened 2026-06-22
710: bug, fix filter chips so that when removed the url queries update

### LINKED ISSUE #710: bug: Filters aren't removed when closed
### Summary

On List view when a filter is selected, a small tag is created below search bar. When this tag is closed (clicking the X), it disappears but the query parameters are still there

### Steps to Reproduce

1. Go on lists view for either `opportunity` , `volunteer` or `agent`
2. Select a filter or filters from the Filter Accordion
3. Click the X mark on one of the filters
4. The query parameters are still in the url

### Expected Behavior

When a filter is closed , it should also remove it from the query parameters

### Actual Behavior

The url  remain unchanged.

### Environment

_No response_

DarrellRoberts:
## Description
Fixes bug whereby when the user clicks the cross on the `FilterChip` , the URL queries don't change. This is on either `volunteer` , `opportunity` or `agent` dashboard list views

## Related Issues
Closes #710 

## Changes
- Add a `getClearSingleFilter` to CardsFilter helpers. 
- Also identified duplicate key `Tandem` with `type` and `services` on NGO directory , so added a dfiferentiator

## Screenshots / Demos
Aufzeichnung 2026-06-23 010057

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes


src/components/Dashboard/Agents/Filters/helpers.ts
```diff
@@ -16,9 +16,15 @@ export const createAgentFilterItems = (
...
-  );
+  const typeFilters = generateNestedFilterControlItems(filter.type, setFilter, "type", (key, parent) => {
+    const fallbackPath = `dashboard.agents.filters.type.${key.toLocaleLowerCase()}`;
+
+    if (key === "tandem" && parent === "type") {
+      return `${t("dashboard.agents.table.type")} ${t(fallbackPath)}`;
+    }
+
+    return t(fallbackPath);
+  });
```
DarrellRoberts, line 28: In NGO directory there is a `type` Tandem filter and `services` Tandem filter, so modified this to prevent duplicate keys

arturasmckwcz:
## Review

This PR fixes a real bug: clicking the ✕ chip in `CardsHeader` was calling `f.onChange` (which updated component state) but never pushed the serialized filter to the URL, causing a desync between visible filters and URL on chip removal. The fix adds a `handleClearFilter` handler to all three dashboard pages and a new `getClearSingleFilter` helper, wiring them through a new `onClearFilter` prop on `CardsHeader`. The approach is sound; two issues need attention.

---

### `CardsHeader.tsx:116–122` — `f.onChange(!f.checked)` should be removed

The new click handler calls both `f.onChange(!f.checked)` and `onClearFilter(...)` in sequence. `onClearFilter` already handles state via `setCardsFilter(cleared)` and pushes the URL. `f.onChange` queues a functional state update to the same setter — React 18 batches both and the direct value update from `onClearFilter` wins, making `f.onChange`'s result discarded.

The one nuance: if `getClearSingleFilter` silently fails to find the key (e.g. `keyValue` is missing and `|| f.label` falls back to a translated string that doesn't match any state key), `f.onChange` becomes the only thing that actually clears the checkbox — an accidental safety net rather than intentional design. Either way, keeping both is fragile: `onChange` is the checkbox-toggle handler; firing it from the chip-removal path means any future side effects added to `onChange` (analytics, sub-state, animations) will silently fire on chip removal too.

**Fix:** remove `f.onChange(!f.checked)` from the XIcon click handler, and ensure `getClearSingleFilter` surfaces failures rather than returning unchanged state silently.

---

### `Volunteers/Filters/helpers.ts:63` / `Opportunities/Filters/helpers.ts:71` — availability items get `keyValue` but no `parentKey`

For `generateNestedFilterControlItems` (district, language, type, engagement) the PR correctly sets both `keyValue: k` and `parentKey: key as string`. For `createAvailabilityFilterItems`, only `keyValue: key` is added; `parentKey` is not set and remains `undefined`.

When the ✕ is clicked on an availability chip (e.g. "morning"), `onClearFilter("morning", undefined)` is called, passing no scope constraint into `getClearSingleFilter`. The function then recursively searches the entire filter tree for any key named `"morning"` and clears every match it finds. In current data this works because availability slot names ("monday", "morning", "weekly") don't collide with district or language keys — but there's no guard against future API data introducing such a collision, which would silently clear two unrelated filters on one click.

**Fix:** add `parentKey: QueryParamsKeys.AVAILABILITY` to the items inside `createAvailabilityGroup`, scoping the search to the availability subtree.

DarrellRoberts:
thanks for reviewing @arturasmckwcz ! I've committed the changes

need4deed:
@DarrellRoberts lint fails

DarrellRoberts:
@need4deed this lint fail is on `develop`

@polomontague can you please resolve/remove the Merge conflict from `package.json` ?
image

It's also on `yarn.lock` if you look at commit https://github.com/need4deed-org/fe/commit/927078741e33ce7920b40c68799d5cf8524f3c23

DarrellRoberts:
@need4deed it's no longer on `develop` so the linting has passed. Thanks @ivannissimrch !

polomontague:
I’ll work on it this evening (my time)

PR #743, opened 2026-07-01
449: adds activity filters to Volunteer Dashboard

### LINKED ISSUE #449: feat: filter and sort volunteers by activity type
## What

The volunteers search/list page has no way to filter or sort by activity type. Add activity type as a filter option so coordinators can quickly find volunteers interested in a specific type.

## Where

`/dashboard/volunteers` — the volunteers list/search page.

## Tasks
- [ ] FE: add activity type filter UI to the opportunities list (checkbox or dropdown, following the same pattern as existing filters)
- [ ] FE: wire the filter to the existing opportunities query/API call
- [ ] BE: confirm `GET /opportunity` supports an `activityType` or `activity` query param filter; add it if missing
- [ ] Sort by activity type (secondary — can be done as a follow-up)

## Related
- BE: https://github.com/need4deed-org/be/issues/234 (opportunity endpoints)

DarrellRoberts:
## Description
Adds `activity` filter to `volunteer` dashboard

## Related Issues
Closes #449 

## Changes
- Adds `activity` to Volunteer Filter Content
- Adds handlers for URL query to add and remove activity query
- Renamed type `CardFilters` to `VolunteerCardFilters` for better semantic naming

## Screenshots / Demos
image

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes


nadavosa:
## Review — 449: adds activity filters to Volunteer Dashboard

The PR correctly wires the activity filter end-to-end: new interface field, default state, `createFilterItems` generation, `AccordionFilter` render, chip display, URL serialization, and URL deserialization. Two bugs were found in the new filter code by comparing it to the established district/language patterns in the same file.

---

### Bug 1 — Deserializer missing `!== undefined` guard

`src/components/Dashboard/Volunteers/helpers.ts`

```ts
// PR (activity) — no guard
queryActivities.forEach((l) => {
  newFilter.activity[l] = true;  // ← any URL param writes here unconditionally
});
```

Every other filter in `deserializeVolunteerFilters` guards before writing:
```ts
// district / language / type / engagement — all have this
if (newFilter.district[d] !== undefined) {
  newFilter.district[d] = true;
}
```

Without the guard, a stale or hand-crafted URL param (e.g. `?activity=OldName`) injects a phantom key into `newFilter.activity`. That key has no matching checkbox in the UI, but it does get picked up by `serializeFilters` on the next render and sent to the API — the backend receives a filter value that doesn't exist in the current data model.

**Fix:**
```ts
queryActivities.forEach((l) => {
  if (newFilter.activity[l] !== undefined) {
    newFilter.activity[l] = true;
  }
});
```

---

### Bug 2 — Serializer uses `|| key` fallback instead of explicit `if (id !== undefined)` guard

`src/components/Dashboard/Volunteers/helpers.ts`

```ts
// PR (activity)
const paramValue =
  (options?.serializeToIDs && options.apiFilterOptions?.activity?.find((d) => d.title === key)?.id) || key;
params.append(EntityTableName.ACTIVITY, String(paramValue));
```

The district and language blocks handle the `serializeToIDs` path differently — they skip appending entirely when the lookup fails:
```ts
// district / language
if (options?.serializeToIDs && options.apiFilterOptions) {
  const districtId = options.apiFilterOptions.district?.find((d) => d.title === key)?.id;
  if (districtId !== undefined) {
    params.append(QueryParamsKeys.DISTRICT, String(districtId));
  }
  // else: nothing appended
} else {
  params.append(QueryParamsKeys.DISTRICT, key);
}
```

With the `|| key` fallback, if `serializeToIDs=true` and the activity title isn't found in `apiFilterOptions.activity`, the title string is sent to the API instead of being silently dropped. `VolunteerListController` calls `serializeFilters` with `serializeToIDs: true` for the API request — so if the options haven't loaded yet or the activity is missing from the list, a title string goes to the backend where an ID is expected.

**Fix:** align with the district/language two-branch pattern instead of `|| key`.

---

Everything else looks good — `EntityTableName.ACTIVITY = "activity"` matches the expected URL param key, `CardFilterKeys → VolunteerCardFilterKeys` is a safe rename with no external consumers, and the `(key) => key` label function is consistent with how district and language filters work.

DarrellRoberts:
thanks @nadavosa for reviewing. I've pushed the changes

PR #752, opened 2026-07-02
fix(750): correct consent link URLs in volunteer form

nadavosa:
## Summary

- Fixes `/{path}/{lang}` → `/{lang}/{path}` URL order in the volunteer form consent links
- Replaces `Subpage.DATA_PROTECTION` (`"data-protection"`) with `Subpage.DATA_PRIVACY` (`"data-privacy"`) so the data privacy link resolves to the existing `/[lang]/data-privacy` page

## Before / After

| | Before | After |
|---|---|---|
| Data privacy link | `/data-protection/de` ❌ (404) | `/de/data-privacy` ✅ |
| Agreement link | `/agreement/de` ❌ (404) | `/de/agreement` ✅ |

## Related

Closes https://github.com/need4deed-org/fe/issues/750

Same fix was applied to the opportunity form in PR https://github.com/need4deed-org/fe/pull/706 — the volunteer form was missed.


ivannissimrch (commented):
I believe this still results in a 404 error. The route segment is [lang], which means that `useParams()` returns `{ lang }`, not `{ lng }`. When I logged it, I got `{ lang: 'de' }`. Using `const { lng } = useParams()` returns undefined, resulting in the render path being `/undefined/data-privacy`.

Question is: Since the middleware already injects the locale for any path that doesn't include a language, should the consent link keep the language in its href, or should we drop it and let the middleware inject the locale?
`<a href={`/${Subpage.DATA_PRIVACY}`}>{t("homepage.footer.legal.dataPrivacy")}</a>{" "}`

ivannissimrch (approved):
I  tested locally on /en and /de; now Data Privacy and Agreement links resolve correctly

Issue #440, opened 2026-05-04
Feature: add tagging ability in comment section


DarrellRoberts:
### Description

In the comment section add a feature that allows the user to tag other users. For example with a `@` character similar to tagging in the Slack platform. 

When the `@` character is present, an autocomplete dropdown will appear whereby the user selects which user they want to tag (again similar to Slack).

When the comment is saved, it will send data to the `be` which will then trigger the Slack Web hook to process it and create a Slack notification, detailing:

- User(s) that has been tagged
- Which user sent the tag
- Path of where it happened
- Mention text content

e.g. "John tagged Sarah in `/opportunity/5` : 'hey `@Sarah`, what do you think of this?'"

### Proposed solution

This will be a big task so I propose breaking it down into these sections

1. Modify `TextArea` component on `Comments` so that when a user is tagged with the @ key, the tag is a different color than the rest of the text (usually blue). There are a few approaches to this but some involve the inject HTML method which we should avoid for XSS attack reasons. I'm drawn towards a CSS Mirroring technique which avoids this. 
May also be better to use a library such as [react-mentions-ts](https://github.com/hbmartin/react-mentions-ts) but I'm hoping this won't be neccessary.
2. Create autocomplete component that is triggered by the @ character. This will work in the same way as the searchbar on the `/dashboard/volunteer` route with a debounced request and can use a Selector component to contain the results (I imagined you might have to create your own rather than using our current one).
3. Modify the comment mutation hook and form that will send the additional request data.  We can use the current comment request and add a nested object like below:
```typescript
{
// existing comment data,
taggedInfo: {
	authorId: number // e.g. 1
	taggedUserIds: Array<number> // e.g. [2, 3], so we can tagged multiple users
	path: string // e.g. "/dashboard/opportunity/1"
	mentionText: string //"hey @user2, @user3, can you help out"
}
}
```

need4deed:
@DarrellRoberts btw tagging in production does not allow to select any of the need4deed coordinators, not sure why

DarrellRoberts:
thanks for flagging @need4deed . I saw in `dev` the users from the `autocomplete` were no longer filtered to `coordinators` , I fixed this in PR #754 . Not sure if this is the issue though in `production` , just a guess as I can only login as Agent in `prod`

PR #754, opened 2026-07-02
Fixes coordinator user filter in tag autocomplete

DarrellRoberts:
## Description
Fetched users from autocomplete was not filtered to coordinators only. Possible the filter changed in `be` . This corrects the filter so that only coordinators are shown.

## Related Issues
Currently no issue created but mentioned on #440 

## Changes
- Changes filter in useGetQuery from `filter: {role: UserRole.COORDINATOR}` to `role: UserRole.COORDINATOR`

## Screenshots / Demos

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes


src/hooks/useGetQuery.tsx
```diff
@@ -44,6 +44,7 @@ interface Params {
   limit?: number;
   sortOrder?: SortOrder;
   filter?: FilterParam;
+  role?: UserRole;
```
nadavosa, line 47: This bolts a one-off `role?: UserRole` field onto the shared, resource-agnostic `Params` interface used by ~30 `useGetQuery` call sites app-wide. The hook already has a generic mechanism for exactly this (`filter?: FilterParam`, a `Record<string, unknown>`), used elsewhere for arbitrary resource filters, e.g. `filter: { status: OpportunityStatusType.NEW }` in `NewestVolunteers.tsx`/`NewestOpportunities.tsx`/`OpportunityListController.tsx`/`AgentListController.tsx`.

`role` is only ever set here, at this one call site. Since `be`'s `userListQuerySchema` has `additionalProperties: false` and expects `role` as a flat param rather than nested, could this instead flow through the generic `filter` path if `getReducedFilter` flattened it accordingly (e.g. treating `filter` entries as top-level query keys), rather than adding a domain-specific field to the shared interface? Otherwise this sets a precedent: the next resource that needs a one-off top-level param gets its own bespoke field added here too, and `Params` keeps growing with fields most callers never use.

nadavosa (commented):
Confirmed the fix is correct — checked the `be` repo directly: `userListQuerySchema` in `src/server/schema/querystring.ts` defines `role` as a top-level query param (not nested under `filter`, and `additionalProperties: false` means the old nested `filter: { role }` shape was silently dropped). So this isn't a guess, it matches the actual backend contract.

One cleanup note inline below.

src/hooks/useGetQuery.tsx
DarrellRoberts: thanks @nadavosa for reviewing! Tbh I'm not sure if there's a clean solution to this cleanup problem you highlighted considering in the `be` it doesn't accept `filter: {role }` anymore.  

The main blocker here is that if we try to handle this inside the hook by flattening the filter object, it actually breaks the other 30 call sites. For example, if a query passes `{ filter: { status: 'new', role: 'coordinator' } }`, we would have to write complex logic inside the generic hook to pluck `role` out to the top level while leaving `status` nested inside the object so the `be` views don't break.

My reasoning for adding it to the shared interface was that `UserRole` is a core concept that we will probably need in the future. Plus, since it’s optional, it won't affect any existing callers if left undefined.

That said, if you have something in mind, I'd love to hear it

DarrellRoberts: alternatively, the other solution would be to change it back to what it was in the `be` where `role` is nested within `filter` . Either way, the `fe` and `be` should be aligned

PR #756, opened 2026-07-02
fix: persist Opportunities sort order in URL (#729)

### LINKED ISSUE #729: fix: Opportunities table sort order resets when navigating back from a profile
## Context

NGO feedback: sorting the Opportunities table by "Event: Nächste zuerst" (soonest first), then opening an opportunity and navigating back, loses the chosen sort — it resets to the default.

## Root cause

`src/components/Dashboard/Opportunities/Opportunities.tsx`:
```ts
const [sortOrder, setSortOrder] = useState<string>(SortOrder.NewToOld);
```
`sortOrder` is plain component state, never written to the URL. By contrast, `cardsFilter` *is* synced to the URL via `serializeOpportunityFilters` + `router.push` — this was the fix for https://github.com/need4deed-org/fe/issues/356 (filters resetting on navigation), but that fix never extended to sort order, which uses a separate mechanism.

## Fix

Persist `sortOrder` the same way filters already are — read the initial value from a URL search param (e.g. `?sort=appointment-proximal`) and include it when pushing filter updates to the URL, so navigating to an opportunity profile and back restores the previously selected sort.

## Acceptance criteria

- [ ] Selecting a sort order, opening an opportunity profile, and navigating back keeps the same sort order applied
- [ ] Sort order is reflected in the URL (shareable/bookmarkable, consistent with existing filter behavior)

thnndnly:
## What & why

Fixes #729. Sorting the Opportunities table (e.g. "Event: Nächste zuerst"),
opening an opportunity, and navigating back lost the chosen sort — it reset
to the default because `sortOrder` was plain component state, never written
to the URL. Filters were already synced to the URL (the #356 fix); this
extends the same mechanism to sort order.

## Changes

- Read the initial sort from the `sort` query param and validate it against
  an allow-list (`parseSortParam`), falling back to the default.
- Write/remove the `sort` param on sort change; reset pagination (`page`)
  the same way filter changes already do.
- Reset sort to default on "Clear all filters".
- Keep sort untouched on filter updates and single-filter clears (it rides
  along in the URL).
- Refactor-only: move the appointment-sort helpers
  (`APPOINTMENT_SORT_VALUES` / `isAppointmentSort`) into `helpers.ts` to
  remove duplication, and add `serializeOpportunityFilters` overloads so the
  `as URLSearchParams` casts can be dropped.

## Notes

- The appointment sorts are client-side only (backend has no such
  `SortOrder`), so they currently sort within the loaded page only — a
  pre-existing limitation, out of scope here. Candidate follow-up: add
  first-class appointment sorting in backend + SDK, then drop the FE
  client-sort and remap.
- `Volunteers.tsx` has the same "sort not persisted" pattern — separate
  follow-up, not addressed here.

### DIFF
--- a/src/components/Dashboard/Opportunities/Opportunities.tsx
+++ b/src/components/Dashboard/Opportunities/Opportunities.tsx
@@ -36,7 +36,7 @@ export function Opportunities() {
   const searchParams = useSearchParams();
   const pathname = usePathname();
   const router = useRouter();
-  const [sortOrder, setSortOrder] = useState<string>(() => parseSortParam(searchParams.get(SORT_PARAM)));
+  const sortOrder = parseSortParam(searchParams.get(SORT_PARAM));
   const tabs = !user
     ? []
     : isAgent
@@ -61,7 +61,6 @@ export function Opportunities() {
 
   const handleSortChange = (order: string) => {
     const sort = parseSortParam(order);
-    setSortOrder(sort);
     const params = new URLSearchParams(searchParams);
     params.delete("page");
     if (sort === DEFAULT_SORT_ORDER) params.delete(SORT_PARAM);
@@ -106,17 +105,10 @@ export function Opportunities() {
   const handleClearAllFilters = () => {
     const cleared = getClearFilter(cardsFilter);
     setCardsFilter(cleared);
-    setSortOrder(DEFAULT_SORT_ORDER);
     const params = serializeOpportunityFilters(cleared, searchParams, false);
     params.delete(SORT_PARAM);
     router.push(pathname + questionMark + params.toString());
   };
-
-  // sync sort on back/forward within the same route (no remount → lazy init won't re-run)
-  const sortParam = searchParams.get(SORT_PARAM);
-  useEffect(() => {
-    setSortOrder(parseSortParam(sortParam));
-  }, [sortParam]);
 
   useEffect(() => {
     if (!apiFilterOptions) return;
--- a/src/components/Dashboard/Opportunities/Opportunities.tsx
+++ b/src/components/Dashboard/Opportunities/Opportunities.tsx
@@ -4,7 +4,7 @@ import { useTranslation } from "react-i18next";
 import { DashboardLayout } from "@/components/Layout";
 import { apiPathOption, questionMark } from "@/config/constants";
 import { useGetVolunteer, useGetQuery } from "@/hooks";
-import { ApiOptionLists, EntityTableName, SortOrder, UserRole } from "need4deed-sdk";
+import { ApiOptionLists, EntityTableName, UserRole } from "need4deed-sdk";
 import { usePathname, useRouter, useSearchParams } from "next/navigation";
 import Filters from "../common/CardsFilter/Filters";
 import CardsHeader from "../common/CardsHeader/CardsHeader";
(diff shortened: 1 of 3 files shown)


src/components/Dashboard/Opportunities/Opportunities.tsx
```diff
@@ -25,12 +31,12 @@ export function Opportunities() {
   const { t } = useTranslation();
   const [isFiltersOpen, setIsFiltersOpen] = useState(false);
   const [numOfOpps, setNumOfOpps] = useState(0);
-  const [sortOrder, setSortOrder] = useState<string>(SortOrder.NewToOld);
   const [cardsFilter, setCardsFilter] = useState(defaultOpportunityCardsFilter);
   const { data: apiFilterOptions } = useGetQuery<ApiOptionLists>({ queryKey: ["options"], apiPath: apiPathOption });
   const searchParams = useSearchParams();
   const pathname = usePathname();
   const router = useRouter();
+  const [sortOrder, setSortOrder] = useState<string>(() => parseSortParam(searchParams.get(SORT_PARAM)));
```
DarrellRoberts: what do you think about removing this `useState` and just maintaining the `sortOrder` state as a URL query?

e.g. you could remove this along with the `setSortOrder` and `useEffect` and just have:
```typescript
const sortOrder = parseSortParam(searchParams.get(SORT_PARAM));
```

as you're already setting the `sortOrder` in the `params`

src/components/Dashboard/Opportunities/Opportunities.tsx
```diff
@@ -94,9 +106,18 @@ export function Opportunities() {
...
+    params.delete(SORT_PARAM);
+    router.push(pathname + questionMark + params.toString());
   };
 
+  // sync sort on back/forward within the same route (no remount → lazy init won't re-run)
+  const sortParam = searchParams.get(SORT_PARAM);
+  useEffect(() => {
+    setSortOrder(parseSortParam(sortParam));
+  }, [sortParam]);
+
```
DarrellRoberts: regarding my earlier comment, you could then also remove this `useEffect`

DarrellRoberts (commented):
looks good and works well!

I just wondered what you think about removing the `sortOrder useState` entirely

thnndnly:
Good call, done. Removed the `sortOrder` state entirely and derive it straight from the `sort` search param. That also let the back/forward sync effect go, since `useSearchParams` already re-renders on URL changes. Ready for another look.

src/components/Dashboard/Opportunities/helpers.ts
```diff
@@ -22,12 +23,40 @@ interface SerializeFiltersOptions {
...
+  searchParams?: ReadonlyURLSearchParams,
+  asString?: true,
+  options?: SerializeFiltersOptions,
+): string;
+export function serializeOpportunityFilters(
+  filter: OpportunityCardsFilter,
+  searchParams: ReadonlyURLSearchParams | undefined,
+  asString: false,
+  options?: SerializeFiltersOptions,
+): URLSearchParams;
```
DarrellRoberts, line 53: personally I've never been a fan of `overloads` but I see why you did it here and it makes it more type-safe.

Alternatively you could use a `generic` like so:

```typescript
export function serializeOpportunityFilters<T extends boolean = true>(
  filter: OpportunityCardsFilter,
  searchParams?: ReadonlyURLSearchParams,
  asString: T = true as T,
  options?: SerializeFiltersOptions,
): T extends false ? URLSearchParams : string {
...
return (asString ? params.toString() : params) as T extends false ? URLSearchParams : string;
```
but that's also not particularly pretty. Up to you

DarrellRoberts (approved):
lgtm 👍 Just had one comment concerning ts but up to you

PR #760, opened 2026-07-03
User menu logout

### LINKED ISSUE #661: fix: header avatar caret opens nothing (no dropdown menu)

kiborrdis:
## Description
Added user menu on user avatar click and logout action

## Related Issues
Closes #661

## Changes
- Added general purpose component for menu. Loosely based on TrustLevelDropdown. After review if everyone okay with it I will replace TrustLevelDropdown implementation with that shared one
_But in general I would advise to discuss installing some ui lib, as there is none currently. Implementing this stuff by hand is waste of time, and it wont be as thought out and accessible as lib implementation_
- Implemented user profile dropdown based on figma design.
_Note that I didnt use background color for user dropdown from figma, as i didnt find any instances of this color in the project. Tell me if you want to use the color form figma layout(Below I attached screenshot of the figma layout for this block for your reference)
And box-shadow is slightly different color, again, i've used the one from code, rather than one from layout. Tell me if you want it changed_
image

- Added useLogout hook, the same way current mutation requests are implemented, wired it to the logout button

## Screenshots / Demos

image

## Checklist
- [ ] Check user menu is opening. Contains name of logged in user and their email
- [ ] Logout button is working. Logout request is sent when button is clicked. Access/refresh tokens removed from cookies. User redirected to /login page

### DIFF
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -489,7 +489,8 @@
       "button": {
         "login": "Anmelden",
         "joinVolunteer": "Finde jetzt dein Ehrenamt",
-        "dashboard": "Dashboard"
+        "dashboard": "Dashboard",
+        "logout": "Abmelden"
       }
     },
     "login": {
--- a/public/locales/en/translations.json
+++ b/public/locales/en/translations.json
@@ -483,7 +483,8 @@
       "button": {
         "login": "Log in",
         "joinVolunteer": "Join as a volunteer",
-        "dashboard": "Dashboard"
+        "dashboard": "Dashboard",
+        "logout": "Log out"
       }
     },
     "login": {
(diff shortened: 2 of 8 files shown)


DarrellRoberts (approved):
lgtm and works great!

I get your point on UI libraries.

We used to have a team of designers who would create specific wireframes, which is why we use our own custom UI. That and it kept the bundle-size lean as well as reduce reliance on third-party design systems. 

But ye, having a UI library would make these tasks as well as accessibility a lot easier, so definitely something to discuss in the standup

PR #771, opened 2026-07-07
562: reset password feature

### LINKED ISSUE #562: feat: add password reset flow (forgot password page + email link)
## What

Add a password reset flow so users can recover their account when they forget their password.

## Flow

1. **Forgot password link** on the login page → user enters their email
2. **API call**: `POST /auth/password-reset` (see https://github.com/need4deed-org/be/issues/582 for BE implementation)
3. **Email sent** with a reset link containing a short-lived token
4. **Reset page**: user clicks the link, lands on `/reset-password?token=...`, enters a new password
5. **API call**: `POST /auth/password-reset/confirm` with token + new password

## FE tasks
- [ ] Add "Forgot password?" link on the login page
- [ ] Build `/forgot-password` page — email input + submit
- [ ] Build `/reset-password` page — new password + confirm fields (reads token from query param)
- [ ] Wire up to BE endpoints
- [ ] Handle errors: invalid token, expired token, email not found
- [ ] Translations: en + de

## Depends on
- https://github.com/need4deed-org/be/issues/582 (BE password reset endpoint)

DarrellRoberts:
## Description
Adds a feature to reset password which follows these steps:

- User clicks "forgot password" on `/login`
- User taken to `/forgotten-password` where they are prompted to enter their email
- User inputs email and clicks submit which sends an email to their address with following path: `/reset-password?token=[token]`
- User navigates to this link and inputs their new password and confirms it.
- After submit, it changes the password and takes user back to `/login`

## Related Issues
Closes #562

#### How to manually test

1. Start `be` local dev
2. Create a user
3. On `fe` go to `/login` and click "forgot password?". Ensure you're sending data to `be` localhost
4. Enter email of the user you created and submit
5. In the terminal logs of `be` you'll see the email link
6. Copy the path of the email link e.g. `/reset-password?token=[token]` and paste it into the `fe` so that you have: `localhost:3000/en/reset-password?token=[token]`
7. Enter new password and submit
8. On success you'll be taken to `/login` where you can test your new password which should work

#### /forgotten-password
image

#### On success
image

#### /reset-password?token=[token]
image

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes


src/components/Login/Login/LoginForm.tsx
```diff
@@ -28,31 +29,14 @@ const useLoginMutation = (onLoginSuccess: () => void) => {
...
-  justify-content: space-between;
-  align-items: center;
-`;
-
-const LoginButtonDiv = styled.div`
-  display: flex;
-  justify-content: center;
-`;
-
 interface LoginFormProps {
```
DarrellRoberts, line 32: refactored LoginForm to put styles in one shared file

src/components/Login/Login/Login.tsx
```diff
@@ -1,69 +1,13 @@
...
-
-        <LoginSubContainer>
-          <ImageWithGradient imageUrl={imageUrl} gradientClass={gradientClassName} />
-        </LoginSubContainer>
-      </LoginContainer>
-    </PageLayout>
+    <LoginLayout heading={t("dashboard.login.login")}>
+      <LoginController />
+    </LoginLayout>
   );
```
DarrellRoberts, line 12: refactored Login to use a shared layout as I use this also for `/forgotten-password` and `/reset-password` pages

src/components/Login/ResetPassword/ResetPasswordForm.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { Button } from "../../core/button";
+import { useMutationQuery } from "@/hooks";
+import { useSearchParams } from "next/navigation";
+import { apiPathPasswordReset } from "@/config/constants";
+import { createResetPasswordSchema } from "./resetPasswordSchema";
+import { StyledForm } from "../styles";
+
+interface ResetPasswordData {
+  newPassword: string;
+  confirmPassword: string;
```
DarrellRoberts, line 14: `confirmPassword` not needed in the `be` (it does nothing). Still wanted another password value so that the user could essentially double-check what they're changing their password to. 
This is so in the form it triggers a `fe` error if `newPassword` and `confirmPassword` do not match

src/components/Login/ForgotPassword/ForgotPasswordForm.tsx
```diff
@@ -0,0 +1,90 @@
...
+        form.handleSubmit();
+      }}
+    >
+      <form.Field
+        name="email"
+        validators={{
+          onChange: ({ value }) => (!value ? t("dashboard.login.emailMissing") : undefined),
+          onChangeAsyncDebounceMs: 500,
+          onChangeAsync: async ({ value }) => {
+            await new Promise((resolve) => setTimeout(resolve, 500));
```
kiborrdis: Why do we need this onChangeAsync with synthetic wait?
IT feels like it's easier to add to onChange, or just use some email validator

src/components/Login/ResetPassword/ResetPasswordForm.tsx
```diff
@@ -0,0 +1,131 @@
...
+      newPassword: "",
+      confirmPassword: "",
+      token: urlTokenParam || "",
+    },
+    onSubmit: async ({ value }) => {
+      resetPassword(value);
+    },
+  });
+  const schema = createResetPasswordSchema(t);
+  console.log("s", schema);
```
kiborrdis: Probably better to remove

src/components/Login/ResetPassword/ResetPasswordForm.tsx
```diff
@@ -0,0 +1,131 @@
...
+            placeHolder={t("dashboard.login.newPassword")}
+            errors={field.state.meta.errors}
+          />
+        )}
+      </form.Field>
+      <form.Field
+        name="confirmPassword"
+        validators={{
+          onChange: ({ value, fieldApi }) => {
+            const resultValue = schema.shape.newPassword.safeParse(value);
```
kiborrdis: About all validation in this file:

I think it's better to try form level validation
https://tanstack.com/form/v1/docs/framework/react/guides/validation
As far as i understand it's the docs for a newer version, but still worth a try to do it like that.

Otherwise I would propose to wrap it in a function. It will remove code duplication with other validators, and will make reading this file easier. For confrimPassword matching with newPassword, I feel, if form level validation wont work, it's easier and more readable to make a check in place

src/components/Login/ResetPassword/resetPasswordSchema.ts
```diff
@@ -0,0 +1,19 @@
...
+    .object({
+      newPassword: z
+        .string()
+        .min(1, { message: t("dashboard.login.passwordMissing") })
+        .min(8, { message: t("dashboard.login.passwordTooShort") })
+        .max(50, { message: t("dashboard.login.passwordTooLong") }),
+      confirmPassword: z.string().min(1, { message: t("dashboard.login.passwordMissing") }),
+    })
+    .refine((data) => data.newPassword === data.confirmPassword, {
+      message: "dashboard.login.passwordNotMatching",
```
kiborrdis: Why not t("dashboard.login.passwordNotMatching") here? You would avoid all this `startsWith('.dashboard')` unpleasantness from component

src/components/Login/ResetPassword/ResetPasswordForm.tsx
```diff
@@ -0,0 +1,131 @@
...
+      <form.Subscribe selector={(state) => state}>
+        {() => (
+          <Button
+            type="submit"
+            text={t("dashboard.login.resetPassword")}
+            backgroundcolor={form.state.canSubmit && !isPending ? "var(--color-aubergine)" : "var(--color-grey-50)"}
+            textColor={form.state.canSubmit && !isPending ? "var(--color-white)" : "var(--color-grey-400)"}
+            textHoverColor="var(--color-magnolia)"
+            disabled={
+              !form.state.canSubmit || !form.state.values.newPassword || !form.state.values.confirmPassword || isPending
```
kiborrdis: I'm almost sure, you can use form.state.isValid instead newPassword and confirmPassword checks. Maybe even omit isValid, since there is a chance it's included in canSubmit

src/components/Login/ForgotPassword/ForgotPasswordForm.tsx
DarrellRoberts: good point, I just copied the email field from the `Login` without questioning this so have tidied it up

src/components/Login/ResetPassword/ResetPasswordForm.tsx
DarrellRoberts: I see what you mean and good suggestion. I find it easier to stick to the zod Schema pattern particularly as we're using it for other validators for user-defined inputs (e.g. on the Profile edits).

I've wrapped the validation logic in one function and it does look a lot cleaner

src/components/Login/ResetPassword/resetPasswordSchema.ts
DarrellRoberts: ye good point, have change this and removed the `startsWith('.dashboard')` from the component

src/components/Login/ResetPassword/ResetPasswordForm.tsx
DarrellRoberts: very true, it would seem that this is indeed covered by the `canSubmit` so have simplified it to that

DarrellRoberts:
thanks @kiborrdis ! Have implemented the changes

src/components/Login/ResetPassword/ResetPasswordForm.tsx
kiborrdis: I meant using form level validation with zod. Based on [that](https://tanstack.com/form/v1/docs/framework/react/guides/validation#setting-field-level-errors-from-the-form-s-validators) I believe, it's probably possible to do something like that


```ts
const form = useForm({
  defaultValues: {
    newPassword: "",
    confirmPassword: "",
    token: urlTokenParam || "",
  },
  onSubmit: async ({ value }) => {
    resetPassword(value);
  },
  validators: {
      onChange: ({ value }) => {
        const result = schema.safeParse(value);

        if (!result.success) {
          const errors = toErrors(result.error.issues);
          // where toErrors transforms zod format to { 'key1': 'errormessage1', 'key2': 'errormessage2', ...etc }
          return {
            fields: errors,
          }
        }
      },
    },
});
```
There is even some package for tanstack/react-form that does that. But it's sus, it's not in their main branch

It also would solve the issue with current code I just thought of: if user typed in a password, then matching confirmPassword, then changed the password, confirmPassword field will not raise an error. I actually didnt run your code and tried, but it feels like it will behave like that

I'm totally okay leaving validation per field. But it's worth checking and fixing this "changing password after you typed in proper confirmPassword" thing

PR #783, opened 2026-07-09
779: adds editable title to OrganisationDetails

DarrellRoberts:
## Description
Adds `title` to agent's `OrganisationDetails` with `EditableField`

## Related Issues
Closes #779 

## Changes
- Adds `title` to `organisationDetailsSchema`
- Add `title` to `OrganisationDetailsEdit`
- Add `title` to `OrganisationDetailsDisplay`
- Add `title` to `updateOrganization` mutation func

## Screenshots / Demos
image

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes


nadavosa (commented):
Small, clean, well-scoped PR — mirrors the existing pattern for every other field in this form (Controller + EditableField, same display/edit split, same i18n key structure). Nothing structurally off.

**Good catch, not a bug:** the diff removes the old `OrganizationDetailsPatch` local type in `useUpdateOrganizationDetails.ts`, which existed specifically because `addressStreet`/`addressPostcode` weren't yet in the SDK's `ApiAgentPatch`. Checked the SDK (`need4deed-org/sdk/src/types/api/agent.ts`) — `ApiAgentPatch` now includes `title`, `about`, `website`, `addressStreet`, and `addressPostcode`, so switching to it directly is legitimate cleanup of now-obsolete tech debt, not a regression.

**Worth confirming (not a confirmed bug):** the SDK's `ApiAgentTitleConflict` type documents a unique-title constraint enforced on agent *creation*, returning a distinctly-shaped `{ conflict: "title", agentId }` 409. The doc comment scopes it to CREATE specifically, so it likely doesn't apply to this edit flow — but worth a quick check with the `be` side that editing an existing org's title to match another org's name fails with a normal validation message rather than an unhandled/oddly-shaped error, since no special handling for that case was added here.

Minor: no tests added, but that's consistent with the rest of this codebase's convention for form fields like this — not a real ding.

DarrellRoberts:
thanks @nadavosa , I tried editing the agent title on the `be` with an existing agent title and got a `500 server error`. I wasn't able to confirm it was a validation error as I couldn't read the terminal (no longer possible on `local be` to create an `admin` account)

Nonetheless it still blocks the user from editing an agent's title to one that already exists

PR #784, opened 2026-07-09
🐛 fix: fold title into useUpdateOpportunityDetails; drop separate hook

arturasmckwcz:
## Summary

Follow-up to #781, now that the full contract is in place:

- Upgrades `need4deed-sdk` to 0.0.125 (`title` is now part of `ApiOpportunityPatch`)
- Passes `title` directly in `useUpdateOpportunityDetails` alongside description, languages, etc. — one PATCH call instead of two
- Deletes `useUpdateOpportunityTitle` (was firing a separate PATCH that the BE rejected with 400 due to `additionalProperties: false` on the old schema)

## Why the previous approach was wrong

`useUpdateOpportunityTitle` sent `{ title }` to `PATCH /opportunity/:id` but the BE schema had `additionalProperties: false` and `title` wasn't in `ApiOpportunityPatch`, so every save silently 400'd. be#743 added `title` to the schema and parser; sdk#151 added it to the type. This PR closes the loop on the FE side.


need4deed (approved):
LGTM

need4deed (approved):
LGTM

PR #785, opened 2026-07-09
🐛 fix: auth hint TTL, useAuth cleanup bug, logout cache clear

arturasmckwcz:
## Summary

- **`AUTH_HINT_MAX_AGE`** now equals `REFRESH_TOKEN_MAX_AGE` (7 days) via a new named constant, so the `is_logged_in` cookie stays valid as long as the refresh token
- **`useAuth`** refactored from `useState`/`useEffect` to direct derivation — the old `return () => setIsAuthorized(false)` cleanup was firing before every re-render, causing edit buttons to briefly disappear on each render cycle
- **`useLogout`** now calls `queryClient.removeQueries({ queryKey: ['user'] })` before clearing the auth hint and redirecting, so stale user data from the current session is not visible to the next one

## Test plan

- [ ] Log in, verify edit buttons are visible and stable on opportunity/volunteer profiles (no flicker)
- [ ] Log out, log back in as a different user — confirm old user data does not appear
- [ ] On localhost (HTTP), confirm `is_logged_in` cookie persists across page reloads after login

### DIFF
--- a/src/components/Dashboard/Profile/sections/ProfileHeader/volunteer/VolunteerHeader.tsx
+++ b/src/components/Dashboard/Profile/sections/ProfileHeader/volunteer/VolunteerHeader.tsx
@@ -51,7 +51,7 @@ type Props = {
 export const VolunteerHeader = ({ volunteer }: Props) => {
   const { t } = useTranslation();
   const dialog = useEngagementStatusDialog(volunteer);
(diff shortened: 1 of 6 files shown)


need4deed (approved):
LGTM

need4deed (approved):
LGTM

PR #787, opened 2026-07-11
Opportunity type change

kiborrdis:
## Description
Adds type switching dialog.
If switching to accompanying type, show accompanying details and forces to fill
If switching to event type, shows event date/time field and forces to fill

## Related Issues
Closes #753

## Changes
- OpportunityEventDateTimeEdit — extracted from OpportunityDetailsEdit (event date/time using useFormContext; availability grid stays inline)
- Added change type dialog shown for admin/coordinator

## Screenshots
image
image
image

When moving away from accompanying:

image

## Checklist
- WITHIN THE SCOPE OF AN ISSUE; No unnecessary files included
- Tests added/updated
- Documentation updated
- CI passes


nadavosa (commented):
Why are we with new boxes and not just the editable sections that appear?


Claude 

**What it does:** Adds a full "Change type" dialog for opportunities — simple type switches for Regular, plus dedicated sub-forms when switching into Accompanying (collects accompanying details inline, required) or Events (collects event date/time, required). Compared to #755, this correctly accounts for the `REGULAR_ACCOMPANYING` compound status in the SDK's `ProfileVolunteeringType` enum — #755's simpler `OpportunityType` enum use would leave that case with nothing pre-selected.

**🔴 Likely blocking — Events transition silently drops data:**
`useUpdateOpportunityType.ts` sends event date/time via an ad-hoc `eventDetails: { eventDate, eventTime }` field, with a comment: *"It's not yet implemented in api, for now keep it like that, change later."* I checked the SDK's `ApiOpportunityPatch` type directly — there is no `eventDetails` (or equivalent) field, only `schedule` for recurring availability. So switching an opportunity to Events via this dialog will likely PATCH a field the backend doesn't recognize — the type flips to Events, a success toast fires, but the event date/time the coordinator just filled in (and which the form requires) probably isn't persisted. This isn't disclosed in the PR description and bypasses the SDK-contract-first convention this project follows elsewhere. I'd treat this as blocking until there's a corresponding sdk/be change, or the Events path is gated off until then.

**🟡 Correctness bug (medium confidence) — REGULAR_ACCOMPANYING → ACCOMPANYING treated as no-op:**
```ts
const toDialogInitial = (type) => type === REGULAR_ACCOMPANYING ? ACCOMPANYING : type;
const isNoop = selected === toDialogInitial(currentType);
```
If an opportunity's current type is `REGULAR_ACCOMPANYING` and a coordinator explicitly selects "Accompanying" (intending to drop the "regular" half), this collapses to `ACCOMPANYING`, matching `selected`, so `isNoop = true`. Save then just calls `onCancel()` and never sends the PATCH — the dialog closes as if it succeeded, but the type never actually changes. Worth confirming whether that transition is expected to be reachable; if so, this needs a fix.

**Minor:** `AccompanyingTypeChangeForm.tsx`'s `useEffect(() => methods.reset(getInitialFormValues(undefined)), [methods])` right after `useForm` is initialized with the exact same `defaultValues` is redundant, though harmless.

No tests, but that's consistent with the rest of the codebase's convention for this kind of feature.

kiborrdis:
What do you mean by boxes and what do you mean by editable sections that appear?

Issue 1:
We discussed this in the slack, We need to finalize the decision on how do you want to approach eventTime/eventDate. 
Separate field in the opportunity table, accompanying details, or store in the schedule. After that this point can be resolved.

Issue 2:
Probably should be able to change to accompanying in that case. Removed toDialogInitial

need4deed:
@kiborrdis yes you are right, in case this want clear this issue https://github.com/need4deed-org/be/issues/746 adding to the table was to resolve this question no?

kiborrdis:
@need4deed As an option yes, or reuse accompanying details for the purposes of not duplicating data further. I'll ping discussion in slack

nadavosa:
## Follow-up review

Re-checked against the current state of things, since a lot has moved since my last pass here:

### Previous blocking finding — resolved ✅
The SDK bump to 0.0.128 in this PR brings a real `event: { date: string; time: string }` field on `ApiOpportunityPatch` (replacing the old ad-hoc `eventDetails` that had no backing contract). `need4deed-org/be#742` — merged into `develop` just now — implements the matching server-side logic: requires `event.date`/`event.time` when switching to Events, requires `accompanyingDetails` when switching to Accompanying with no existing accompanying record, and clears the `Accompanying` row when switching back to Regular. I traced all three sub-forms (`PlainTypeChangeForm`, `AccompanyingTypeChangeForm`, `EventTypeChangeForm`) against that backend validation and the payloads match. This is no longer a silent-data-loss situation — nice fix.

### Previous isNoop finding — resolved ✅
`toDialogInitial` is gone; `REGULAR_ACCOMPANYING → ACCOMPANYING` now correctly falls through to `PlainTypeChangeForm` and sends a real PATCH instead of treating it as a no-op.

### New: stale `eventDate`/`eventTime` defaults in `OpportunityDetailsEdit.tsx`
Not touched by this diff, but worth flagging now that event date/time actually persists (via `be#742`, backed by the `Accompanying` record): `OpportunityDetailsEdit.tsx`'s form still hardcodes `eventDate: null, eventTime: ""` in `defaultValues`, regardless of the opportunity's actual stored event date/time. Previously this was harmless because the field was never really persisted server-side either way; now that it is, opening "Edit" on an existing Events-type opportunity will always show a blank date/time, hiding whatever was set via the new type-change dialog. Not something this PR needs to fix, but worth a follow-up issue since this PR is what makes the gap user-visible for the first time.

### Minor: `EventTypeChangeForm` defaults `eventDate` to today
`defaultValues: { eventDate: new Date(), eventTime: "" }` — since the date is pre-filled but `eventTime` starts blank (and is the only field forced via required-state styling), a coordinator who only fills in the time could end up submitting today's date without ever consciously choosing it. Consider defaulting `eventDate` to `null` too so both are equally "untouched" until the user picks something, consistent with how `eventTime` behaves.

### Still open (unchanged from before)
`AccompanyingTypeChangeForm.tsx`'s `useEffect(() => methods.reset(getInitialFormValues(undefined)), [methods])` immediately after `useForm` is initialized with the exact same `defaultValues` is still redundant (harmless, just dead code).

Nothing blocking anymore from my side — the SDK/backend alignment was the big one and it's resolved. The stale-defaults issue is worth a quick follow-up but doesn't need to hold this PR up.

PR #798, opened 2026-07-14
fix: preserve volunteer param in opportunity table view (#739)

yi171:
## Description
fix bug where the `?volunteer` query param was dropped when clicking an opportunity from the list/table view, such that 'Suggest' button never appears.
fix is done via passing volunteerId prop from OpportunityListController down through the table list to OpportunityTableRow, similar to how opportunityId is passed down in VolunteerTableRow.

## Related Issues
Closes #739  

## Screenshots / Demos
screenshot-1784001073600

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes


DarrellRoberts (approved):
very nice, lgtm!

PR #799, opened 2026-07-14
fix: wire real data into agent profile Volunteers section

nadavosa:
## Description
`VolunteerAgents.tsx` (the "Ehrenamtliche"/Volunteers section on the NGO/agent's own profile) never fetched any data — `useTabTransitions` was called with a hardcoded `[]`, so every status tab always showed 0, even when the NGO genuinely had volunteers matched via its opportunities.

## Related Issues
Closes #777

## Changes
- Wires `VolunteerAgents` up to the new `GET /agent/:id/volunteer-linked` endpoint (need4deed-org/be#762 / need4deed-org/be#759), mirroring `OpportunityVolunteers.tsx`'s existing fetch/tabs/mutation pattern (same `useTabTransitions`, same `useUpdateOpportunityVolunteerStatus`/`useDeleteOpportunityVolunteer` hooks)
- Renders the previously-unused `AccordionVolunteer`/`VolunteerDetail` components (they existed fully built but were never imported by `VolunteerAgents.tsx`)
- `MappedVolunteerAgent` type now aliases `ApiVolunteerOpportunityGet` directly (the actual shape the endpoint returns) instead of an ad-hoc `ApiVolunteerGetList & { status }` combo
- Fixed two latent bugs in the never-rendered `AccordionVolunteer.tsx`:
  - Used `statusEngagement`/`statusType` field names that don't exist on `ApiVolunteerOpportunityGet` (the real fields are `engagement`/`volunteeringType`) — always evaluated to `undefined`
  - "Go to profile" linked to `volunteer.id` (the opportunity-volunteer match id) instead of `volunteer.volunteerId` — would have routed to the wrong profile page
  - Also removed a hardcoded placeholder date (`12.02.2025`) in favor of the real `updatedAt`

## Note
No backend available in my working environment to test the full round-trip live (needs need4deed-org/be#762 merged and deployed first). Typecheck, lint, and the pre-commit hooks all pass. The component logic mirrors `OpportunityVolunteers.tsx` closely — would appreciate a manual check once both PRs are merged to a shared environment.

## Depends on
- need4deed-org/be#762 (adds the `GET /agent/:id/volunteer-linked` endpoint this relies on)

## Checklist
- [ ] Tests added/updated — no dedicated tests, consistent with `OpportunityVolunteers.tsx`/`VolunteerOpportunities.tsx`, which also have no test coverage
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/src/components/Dashboard/Profile/sections/VolunteerAgents/AccordionVolunteer.tsx
+++ b/src/components/Dashboard/Profile/sections/VolunteerAgents/AccordionVolunteer.tsx
@@ -32,7 +32,7 @@ export const AccordionVolunteer = ({
   const router = useRouter();
 
   const handleGoToProfile = () => {
-    router.push(`/${i18n.language}/dashboard/volunteers/${volunteer.id}`);
+    router.push(`/${i18n.language}/dashboard/volunteers/${volunteer.volunteerId}`);
   };
 
   const engagementStatusLabels = createEngagementStatusLabelMap(t);
@@ -44,11 +44,8 @@ export const AccordionVolunteer = ({
       <Heading4 margin={0} color="var(--color-midnight)">
         {volunteer?.name}
       </Heading4>
-      <ProfileStatusBadge
-        status={volunteer?.statusEngagement}
-        label={engagementStatusLabels[volunteer?.statusEngagement]}
-      />
-      <ProfileStatusBadge status={volunteer?.statusType} label={statusLabels[volunteer?.statusType]} />
+      <ProfileStatusBadge status={volunteer?.engagement} label={engagementStatusLabels[volunteer?.engagement]} />
+      <ProfileStatusBadge status={volunteer?.volunteeringType} label={statusLabels[volunteer?.volunteeringType]} />
     </>
   );
 
@@ -56,7 +53,7 @@ export const AccordionVolunteer = ({
     <Accordion
       data-testid="volunteer-accordion"
       headerLeft={headerLeft}
-      subtitle={`${t(getDatePrefixKey(currentStatus))} 12.02.2025`}
+      subtitle={`${t(getDatePrefixKey(currentStatus))} ${new Date(volunteer.updatedAt).toLocaleDateString("de-DE")}`}
       onGoToProfile={handleGoToProfile}
     >
       <VolunteerDetail
--- a/src/components/Dashboard/Profile/sections/VolunteerAgents/VolunteerAgents.tsx
+++ b/src/components/Dashboard/Profile/sections/VolunteerAgents/VolunteerAgents.tsx
@@ -1,24 +1,88 @@
+import { apiPathAgent, cacheTTL } from "@/config/constants";
+import { useGetQuery } from "@/hooks/useGetQuery";
+import {
+  useDeleteOpportunityVolunteer,
+  useUpdateOpportunityVolunteerStatus,
+} from "@/hooks/useUpdateOpportunityVolunteerStatus";
+import { Id, OpportunityVolunteerStatusType } from "need4deed-sdk";
+import { useMemo } from "react";
 import { useTranslation } from "react-i18next";
 import { SectionEmptyState } from "../shared/styles";
 import { Tabs } from "../shared/Tabs";
-import { TAB_STATUS_ORDER, useTabTransitions } from "../shared/useTabTransitions";
+import { ITEM_STATUS_REMOVED, TAB_STATUS_ORDER, useTabTransitions } from "../shared/useTabTransitions";
+import { AccordionVolunteer } from "./AccordionVolunteer";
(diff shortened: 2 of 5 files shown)


arturasmckwcz:
## Review

### Overview
Fixes #777: `VolunteerAgents.tsx` was calling `useTabTransitions([])` with a hardcoded empty array, so the "Ehrenamtliche" tab on an agent's own profile always showed zero volunteers. This PR:
- Fetches from the new `GET /agent/:id/volunteer-linked` endpoint (be#762, still open)
- Mirrors the existing `OpportunityVolunteers.tsx` pattern almost verbatim (same hooks, same tab/status-override machinery, same mutation wiring)
- Finally renders the previously dead-code `AccordionVolunteer`/`VolunteerDetail` components
- Fixes two latent bugs in `AccordionVolunteer.tsx` that only mattered once the component was actually rendered: wrong field names (`statusEngagement`/`statusType` → `engagement`/`volunteeringType`) and wrong link target (`volunteer.id` → `volunteer.volunteerId`)
- Retypes `MappedVolunteerAgent` to alias `ApiVolunteerOpportunityGet` directly

### Correctness
- **Type choice is verified correct.** Checked the be#762 branch: the endpoint reuses the same DTO (`opportunityOpportunityVolunteerDTO`) and response schema as `GET /opportunity/:id/volunteer-linked`, so `ApiVolunteerOpportunityGet` is the right SDK type here — not the superficially similar but PII-masked `ApiAgentOpportunityVolunteer` type used elsewhere in the SDK for a different endpoint. Good call reusing the existing type rather than inventing one.
- **Field-name and link-target fixes are correct** against the SDK's `OpportunityVolunteer`/`ApiVolunteerOpportunityGet` shapes (`volunteerId`, `engagement`, `volunteeringType`, `updatedAt` all exist; the old names didn't).
- **Cache-invalidation gap (pre-existing, but now exposed a third way):** `useUpdateOpportunityVolunteerStatus`'s `getComplementaryPrefix` only knows two query-key prefixes:
  ```ts
  return queryKey[0] === "opportunity-volunteers" ? "volunteer-opportunities" : "opportunity-volunteers";
  ```
  This PR introduces a third caller with queryKey `["agent-volunteers", ...]`. Since it isn't `"opportunity-volunteers"`, it falls into the `else` branch and invalidates `"opportunity-volunteers"` — but never `"volunteer-opportunities"`, which is the view that actually needs refreshing (a volunteer's own dashboard showing their opportunity status) when an agent changes that volunteer's status from this new section. Worth updating `getComplementaryPrefix` (or the call site) to invalidate both complementary views when called from `agent-volunteers`.

### Consistency with existing pattern
`VolunteerAgents.tsx` is close to a line-for-line mirror of `OpportunityVolunteers.tsx` (confirmed by diffing both). Two small inconsistencies vs. the sibling, not bugs:
- `OpportunityVolunteers`'s empty state has `data-testid="volunteers-empty-state"`; the new one doesn't get an equivalent testid.
- `OpportunityVolunteersContainer` passes `id={"opportunity-volunteers-container"}`; the agent version doesn't set an analogous `id`.

### Pre-existing risk now made user-visible
`VolunteerAgents/VolunteerDetail.tsx` was dead code before this PR and isn't touched by it, but it's now rendered for real users for the first time. It's visibly diverged from the more-developed sibling `OpportunityVolunteers/VolunteerDetail.tsx`:
- No tag-overflow truncation (sibling caps at 3 tags + "+N"; this version renders all activity/skill tags unbounded — a volunteer with many skills could break the accordion layout)
- No language-proficiency grouping
- No `ActivityLog` integration for ACTIVE/PAST volunteers
- Uses older i18n keys (`dashboard.volunteerProfile.profileSection.*`) rather than the newer `dashboard.volunteers.*` keys the sibling now uses — both key sets still exist in the locale files, so nothing is broken, but it signals this component predates a UI refresh its sibling received.

Out of scope for this PR, but worth a follow-up ticket since it's no longer unreachable code.

### Testing
No dedicated tests, consistent with the sibling components. Author noted be#762 isn't merged yet — checked that branch and it already includes a fix (commit `1686d97b`) for a response-schema bug that would have 500'd this exact endpoint on any non-empty result, so the round-trip should work once both land. Worth the manual smoke test the author requested once merged/deployed together.

### Verdict
Solid, well-scoped bug fix that correctly reuses an established pattern and fixes two genuine latent bugs. One real (if pre-existing) cache-invalidation gap worth addressing, and one heads-up about newly-user-visible UI debt that should get its own follow-up rather than block this PR.

nadavosa:
Fixed the cache-invalidation gap — \`getComplementaryPrefix\` is now a lookup table keyed by all three query-key prefixes, and \`agent-volunteers\` invalidates both \`volunteer-opportunities\` and \`opportunity-volunteers\` on status update/delete, per your suggestion. Typecheck/lint clean.

PR #802, opened 2026-07-15
790: agent own profile missing profilelayout

DarrellRoberts:
## Description
Fixes missing profile layout for agent's own profile (when they click on profile)

## Related Issues
Closes #790 

## Changes

## Screenshots / Demos
image

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/src/app/[lang]/dashboard/profile/page.tsx
+++ b/src/app/[lang]/dashboard/profile/page.tsx
@@ -1,6 +1,6 @@
 "use client";
 import CenteredWrapper from "@/components/core/common/CenteredWrapper";
-import { AgentProfileController } from "@/components/Dashboard/Profile/AgentProfileController";
+import ProfileLayout from "@/components/Dashboard/Profile/ProfileLayout";
 import { Paragraph } from "@/components/styled/text";
 import { useGetCurrentAgent } from "@/hooks/useGetCurrentAgent";
 import { useTranslation } from "react-i18next";
@@ -25,5 +25,5 @@ export default function DashboardProfilePage() {
     );
   }
 
-  return <AgentProfileController entityId={String(agentId)} />;
+  return <ProfileLayout entityId={String(agentId)} entityType={"agent"} />;
 }


DarrellRoberts:
^this is the quick fix solution.

-----
If we wanted something more scalable, we could do something like this for the `DashboardRoutes` in the `NavigationBar.tsx`

```typescript
  const profileLinkMap = {
    [UserRole.AGENT]: user?.agentId ? `/dashboard/agent/${user?.agentId}` : DashboardRoutes.Profile,
    [UserRole.USER]: DashboardRoutes.Profile,
    [UserRole.VOLUNTEER]: DashboardRoutes.Profile,
    [UserRole.COORDINATOR]: DashboardRoutes.Profile,
    [UserRole.ADMIN]: DashboardRoutes.Profile,
  };
...
    {
      label: t("dashboard.home.sidebar.profile"),
      text: userInitials,
      route: user && user.role ? profileLinkMap[user.role] : DashboardRoutes.Profile
    },
```

That way, we're prepared if we wanted for example, a `volunteer` user to view their own profile that way.
Currently the `/profile` page serves only `agent` users

nadavosa:
## Review

**What it does:** Fixes #790 — the NGO/agent self-view profile page (`/dashboard/profile`) rendered `<AgentProfileController>` directly, without the `DashboardLayout` wrapper the rest of the dashboard uses, causing content to be truncated/overlapped by the sidebar. Swaps to `<ProfileLayout entityId={...} entityType="agent">`.

**Correctness:** Verified this is not a shortcut — `ProfileLayout` wraps `ProfileController` in `DashboardLayout`, and `ProfileController`'s `entityType="agent"` dispatches to the exact same `AgentProfileController` that was rendered before, so nothing about the agent-profile data/rendering logic changes — only the missing layout wrapper gets added. Also confirmed this brings `/dashboard/profile` in line with the already-established pattern: `/dashboard/agents/[id]/page.tsx` (viewing *other* agents' profiles) already uses `ProfileLayout` + `entityType="agent"` identically. This was the one inconsistent page.

**One minor observation, not blocking:** the two early-return states above the fix (`isLoading` and `!agentId` / "not set up yet") still return bare `<CenteredWrapper>` without `DashboardLayout` — unchanged by this PR. If the truncation bug can also show during the brief loading flash or the "profile not set up" empty state, it'd still reproduce there. Given #790's repro steps describe the *loaded* profile content specifically, this is likely fine as-is (a brief loading/empty placeholder arguably doesn't need full dashboard chrome), just flagging in case those states matter too.

**Risk:** Very low — 2-line diff, swaps to an already-proven, identical pattern used elsewhere in the codebase.

No blockers from this pass.

PR #805, opened 2026-07-15
fix: send postcode to registration address-match search

nadavosa:
## Summary

- The registration completion step's address-match picker (`GET /agent/register/search`) has required both `street` and `postcode` since https://github.com/need4deed-org/be/commit/fbcd4426ad72aad154667c7c5054f3dfd63e78cd (2026-06-11), but `useAgentAddressLookup` never sent `postcode` — only `street`.
- Result: the picker has always returned zero matches since that backend change landed, so every registrant silently ended up creating a duplicate NGO instead of being offered to join the existing one.
- Fix: pass `formData.addressPostcode` through `useAgentAddressLookup`, debounce it the same way as the street, include it as `&postcode=` on the search request, and require it in `enabled` (matching the backend's own guard).

Closes https://github.com/need4deed-org/fe/issues/804

## Test plan

- [x] `yarn typecheck` — clean
- [x] Manually verify: register a new NGO account, complete profile with an address that matches an existing agent in the same email domain, confirm the match banner + checkmark now appears


DarrellRoberts:
from the offset it would appear to work:

image

This check also disappears/appears whenever I change either the street address or postcode so it is responsive.

However, would you not want to disable the `Next` button until the user has confirmed?

DarrellRoberts:
e.g. on `/components/AgentRegistration/ProfileCompletion.tsx` , add `showBanner` tp `disabled boolean`
```typescript
            <Button
              text={t("agentRegistration.next")}
              backgroundcolor="var(--color-aubergine)"
              textColor="var(--color-white)"
              onClick={handleNext}
              disabled={tokenMissing || showBanner}
            />
```

nadavosa:
## Review

### Overview
Root-causes and fixes a real bug: the backend's `GET /agent/register/search` has required `postcode` since 2026-06-11, but `useAgentAddressLookup` never sent it, so the registration address-match picker has been returning zero matches ever since — silently sending every registrant down the "create duplicate NGO" path instead of offering to join the existing one. The fix threads `formData.addressPostcode` through the hook, debounces it like `street`, adds it to the query key/URL, and requires it in `enabled`. Good, well-explained PR description tying the fix to the specific backend commit and issue.

### Prior review feedback — not yet addressed
@DarrellRoberts's comment on this PR suggested disabling the "Next" button while `showBanner` is true, so a user can't advance past a detected match without either confirming or explicitly dismissing it. That hasn't been applied yet — in the current diff, the `Next` button (`ProfileCompletion.tsx`) is still only `disabled={tokenMissing}`, with no `showBanner` in the condition. Worth resolving before merge per the earlier suggestion (`disabled={tokenMissing || showBanner}`).

### Correctness
- **Dismiss tracking isn't postcode-aware.** `isDismissed = dismissedAddress === debouncedAddress` in `useAgentAddressLookup.ts` keys the dismissal only off the street value. Now that postcode is a first-class part of the match query (this PR's change), a user could: get a match banner, dismiss it, then change only the postcode to a value that legitimately matches a *different* agent — but since `dismissedAddress` still equals `debouncedAddress`, `isDismissed` stays `true` and the new, different match never shows. Consider keying dismissal on the pair, e.g. `dismissedAddress === \`${debouncedAddress}|${debouncedPostcode}\``, so a postcode change re-opens the possibility of a (different) match banner.

### Test coverage
- No unit test exists (or was added) for `useAgentAddressLookup`. Given this hook has non-trivial state (debounce, dismiss, match/select logic) and this is the second bug found in it in as many months, a test covering "match found → dismiss → postcode changes → new match appears" would help prevent regressions here going forward. Not blocking, but worth considering.

Solid, well-scoped fix for the actual reported bug. Main open item is the unaddressed "disable Next while a match banner is showing" suggestion from the earlier review comment.

PR #821, opened 2026-07-17
🐛 fix: send numeric option ids for opportunity-create activities/skills/languages

arturasmckwcz:
## Summary
- `buildCreatePayload` (dashboard's `NewOpportunity.tsx`, `POST /opportunity/`) sent stringified numeric option ids through fields typed for the legacy public form's free-text/ISO-code strings, so `be`'s create handler silently dropped all activities, skills, and languages on save — need4deed-org/be#774.
- The local `OpportunityFormDataWithAgentSubmitter` type was defined here (with a comment saying "not yet in SDK") and masked the mismatch: both the legacy and dashboard shapes were `string[]`, just carrying different meanings.
- `need4deed-sdk@0.0.127` now exports this type properly, with its own id-based fields (`languageIds`/`activityIds`/`skillIds`/`districtIds: number[]`). Dropped the local shadow type and updated `buildCreatePayload` to send real numbers through the new fields.

## Test plan
- [x] `yarn typecheck`
- [x] `yarn lint` (no new warnings)
- [x] `yarn build`
- [x] Verified end-to-end against need4deed-org/be#775: ran a local `be` (fix branch) + this `fe` branch together, created an opportunity through the fe dev server's own `/api/*` proxy with the new payload shape, and confirmed in the DB that activities/skills/languages persisted correctly (previously all silently empty)
- Note: no browser-automation tool was available in this environment, so the click-through UI itself wasn't exercised — the verification above drove the real network path (fe proxy → be → DB) with the exact payload `buildCreatePayload` produces.

Depends on need4deed-org/be#775 being merged/deployed for the fix to take effect end-to-end (this PR alone changes only the request payload).

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "date-fns": "^4.1.0",
     "email-validator": "^2.0.4",
     "i18next": "^25.3.2",
-    "need4deed-sdk": "0.0.125",
+    "need4deed-sdk": "0.0.127",
     "next": "15.3.8",
     "react": "^19.0.0",
     "react-day-picker": "^9.13.0",
--- a/src/components/Dashboard/NewOpportunity/NewOpportunity.tsx
+++ b/src/components/Dashboard/NewOpportunity/NewOpportunity.tsx
@@ -39,34 +39,14 @@ import { Heading2, Heading4 } from "@/components/styled/text";
 import { ShootingStarIcon, ArrowLeftIcon } from "@phosphor-icons/react";
 import { de, enUS } from "date-fns/locale";
 import { TFunction } from "i18next";
-import { Lang, OptionItem, TranslatedIntoType, VolunteerStateTypeType } from "need4deed-sdk";
-
-// Not yet in need4deed-sdk — defined locally until the SDK is updated.
-type OpportunityFormDataWithAgentSubmitter = {
-  title: string;
-  opportunity_type: "accompanying" | "volunteering";
-  vo_information: string | null;
-  volunteers_number: number;
-  languages: string[];
-  activities: string[];
-  skills: string[];
-  timeslots: [number, string][] | null;
-  onetime_date_time: string | null;
-  accomp_address: string | null;
-  accomp_postcode: string | null;
-  accomp_datetime: string | null;
-  accomp_name: string | null;
-  accomp_phone: string | null;
-  accomp_information: string | null;
-  accomp_translation: `${TranslatedIntoType}` | null;
-  berlin_locations: string[] | null;
-  category: string;
-  category_id: string;
-  language: `${Lang}`;
-  agent_id: number;
-  submitted_by_id: number | null;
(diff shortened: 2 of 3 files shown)


nadavosa:
## Review

**What it does:** Fixes most of #774 by replacing the local shadow type in `NewOpportunity.tsx` with the SDK's new `OpportunityFormDataWithAgentSubmitter` (published as `need4deed-sdk@0.0.127`), sending real numeric IDs (`languageIds`/`activityIds`/`skillIds`/`districtIds`) instead of stringified values through fields previously typed for the legacy form's free-text/ISO-code strings.

**Verified independently:**
- Confirmed `need4deed-sdk@0.0.127` is actually published on npm.
- Read the SDK source directly: `OpportunityCreateFormDataProps` is a genuinely separate type from `OpportunityLegacyFormDataProps` (not derived from it), with a doc comment explaining why — exactly the "don't overload one shared type for two incompatible shapes" direction from #774.
- `districtIds: null` is valid: the field is optional in the SDK type, and `VoidableUndefined<T>` allows `null` on already-optional fields.
- Checked the companion `be` PR (#775): adds a genuinely new `dealParserOpportunityCreate` using `findBy({ id: In(ids) })`, and explicitly leaves `dealParserOpportunity`/`getProfileEntityByTitle`/`getLanguageTitle`/`/opportunity/legacy` untouched.
- No leftover references to the old `languages`/`activities`/`skills`/`berlin_locations` field names anywhere in the file.

**One real, unresolved gap — not introduced by this PR, but worth flagging before considering #774 fully closed:**

`buildCreatePayload` still merges three semantically distinct language groups into one flat array before sending:
```ts
const languageIds = [...new Set([...mainLangIds, ...residentsLangIds, ...refugeeLangIds])];
```
By the time this reaches `be`, there's no way to tell which ID was "main communication" vs "residents speak" vs "refugee language." `dealParserOpportunityCreate` (be#775) tags every resolved language with `purpose: LangPurpose.RECIPIENT` uniformly (mirroring what the legacy parser has always done — checked, this isn't a new regression, it's pre-existing on both code paths). But `OpportunityDetailsDisplay.tsx` filters "Main communication" specifically by `LangPurpose.GENERAL` — so even after this PR + be#775 merge, Main communication will show empty (no more "Ghotuo", but still wrong), and Residents Speak will incorrectly absorb whatever was meant for Main Communication.

Fixing this needs changes on both sides (FE sending purpose-tagged groups instead of one merged array, BE tagging accordingly per group), so it's a reasonable follow-up rather than scope creep on this PR — just flagging so #774 doesn't get closed as fully resolved when it isn't quite.

**Risk of this PR as scoped:** Low — does exactly what it claims, correctly, and the SDK/BE dependency chain is real and verified.

need4deed (approved):
Reviewed — matches the be#775 SDK contract, typecheck/lint/build clean vs. develop baseline, CI lint passing. Approving.

PR #823, opened 2026-07-18
feat: make card grid responsive

### LINKED ISSUE #665: feat: responsive card grid for volunteers & opportunities (currently fixed at 3 columns)
### Summary
The card view on the volunteers and opportunities lists is not responsive. It
renders a fixed 3-column grid of fixed-width (320px) cards, left-aligned, so wide
screens waste a large right gutter — it looks like another card would fit.

### Measurements (card view)
| Viewport | Columns | Grid width | Wasted right gutter |
|----------|---------|------------|---------------------|
| 1440px   | 3       | 992px      | ~322px              |
| 1728px   | 3       | 992px      | ~542px (~1.7 cards) |
| 1920px   | 3       | 992px      | ~638px (~2 cards)   |

### Technical notes
- Grid: `src/components/core/paginatedGrid/PaginatedGrid.tsx`
  (`grid-template-columns: repeat(${columns}, 1fr)`).
- `columns` comes from `CARD_COLUMNS = 3` (`src/config/constants.ts`), only
  reduced by 1 when the filter panel is open — never tied to viewport width.
- Cards have fixed `width: 320px` (`--dashboard-volunteers-card-width`, used by
  both card types), so the grid collapses to ~992px and sits left-aligned.

### Important coupling — needs a decision
Page size is derived from the grid: `itemsPerPage = columns * rows`
(`PaginatedGrid.tsx`). Making columns fluid changes how many items are fetched
per page, so we must decide how pagination behaves:

- **Option A (small):** `repeat(auto-fill, minmax(320px, 1fr))`, keep a fixed
  page size (e.g. 9 or 12); last row may be unevenly filled. Low risk.
- **Option B (larger):** compute columns from container width and recompute
  `itemsPerPage` on resize (refetch on resize). Evenly filled rows, more logic.

yi171:
## Description
the card grid was previously hard coded to 3 columns, and `isFiltersOpen` needed to be passed to the controllers to manually subtract a column when the filter was open.

the dashboard width ,` var(--dashboard-base-container-width)` in `components/styled/container.tsx `was also capped such that only a maximum of three 320px cards could fit in the container.

fix: the grid now uses CSS `auto-fill` to automatically calculate columns based on the container width, which also removes the need for using `isFiltersOpen`.  

` var(--dashboard-base-container-width)` was changed from 1340 px to 1600 px, and checked that the rest of the pages  (Opportunities,  Volunteers, Agents, Homepage ) still display normally after the change

number of items fetched is hardcoded to `12`, such that whether there are 2, 3, or 4 columns, the bottom row can be fully filled. there is no cap on max columns since the dashboard width is already capped, and can fit at most four 320px cards.

## Related Issues
Closes #665  

## Changes
- **`PaginatedGrid.tsx`**: replace hard-coded `columns` and `rows` with  `itemsPerPage`. update grid CSS to use `repeat(auto-fill, min(100%, var(--card-width, 320px)))`.
- replace columns/rows with itemsPerPage across all CardList and ListController components. 
- : remove `isFiltersOpen` from `Agents.tsx`, `Opportunities.tsx`, `Volunteers.tsx` and all ListControllers
- **`styles.ts`**:  add `--card-width` variables to list containers and `flex: 1` 
- **`constants.ts`**: set `CARD_LIMIT` to `12`
- increase ` var(--dashboard-base-container-width)` from 1340 px to 1600 px,

## Screenshots / Demos
**4 cols**
Screenshot 2026-07-18 at 11-16-20 Need4Deed
**3 cols**
Screenshot 2026-07-18 at 11-16-37 Need4Deed

**2 cols**
Screenshot 2026-07-18 at 11-16-57 Need4Deed
**with filter open**
Screenshot 2026-07-18 at 11-27-05 Need4Deed

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes


arturasmckwcz:
## Review: feat: make card grid responsive

**Summary:** Replaces the hardcoded 3-column `PaginatedGrid` with a CSS `grid-template-columns: repeat(auto-fill, min(100%, var(--card-width)))` layout, removes the `isFiltersOpen`-driven column/row juggling in `Agents`/`Opportunities`/`Volunteers`, and switches pagination from `columns × rows` to a flat `itemsPerPage` (hardcoded `CARD_LIMIT = 12`). Net effect: grid columns now respond to container width instead of being manually recalculated when the filter panel opens/closes. 13 files, +34/−60 — a clean, well-scoped diff that matches its stated intent, and the removal of `CARD_COLUMNS`/`CARD_ROWS`/`isFiltersOpen` plumbing is complete (verified no dangling references remain).

### Strengths
- `PaginatedGrid` is simplified correctly: `itemsPerPage` replaces the derived `columns * rows`, and the styled `GridContainer` no longer needs props (`columns`/`rows` → plain `styled.div`), which is a nice simplification.
- The `isFiltersOpen` removal is thorough — it's still used locally in `Agents.tsx`/`Opportunities.tsx`/`Volunteers.tsx` to toggle the `Filters` panel itself, but correctly no longer threaded through to `CardList`/`PaginatedGrid`.
- `CARD_LIMIT` is now a single literal (12) rather than a derived `3*3`, and every consumer (`AgentListController`, `OpportunityListController`, `VolunteerListController`) was updated consistently.
- Repo has no test framework wired up for these dashboard components, so the unchecked "Tests added" box is consistent with existing convention, not a gap introduced by this PR.

### Issues / risks

1. **Volunteers container doesn't wire up `--card-width` (inconsistency, currently masked).** `AgentCardListContainer` and `OpportunityCardListContainer` both set `--card-width: var(--dashboard-*-card-width)`, so their grid tracks track the actual card width. `VolunteerCardListContainer` (`VolunteerCardList.tsx:16-20`) only got `flex: 1; min-width: 0;` — it never sets `--card-width`, so it silently falls back to `PaginatedGrid`'s hardcoded default of `320px`. This happens to work today only because `--dashboard-volunteers-card-width` also equals `320px` in `globals.css`. If that variable is ever changed (e.g. a future responsive breakpoint for volunteer cards), the Volunteers grid will silently drift out of sync with the actual card width while Agents/Opportunities won't. Worth adding `--card-width: var(--dashboard-volunteers-card-width);` there for consistency.

2. **No cap on max columns, but `CARD_LIMIT=12` assumes a max of 4.** The PR description explicitly reasons "whether 2, 3, or 4 columns, the bottom row can be fully filled" — but `grid-template-columns: repeat(auto-fill, min(100%, var(--card-width, 320px)))` has no upper bound, and there's no `max-width` constraining the dashboard content area (`DashboardLayout.tsx` has none). On a wide/ultrawide viewport, the container could fit 5+ 320px columns, at which point 12 items no longer divides evenly and the last row will be partially filled — the exact problem this PR set out to fix, just re-appearing at a wider breakpoint. Not a regression (the old fixed-3-column grid never adapted at all), but worth either capping the dashboard content `max-width` or explicitly accepting this as an edge case.

3. **Naming: `--dashboard-volunteers-card-width` is reused for the Opportunities container.** `OpportunityCardListContainer` sets `--card-width: var(--dashboard-volunteers-card-width)` — this is a pre-existing naming quirk in the file (the `Opportunities` module already reuses `volunteers`-prefixed variables like `--dashboard-volunteers-container-gap` throughout), not something introduced by this PR, so not a new bug — but it's a confusing name to build on.

### Minor
- `flex: 1; min-width: 0; width: 100%;` is added to both `AgentCardListContainer` and `OpportunityCardListContainer`, but `width: 100%` is redundant alongside `flex: 1` in a row flex container — harmless, but could be dropped for consistency with the Volunteers container which omits it.
- `justify-items: start` on `GridContainer` is a good touch to prevent the last partial row's cards from stretching.

Overall a solid, focused refactor. Item 1 is a one-line fix worth making before merge; item 2 is worth a comment/decision from the author but not necessarily a blocker.

PR #824, opened 2026-07-18
fix: hide volunteer matching actions from NGO/agent view (#815)

yi171:
## Description
fix bug where NGO/agent users can see 'Match', 'Not a match' buttons etc. on a Volunteer profile, which gave "Permission denied" errors when clicked. 

Additional fix: after hiding the matching action buttons, agents can still expand the accordion drop-down next to 'Go to Profile' to see empty space. Hence `AccordionOpportunity.tsx` and `Accordion.tsx` are also edited to render the accordion only if there is content to be expanded.

## Related Issues
Closes #815

## Changes
- **`VolunteerOpportunities.tsx`**: gate handler props behind the existing `useAuth().isAuthorized` hook
- **`AccordionOpportunity.tsx`**: add `hasActions` and `hasActivityLog` to check if there is content for the accordion. `OpportunityDetail` child is now only passed if there is content to display.
- **`Accordion.tsx`**:  wrap `CaretIcon` so it only renders if children are provided

## Screenshots / Demos
before updating the accordion
Screenshot 2026-07-18 at 14-58-04 Need4Deed

after updating accordion
image

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes


arturasmckwcz (approved):
thanks a lot! LGTM

PR #825, opened 2026-07-18
fix: send activityIds/skillIds/languageIds on opportunity create

arturasmckwcz:
## Summary

Follow-up to need4deed-org/be#774 (and its be-side fix, need4deed-org/be#776).

`NewOpportunity.tsx` defined `OpportunityFormDataWithAgentSubmitter` locally (with a `// Not yet in need4deed-sdk` comment) and had `buildCreatePayload()` send `languages`/`activities`/`skills` as stringified numeric option ids. The `be` route these hit resolves those fields by title/ISO-code, so an id string never matches — every opportunity created from the dashboard silently lost its activities/skills/languages.

- Bumps `need4deed-sdk` to `0.0.127`, which now publishes this form's own contract (`OpportunityFormDataWithAgentSubmitter`) with `activityIds`/`skillIds`/`languageIds`/`districtIds: number[]` fields — imported instead of redefined locally.
- `buildCreatePayload()` now sends numeric `activityIds`/`skillIds`/`languageIds` (via `.map(i => i.id)` instead of `.map(i => String(i.id))`).
- `opportunity_type` now uses the SDK's `OpportunityLegacyType` enum instead of raw string literals (required once the field type came from the SDK).
- `districtIds` is sent as `null` (no UI collects it here, mirroring the previous `berlin_locations: null`).

## Test plan

- [x] `yarn typecheck` — passes
- [x] `yarn lint` — no new warnings/errors vs. `develop`
- [x] `yarn build` — production build succeeds
- [ ] Manual end-to-end: create an opportunity from the dashboard with activities/skills/languages selected, reopen it, confirm they persisted (needs the `be` fix in #776 deployed together with this)

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "date-fns": "^4.1.0",
     "email-validator": "^2.0.4",
     "i18next": "^25.3.2",
-    "need4deed-sdk": "0.0.125",
+    "need4deed-sdk": "0.0.127",
     "next": "15.3.8",
     "react": "^19.0.0",
     "react-day-picker": "^9.13.0",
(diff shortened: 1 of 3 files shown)


arturasmckwcz:
Closing as a duplicate of #821, which fixes the same thing (and was verified end-to-end against need4deed-org/be#775, not just statically). Superseded by #821.

PR #826, opened 2026-07-18
Polo/refactor new opportunity

polomontague:
## Description
Cleaned up the complex NewOpportunity component by splitting up it's multiple responsibilities into single responsibility components. Also split the fields in OpportunityDetailsFields up into their own components in the new ./fields directory so that OpportunityDetailsFields can cleanly handle just the conditional rendering logic for the 3 different opportunity types.

## Related Issues
Closes #768 (mandatory if applicable)

## Changes
- Moved StylesComponents into ./styled.ts
- Moved header schema into headerSchema.ts
- Split OpportunityDetailsFields into a separate module for each field so that its conditional rendering logic for the 3 types could be clearly deciphered. Now OpportunityDetailsFields only handles the conditional rendering.
- Moved helper functions into new ./helpers.ts module
- Split the header section out of “NewOpportunity” into it’s own component “OpportunityHeaderCard”

## Screenshots / Demos

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes


arturasmckwcz:
@polomontague the same as with #827 please resolve merge conflicts

src/components/Dashboard/NewOpportunity/NewOpportunity.tsx
```diff
@@ -525,49 +148,9 @@ export function NewOpportunity() {
       <Heading2>{t("dashboard.newOpportunity.title")}</Heading2>
 
       {/* Header card — title input + volunteer type selector */}
```
DarrellRoberts: could you also remove the comments? These are left over from when the AI created the component I think

DarrellRoberts (commented):
just wanted to add a small point

src/components/Dashboard/NewOpportunity/NewOpportunity.tsx
polomontague: I've gone ahead and removed all the comments from NewOpportunity.tsx

PR #822, opened 2026-07-17
Remove agent contact address

ivannissimrch:
Summary
  - Per the standup decision (2026-07-17): a contact person doesn't need an address; the organization address details are the only ones we keep. This removes the address input from the Kontaktdaten section.
  - Removed: the address form field, the display row, the zod schema entry, the address object from the PATCH payload, the en/de translation keys, and the now-unused agentAddressUtils (formatAddress/parseAddress). 
 
- Note: existing contact address rows are still in the database.
@arturasmckwcz related to #778:  the 500 error was happening when a contact with a NULL address was edited and saved; contacts with an address saved fine. Before removing the address field(this PR)
On my local BE while testing, I stopped it with two changes: (1) guarding the read in `person.routes.ts` for `person.address?.postcodeId` and (2) skipping empty address objects in `dtoParsePerson`. But this PR stops sending the address field entirely. Will this PR make these changes on the BE unnecessary? 

The branch is based on #808 and includes its SDK bump and agentId commits, since I needed that for testing.

## Screenshots / Demos
https://github.com/user-attachments/assets/085d4911-f036-40e6-81c8-d0c0fdb7f0c6

## Checklist
- [ ] WITHIN THE SCOPE OF AN ISSUE; No unnecessary files included
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes

Issue #809, opened 2026-07-16
bug: editing contact address doesn't fix a corrupted duplicate-postcode display (e.g. "12051, 12051")


nadavosa:
## Problem

On a volunteer's profile, the "Kontaktdaten" (Contact details) section sometimes shows the address as two postcodes, e.g. `Adresse: 12051, 12051`. Editing the field and saving shows a success toast ("Kontaktdaten erfolgreich aktualisiert"), but the address still displays as two postcodes afterward — the edit doesn't actually fix it.

Reported as happening on opportunity/agent profiles too, not just volunteers.

## Where

`src/components/Dashboard/Profile/sections/ContactDetails/volunteer/VolunteerContactDetails.tsx`, backed by `volunteerAddressUtils.ts`.

## Likely root cause

The address is edited as a **single free-text input** (`VolunteerContactDetailsEdit.tsx`), pre-filled by joining the structured address into one string:

```ts
// volunteerAddressUtils.ts
export const formatAddress = (address) => {
  ...
  return [address.street, address.city, postcode].filter(Boolean).join(", ");
};

export const parseAddress = (addressString) => {
  const [street = "", city = "", postcode = ""] = addressString.split(",").map((part) => part.trim());
  return { street, city, postcode };
};
```

`formatAddress` **filters out empty segments** before joining, but `parseAddress` assumes a **fixed 3-part position** (street, city, postcode) when splitting back. If `street` is empty, `formatAddress` produces a 2-part string (e.g. `"12051, 12051"` — city and postcode both happen to be the postcode value already). When that gets parsed back on save, the position mapping is now wrong: the first comma-separated part lands in `street` and the second in `city`, with `postcode` defaulting to empty — so the round-trip can silently corrupt the address further, or simply cannot represent/repair a record that's missing a segment.

This is a duplicated pattern — the exact same `formatAddress`/`parseAddress` logic (byte-for-byte) also exists in `src/components/Dashboard/Profile/sections/ContactDetails/agent/agentAddressUtils.ts`, so the same failure likely reproduces on agent/opportunity contact editing too.

## Steps to reproduce

1. Find a volunteer (or agent) profile whose stored address is missing a street (so `city`/`postcode` both end up holding a postcode-like value, e.g. "12051")
2. Open Kontaktdaten → Bearbeiten
3. The address field pre-fills as `"12051, 12051"` with no way to tell which part is city vs. postcode
4. Edit and save
[...]

PR #828, opened 2026-07-19
fix: split volunteer contact address into street and postcode fields

ivannissimrch:
## Description

  The volunteer contact address was edited as a single free-text field, joined from the structured address on load and split back by comma position on save. `formatAddress` dropped empty segments before joining, but `parseAddress` assumed a fixed 3-part order, so a record missing a street round-tripped into the wrong fields and couldn't be repaired through the UI.

This replaces that field with separate **street** and **postcode** inputs and removes the format/parse round-trip entirely.

`city` is sent as an empty string on save. This is deliberate: on a PATCH, omitting the field leaves the corrupted value in the database, so the edit would looksuccessful while the bad data survived. Sending `""` actually clears it.

  ## Related Issues

  Refs #809  deliberately not closing, see scope below.

  ## Changes

  - Split the address field into `street` and  `postcode` same pattern as registration
  - Deleted `volunteerAddressUtils.ts`; the position-based `split(",")` parse was the corruption source and now has no callers
  - Both fields required,
  - Display joins with `filter(Boolean)` so a record with a missing segment renders `"12051"`, not `", 12051"`
  - Added `street` / `postcode` labels and validation keys on en, de; removed the now-unused `addressRequired.`

  ## Scope

 Volunteer only. 
**Agent contact address** is being removed entirely in #822, per the 07/17 standup, and will use Organisation details, which is already split and working correctly.
 **Opportunity details** has No address field  there is a section on oportunies  **Refugee accommodation center** but is a be issue see below

  ### Refugee accommodation center on Opportunities details
  
  The FE sends the address as one string to `PATCH /opportunity/:id`. The BE parser only saves the name and drops the address and district. Verified on opportunity 12:
The PATCH payload has the new address; the GET after save returns `agent.address: ""`, and only `name` persisted.
Splitting the field on the FE wouldn't fix it: the SDK type is a flat string, so I'd have to join the two inputs back into one string to send, which is the same round-trip this PR removes. Needs the SDK shape and BE parser first.
  ## Checklist

  - [ ] Tests added/updated 
  - [ ] Documentation updated 
  - [x] CI passes


arturasmckwcz (approved):
@ivannissimrch thanks a lot, LGTM

PR #829, opened 2026-07-20
fix: restrict "Main communication" language field to German/English

arturasmckwcz:
## Summary

- Closes #813: the New Opportunity and Opportunity Details "Main communication" language field showed the full master language list instead of a restricted set.
- Restricts the dropdown for that field (in both `NewOpportunity.tsx` and `OpportunityDetailsEdit.tsx`) to German/English only, via a shared `getMainCommunicationLanguageOptions` helper. "Residents speak" keeps the full list unchanged.
- Adds schema-level validation (`opportunityDetailsSchema.ts`) enforcing that the resolved selection is one of: none, German alone, or German+English together — never English alone or any other language, since the dropdown restriction alone doesn't stop that combination.
- Adds i18n strings for the new validation message (en/de).

Pairs with a matching backend PR that enforces the same rule server-side on `PATCH /opportunity/:id`: need4deed-org/be#778

## Test plan

- [x] `yarn typecheck` passes
- [x] `yarn lint` passes on changed files
- [ ] Manually verify in the dashboard: creating/editing an opportunity only offers German/English for "Main communication", and saving with just English shows the new validation error


nadavosa:
## Review

### Overview
Restricts the "Main communication" language dropdown to German/English in `NewOpportunity.tsx` and `OpportunityDetailsEdit.tsx` via a new `getMainCommunicationLanguageOptions` helper, and adds `superRefine` validation in `opportunityDetailsSchema.ts` enforcing none/German-only/German+English. Straightforward, well-scoped change with matching i18n strings and a paired backend PR (need4deed-org/be#778) for server-side enforcement.

### Correctness
- **Legacy/out-of-set values silently bypass validation.** `resolveLanguageTitle` only resolves against the restricted `mainCommunicationLanguageOptions` (German/English). If an existing opportunity's `mainCommunication` currently holds a different language (e.g. saved before this restriction existed), that entry resolves to `undefined` and gets filtered out of the `titles` set entirely — so `isNone` becomes true and validation passes even though the underlying (unresolved) row is still submitted via `toLangOptionItems` against the *full* `apiLanguages` list in `OpportunityDetailsEdit.tsx`/`NewOpportunity.tsx`. Net effect: editing such an opportunity without touching "Main communication" shows no validation error client-side, and the invalid language would only be caught by the backend PR. Worth confirming this is acceptable (relying on backend as the source of truth) or tightening the client check to also flag "unresolvable" languages.

### Code quality / reuse
- `resolveLanguageTitle` (opportunityDetailsSchema.ts) re-implements the same numeric-ID-vs-translated-title resolution pattern that already exists twice, nearly verbatim, as `toLangOptionItems` in both `NewOpportunity.tsx` and `OpportunityDetailsEdit.tsx`. Given this is now a third copy, consider extracting a single shared resolver (e.g. into `formatters.ts`) that both the submit-time mapping and the schema validation call, to avoid the two implementations drifting.
- `getMainCommunicationLanguageOptions` filters by `l.title.toLowerCase() in ["german", "english"]`. `ApiLanguageOption` already exposes an `isoCode` field — matching on `isoCode` (`"de"`/`"en"`) would be more robust than a hardcoded English-string match if the language master list's `title` ever changes casing/wording.

### Test coverage
- No unit tests were added for the new `superRefine` branching (none / German-only / German+English / invalid) or for `resolveLanguageTitle`'s two resolution paths (numeric ID vs. translated title). The PR's test plan relies entirely on manual verification. Given the branching logic and the edge case above, a few schema-level unit tests would be valuable and cheap to add.

### Minor
- `mainCommunicationInvalid` copy ("Main communication must be German, or German and English") reads a little awkwardly — consider "...must be German only, or German and English" for clarity, in both `en`/`de` strings.

Nothing blocking; the core restriction + validation logic looks correct for the common path. Main things I'd want addressed before merge are the legacy-data validation gap and, ideally, de-duplicating the language-resolution logic.

need4deed (approved):
Reviewed — legacy/unresolved-language gap, de-dup, and isoCode-matching feedback all addressed correctly. Resolved a substantive merge conflict against develop (event-date/time form restructure) and verified with typecheck/lint/production build. Safe to merge.

PR #830, opened 2026-07-20
feat: allow agent to change status of their own opportunity

arturasmckwcz:
## Summary
- `OpportunityHeader.tsx`'s "Change status" button was gated behind `isAuthorized` (admin/coordinator only), so an agent had no way to mark their own opportunity inactive once it's no longer relevant.
- Now also shown when the viewer is an `AGENT` and the opportunity belongs to their own agent (`currentUser.agentId === opportunity.agent.id`), mirroring the ownership check the backend now enforces.
- Depends on the backend authorization change in need4deed-org/be#781 (`PATCH /opportunity/:id` now allows an AGENT to patch just `statusOpportunity` on their own agent's opportunity).

## Test plan
- [x] `yarn typecheck` passes
- [x] `yarn lint` passes (no new warnings/errors)
- [ ] Manual: log in as an agent, open one of that agent's own opportunities, confirm "Change status" button now appears and the dialog still updates status correctly (not verified live in this session — flagging for reviewer/QA since local backend networking wasn't available to exercise the full login + dashboard flow end-to-end)

Fixes #820


need4deed (approved):
Reviewed — gate is airtight (no alternate dialog trigger), payload shape matches be#781's field restriction exactly. No blocking issues.

Issue #810, opened 2026-07-16
feat: support multiple contact persons on NGO/agent profile (+ Add contact)


nadavosa:
## Context

The NGO/agent profile's Contact Details section only ever shows and edits a single "representative" contact (`AgentContactDetails.tsx`, backed by `ApiAgentProfileGet.representative: ApiRepresentativeGet`).

But the backend already supports multiple people per agent under the hood: `agent_person` is an m2m table (`AgentPerson` entity) linking many `Person` records to one `Agent`, each with its own role (`AgentRoleType`: social-worker, volunteer-coordinator, manager, project-coordinator, psychologist, project-staff, childcare-worker, other). Multiple rows already get created in practice — e.g. every time someone registers and joins an existing NGO (`POST /agent/register`), a new `AgentPerson` row is added for that org.

The backend's `Agent.representative` getter (`src/data/entity/opportunity/agent.entity.ts`) explicitly collapses this down to just **one** contact before it reaches the API — preferring `VOLUNTEER_COORDINATOR`, else the first row found — so any additional contacts an NGO actually has on file are silently invisible on their own profile.

## Required behavior

- The NGO profile's Contact Details section should show **all** contacts an agent has (all `AgentPerson` rows + their linked `Person` details), not just the one "representative."
- Add a **"+ Add contact"** action that lets an NGO add another contact person (name, role, email, phone, address) to their own profile.
- Existing contacts should remain editable/removable (scope removal/edit UX to whatever's reasonable — at minimum, adding should work; editing multiple already partially exists via the current single-contact edit form and can be generalized).

## What needs to change, by repo

**`sdk`** (must land first, per API-contract-first policy):
- Add a `contacts: ApiRepresentativeGet[]` (or similarly named) field to `ApiAgentProfileGet`, replacing or supplementing the current singular `representative`.
- Add request/response shapes for creating a new contact (e.g. `ApiAgentContactPost`) and, if in scope, updating/removing one.

**`be`**:
- Stop collapsing `agent.agentPerson` down to a single `representative` in `dto-agent.ts` — serialize the full list instead.
- Add an endpoint to create a new `AgentPerson` (+ `Person`) for an existing agent from the dashboard (today, `AgentPerson` rows are only ever created via the registration/join flow in `register.routes.ts`, not from an already-registered NGO's own profile).
- Decide on authorization: who can add/edit/remove contacts for their own agent (likely any active member, mirroring the existing membership model).

**`fe`**:
[...]

PR #844, opened 2026-07-21
feat: show all agent contacts + Add contact action on NGO profile

arturasmckwcz:
## Summary
Third and final repo in the multi-repo effort for #810 (support multiple contact persons on NGO/agent profile), following:
- need4deed-sdk 0.0.130 (`contacts: ApiAgentMembership[]` + `ApiAgentContactPost`)
- need4deed-org/be#782 (`GET /agent/:id` now serializes `contacts`; new `POST /agent/:id/contact`) — **not yet merged/deployed**, so contact creation won't succeed end-to-end until it lands.

Changes:
- `AgentContactDetailsDisplay` now lists every `AgentPerson` membership from `agent.contacts` (excluding the one already shown/editable above via the existing representative form) under an "Other contacts" heading, instead of only ever showing the single collapsed representative.
- New **"+ Add contact"** action opens `AddContactDialog` — a form (first/last name, role, email, phone, street, postcode) that posts via a new `useCreateAgentContact` hook to `POST /agent/:id/contact`, invalidating the `["agent", id]` query on success so the list refreshes.
- Editing the primary representative is unchanged — the existing form/hook (`useUpdateAgentContact`, `AgentContactDetailsEdit`) is untouched.
- Editing/removing additional contacts individually is **not** in this PR — per the issue's own "at minimum, adding should work; editing multiple ... can be generalized [later]" — left as a follow-up.
- Bumped `need4deed-sdk` to `0.0.130`.

## Not in this PR
- Individual edit/remove for non-primary contacts (follow-up).

## Test plan
- [x] `yarn typecheck`
- [x] Dev server boots cleanly with these changes, no console errors (verified via headless browser against the public login redirect)
- [ ] **Not verified end-to-end**: the agent profile page requires an authenticated AGENT/COORDINATOR session, which I don't have test credentials for in this environment, and the live backend this repo's `.env` points at doesn't have be#782 deployed yet (so `agent.contacts` isn't in its `GET /agent/:id` response and `POST /agent/:id/contact` would 404). Please manually verify the full flow (list rendering, add-contact dialog, success/error toasts) once be#782 is merged/deployed, ideally against a local `docker compose` stack running both branches.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "date-fns": "^4.1.0",
     "email-validator": "^2.0.4",
     "i18next": "^25.3.2",
-    "need4deed-sdk": "0.0.128",
+    "need4deed-sdk": "0.0.130",
     "next": "15.3.8",
     "react": "^19.0.0",
     "react-day-picker": "^9.13.0",
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -1116,6 +1116,17 @@
         "cancel": "Abbrechen",
         "saveChanges": "Änderungen speichern",
         "edit": "Bearbeiten",
+        "additionalContacts": {
+          "title": "Weitere Kontakte"
+        },
+        "addContact": {
+          "button": "Kontakt hinzufügen",
+          "title": "Kontakt hinzufügen",
+          "submit": "Hinzufügen",
+          "addressStreet": "Straße",
+          "addressPostcode": "Postleitzahl",
+          "success": "Kontakt erfolgreich hinzugefügt"
+        },
         "validation": {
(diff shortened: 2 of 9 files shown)


need4deed (approved):
Approving after thorough review + live E2E verification against the real merged be#782/#783 code: positive add-contact flow works end-to-end (201, toast, list refresh via query invalidation), empty-state hiding for the 'other contacts' section is correct, and the one gap I flagged (Add contact button not gated by hasEditingRights) isn't actually reachable in practice — confirmed a non-member gets redirected before the profile page renders. Follow-up i18n fix for a pre-existing EditableField bug found during this review shipped separately as #845.

PR #845, opened 2026-07-21
fix: translate "Select option(s)..." placeholder in EditableField

arturasmckwcz:
## Summary
`EditableField`'s checkbox-list/radio-list dropdown placeholder ("Select option..." / "Select options...") was hardcoded in English regardless of the active locale — confirmed live: on the German UI it showed up untranslated (e.g. in the role picker of fe#844's new "Add contact" dialog). This is a pre-existing bug in the shared component itself, not something #844 introduced — every checkbox-list/radio-list field across the app was affected (any Dashboard/Profile edit form using that field type).

## Fix
- Added `dashboard.common.selectOption` / `dashboard.common.selectOptions` (alongside the existing `dashboard.common.loading` in that same generic namespace) to both `en` and `de` translation files.
- `EditableField.tsx` now uses `useTranslation()` and these keys instead of the two hardcoded literals.

## Test plan
- [x] `yarn typecheck`
- [x] Verified live: this exact placeholder was showing untranslated on `/de/...` before the fix (screenshot taken while reviewing fe#844); confirmed the translation keys resolve correctly for both locales.

### DIFF
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -341,6 +341,8 @@
   "dashboard": {
     "common": {
       "loading": "Wird geladen...",
+      "selectOption": "Option auswählen...",
+      "selectOptions": "Optionen auswählen...",
       "datePrefix": {
         "appliedOn": "Beworben am",
         "matchedOn": "Zugeordnet am",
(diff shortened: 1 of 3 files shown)


need4deed (approved):
Approving — small, well-scoped i18n fix for a pre-existing bug in the shared EditableField component, confirmed live during the fe#844 review.

PR #850, opened 2026-07-21
fix: enforce a consistent 8-day accompanying appointment lead time

arturasmckwcz:
## Summary
Three entry points for an Accompanying opportunity's appointment date each enforced a different rule:
- **New Opportunity creation**: no lead-time restriction at all — a coordinator/agent could pick tomorrow, or today.
- **Editing an existing opportunity's Accompanying Details** and **the "Change type → Accompanying" dialog**: both used `getMinAppointmentDate()`, which counted 7 *weekdays* ahead (skipping Sat/Sun) instead of a flat calendar-day offset — landing anywhere from 9 to 11 calendar days out depending on today's weekday, and shifting week to week.

None matched the legacy form's actual rule: exactly 8 calendar days from now, every time.

## Fix
- `getMinAppointmentDate()` now returns `today + 8 calendar days` (flat offset, no weekday counting).
- `NewOpportunity.tsx` now calls the same shared helper instead of a bare `new Date()`, so all three entry points enforce the identical rule from one place.

Verified downstream: `minAppointmentDate` feeds directly into each form's `DatePicker`'s `minDate` prop — a real UI constraint, not just a display hint — and there's no separate/duplicated day-offset check elsewhere (e.g. in the zod schema) that also needed updating.

## Test plan
- [x] `yarn typecheck`

### DIFF
--- a/src/components/Dashboard/NewOpportunity/NewOpportunity.tsx
+++ b/src/components/Dashboard/NewOpportunity/NewOpportunity.tsx
@@ -16,6 +16,7 @@ import {
 } from "@/components/Dashboard/Profile/sections/OpportunityDetails/opportunityDetailsSchema";
 import { resolveFormLanguageToOption } from "@/components/Dashboard/Profile/sections/OpportunityDetails/formatters";
 import { AccompanyingDetailsEdit } from "@/components/Dashboard/Profile/sections/AccompanyingDetails/AccompanyingDetailsEdit";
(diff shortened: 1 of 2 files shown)


need4deed (approved):
Approving — verified the date math by hand (correctly lands on exactly day+8), confirmed via grep that all three entry points now delegate to the single fixed helper with no missed call sites, and confirmed the value feeds a real DatePicker.minDate constraint downstream.

PR #853, opened 2026-07-22
Unify agent contact rows: consistent display + editable for every contact

arturasmckwcz:
## Summary

Closes #848.

Every contact on an agent's profile — the primary representative and every additional one — now renders and edits identically: one row per person (name, role, email, mobile, landline, address), each with its own **Edit** button. Previously the primary was collapsed into a section-wide edit form while additional contacts were read-only summary lines with no way to edit them at all.

"+ Add contact" and per-row "Edit" now share one dialog and one field set, per the agreed design (**keep address, add the missing fields to both**): First Name*, Middle Name, Last Name*, Role*, Email, Mobile, Landline, Street, Postcode.

This requires the new `PATCH /agent/:id/contact/:membershipId` endpoint — need4deed-org/be#784 (published as `need4deed-sdk@0.0.132`) — since `PATCH /person/:id` is strictly self-only and had no path for an agent to edit another member's contact.

## Changes

- `ContactRow` — shared uniform display row + inline Edit button, used for every contact (primary and additional alike).
- `ContactFormDialog` (replaces `AddContactDialog`) — generalized to handle both create and edit via an optional `contact` prop.
- `contactFormSchema` (replaces `addContactSchema` + `agentContactDetailsSchema`) — one unified zod schema for both flows.
- `useUpdateAgentContactMembership` — new PATCH hook, mirrors `useCreateAgentContact`.
- `AgentContactDetails` — rewritten to map `agent.contacts` directly to rows; no more special-casing the primary. `ref`/`onEditingChange` kept as no-ops for API compatibility with the shared `ContactDetails` switchboard (volunteer/opportunity sections still use the old section-wide-edit pattern).
- `useAgentProfileSections` — removed the section-wide "Edit" header button for Contact Details (editing is now per-row) and the now-unused ref/state wiring.
- Removed: `AgentContactDetailsDisplay`, `AgentContactDetailsEdit`, `agentContactDetailsSchema`, `AddContactDialog`, `addContactSchema`, `useUpdateAgentContact`.
- Added `editContact.*` translation keys (en/de) alongside the existing `addContact.*` keys.
- Bumped `need4deed-sdk` to `0.0.132`.

## Not in this PR

be#784 (the PATCH endpoint) is open but not yet merged. This PR targets its contract regardless, per the established sdk → be → fe pattern used for #810.

## Test plan

Live-verified against a tunneled copy of the dev DB (be run locally on the `848-agent-contact-crud` branch, `RUN_MIGRATIONS=0`):

- [x] Contact list renders one uniform row per contact, including the primary, with role/email/mobile/landline/address inline
- [x] Section header no longer has an "Edit" button; each row has its own
- [x] Edit dialog opens pre-filled with a contact's real data (primary and a non-primary contact)
- [x] Add dialog has the identical field set to Edit (address included)
- [x] Full submit round-trip through the actual UI: edited a contact's landline, saw the success toast and the row update in place via cache invalidation
- [x] `yarn typecheck` and `yarn lint` clean (only the 3 pre-existing unrelated warnings)
- [x] All test data cleaned up afterward; dev DB confirmed restored to its exact prior state


need4deed (approved):
Reviewed (see review comment above). Every contact now renders and edits identically via ContactRow/ContactFormDialog, backed by the merged be#784 endpoint. Live-verified end-to-end against the real dev DB, CI green, no lint/type issues.

Approving.

PR #864, opened 2026-07-23
feat: coordinator-only delete (opportunity/volunteer/NGO) and transfer-to-NGO

nadavosa:
## Summary
Closes [#863](https://github.com/need4deed-org/fe/issues/863). Backend companion: [need4deed-org/be#788](https://github.com/need4deed-org/be/pull/788).

Adds a "Danger zone" section at the very bottom of all three profile pages (Opportunity, Volunteer, NGO), below Comments, visible only to COORDINATOR/ADMIN — reuses the exact same `isAuthorized` gate already used for the Comments section itself, so no new authorization concept is introduced.

- **Delete** button on all three profile types. Opens the existing shared `ConfirmationDialog` (already used by `SuggestDialog`/`DeleteCommentDialog` — no new dialog component). On confirm, calls the new `DELETE /opportunity/:id` / `DELETE /volunteer/:id` (be#788) or the existing `DELETE /agent/:id`, then redirects to that entity's list page.
- **Transfer to another NGO** button, Opportunity profiles only, to the left of Delete. Clicking it navigates to the Agents list with `?transferOpportunity=<id>`. On that list (in both card and table view), clicking an NGO — instead of opening its profile — opens a confirmation dialog right there; on confirm it calls the *existing* `PATCH /opportunity/:id` with `{ agent: { id } }` (no new backend endpoint — this relink path already existed) and redirects back to the opportunity's profile, now showing the new NGO.

## Reuse before create
- `ConfirmationDialog` (`Profile/sections/shared/ConfirmationDialog.tsx`) reused as-is for every delete/transfer confirmation — no new dialog component.
- New hooks (`useDeleteOpportunity`, `useDeleteVolunteer`, `useDeleteAgent`, `useTransferOpportunityToAgent`) are thin `useMutationQuery` wrappers, following `useDeleteComment`'s exact pattern (see `mutate(undefined)` call convention).
- The Agents-list "select instead of navigate" mechanism is a small `onSelect?` prop threaded through `AgentCard`/`AgentTableRow` → `AgentCardList`/`AgentTableList` → `AgentListController`, mirroring the existing `opportunityId` prop already threaded through the equivalent Volunteers-list chain for volunteer-matching. When absent (normal browsing), both components render exactly as before (`as={Link}`); only when transferring does the card/row switch to an `onClick` that opens the dialog instead of navigating.
- One new small component, `DangerZoneButtonRow` (button row layout only — Delete always, Transfer only when passed), and one new `IconName.Trash` mapped to phosphor's `TrashIcon`.

## i18n
Added `dangerZone` keys under `opportunityProfile`, `volunteerProfile`, and `agentProfile`, and `transferMode` under `agents`, in both `en` and `de` — validated both JSON files still parse and the new keys land under the correct parent (there's an easy trap here: `volunteerProfile` and `documentSection` share several identical-looking nested blocks, so a first pass landed a key in the wrong parent — caught and fixed before committing).

## Test plan
- [x] `yarn typecheck`
- [x] `yarn lint` (clean on all touched files; 3 pre-existing warnings elsewhere are unrelated)
- [ ] No live backend/Postgres available in this environment, so this hasn't been exercised in a browser against real data — please verify the golden path (delete confirmation → redirect; transfer → pick NGO → confirm → land back on the now-transferred opportunity) before merging.


ivannissimrch (commented):
Took this over per standup. Tested the full flow locally (transfer on develop BE; delete with the 787-delete-opportunity-volunteer BE branch pulled in). Found and fixed: all four new mutation hooks useTransferOpportunityToAgent, useDelete{Opportunity,Volunteer,Agent} were missing queryKeyToInvalidate, so the UI showed stale data after every transfer/delete until a manual refresh. Added invalidation keys verified the item now vanishes from the list / the NGO section updates without a refresh. (commit 9caee2b)
Verified working: delete all 3 types and  transfer.  Blocker: delete depends on be#788  404s on develop until that merges.
Notes: stale contact after transfer is intended per the spec acceptance criteria. Transfer UX works; a picker could be a nice future add if the NGO list gets 
 long.

PR #875, opened 2026-07-25
Render multiple agent-address match candidates

arturasmckwcz:
## Description

**Updated** — the original single/multi-banner approach was redone per direct feedback after using the actual picker. This now matches the redesigned backend (need4deed-org/be#801: `GET /agent/register/search` drops postcode and returns every matching candidate, prefix-matched on `address.street` or contains-matched on `title`).

- `useAgentAddressLookup`: dropped `addressPostcode` entirely (param, debounce, query string, gating). Search fires once `street` is 3+ characters, full stop.
- `ProfileCompletion`: always renders as a plain list — one header line at the top, each candidate as a clickable row (the row **is** the CTA, no separate per-row button), one "No, enter manually" action at the bottom.
- Visual hierarchy fix: the header is plain descriptive copy (new `MatchListCopy`, matches this step's existing description styling) instead of a colored banner that read as a CTA; the actual CTA ("No, enter manually") is now styled `$primary` so it reads as the deliberate action it is.
- Pluralization fix: replaced a hardcoded-plural string with i18next's `_one`/`_other` key suffixes (`matchesFound_one`/`matchesFound_other`, en + de) — verified against a real i18next instance for count=1 and count=3 in both languages, no plural logic in the component itself.
- UX guard: the "Next" button is now disabled while the candidate list is still open, so a registrant can't move on without either picking an org or explicitly dismissing.

No new UI library or styling primitives — reuses/extends the existing `MatchBanner`/`MatchActions`/`SmallButton` family plus one new `MatchRow` (clickable row) and `MatchListCopy` (plain copy line). There's no Figma spec for this state, so this is a functional, low-risk layout rather than a designed one.

## Related Issues
Closes #874

## Changes
- `useAgentAddressLookup.ts`: dropped postcode entirely
- `ProfileCompletion.tsx`: list-only rendering, restyled header/CTA, disabled "Next" while list is open
- `styled.ts`: new `MatchRow`, `MatchListCopy`
- `public/locales/{en,de}/translations.json`: `matchesFound_one`/`matchesFound_other`; removed now-dead `matchFound`/`useThisOrg`/`completion.matched` keys

## Screenshots / Demos
Not captured — full picker QA needs a running backend with seeded matching agents and a real verify token. Verified instead: `yarn typecheck`/`yarn lint` clean, translation files parse, i18next plural resolution checked directly against a live i18next instance (all 4 combinations), and the address-completion page renders without error.

## Checklist
- [x] Tests added/updated — n/a, no test framework configured in this repo
- [ ] Documentation updated
- [x] CI passes (pending)

### DIFF
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -2039,6 +2039,7 @@
       "title": "Konto-Einrichtung abschließen",
       "subtitle": "Erzählen Sie uns von Ihrer Organisation.",
       "matchFound": "Wir haben eine Organisation gefunden, die zu dieser Adresse passt: {{name}}. Möchten Sie diese verwenden?",
+      "matchesFound": "Wir haben {{count}} Organisationen gefunden, die zu dieser Adresse passen. Wählen Sie Ihre aus oder erstellen Sie eine neue.",
       "useThisOrg": "Ja, diese verwenden",
       "skip": "Nein, manuell eingeben",
       "matched": "Organisation gefunden: {{name}}",
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -2038,11 +2038,9 @@
     "completion": {
       "title": "Konto-Einrichtung abschließen",
       "subtitle": "Erzählen Sie uns von Ihrer Organisation.",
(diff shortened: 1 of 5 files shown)


need4deed (approved):
Reviewed. Restyling and pluralization fixes confirmed correct; combined effect with be#802's broader agent matching is acknowledged and mitigated by the existing join-confirmation step. LGTM.

PR #893, opened 2026-07-30
fix broken NGO-details editing, move Contact Person section up, and make it a contact picker 

ivannissimrch:
## Description
The RAC section was showing edit fields for address and district that the backend never persisted. Those are gone it ony shows name and edits name.

Contact Details moved up under the NGO section, and its phone/email are now read-only, with the name becoming a dropdown of the NGO's registered contacts.

  - Removed address/district from the RAC section schema, display, edit form, and PATCH payload. Also drops the AgentUpdateData.district type drift noted in the issue, since the field is gone
  - Contact Details now renders directly below the RAC section
  - Contact phone, email, and ways-to-contact are display-only; 
  - Contact name is a radio-list dropdown of the linked NGO's contacts, sending contact.id instead of a free-text name
  - New useAgentContactOptions hook for the agent-contacts fetch
  - Removed the orphaned i18n keys.
 
### Question 
I removed address/district from the RAC display as well as the edit form. Happy to restore them read-only if you'd rather keep them visible on the opportunity page.

Heads up: the picker sends { contact: { id } }, and right now the backend ignores it. You can pick a name and hit save, but the contact won't actually change until the relink lands. Waiting on the SDK/BE issue for that.

## Related Issues
Closes #869

## Screenshots / Demos

https://github.com/user-attachments/assets/e3b6d641-16ce-4fcc-bd05-34f542ed7ac4

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes


arturasmckwcz (approved):
now it's good thanks a lot!!!

PR #898, opened 2026-08-03
feat: make the asset CDN origin a build input

theodore-evans:
## Why

The CloudFront distribution is hardcoded in two places, so repointing the frontend at a different asset host means editing source: `next.config.ts` holds the hostname for the image allowlist, and the `Dockerfile` bakes the asset URLs as fixed `ENV`s. `src/config/constants.ts` already reads the env vars correctly; only the build inputs were stuck.

## What changes

- `Dockerfile`: the two asset URLs become `ARG`s with the current CloudFront values as defaults, matching how `API_URL` is already handled.
- `next.config.ts`: the image-allowlist hostname is derived from `NEXT_PUBLIC_CLOUDFRONT_URL` rather than repeated, with the current value as fallback. A malformed URL fails the build.
- A set-but-empty value is rejected, because `next.config` would keep the default host while app code emitted root-relative paths: every CDN image 404s with a green build.
- `docker-compose.yml` passes the args through, since `NEXT_PUBLIC_*` is inlined at build time and `env_file` arrives too late to have any effect. The file's indentation is normalised to two spaces in passing, which is most of the diff noise there.

Env var names are unchanged: renaming to something provider-neutral touches every consumer plus the deploy secrets, and is cosmetic relative to the migration.

## Verification

`yarn build` green with defaults; with the override, the new hostname is baked into the server chunks; malformed and empty values both fail the build. Lint and typecheck clean.

## Note for the maintainers

This repo has no image build/publish workflow (only lint and build-check), and `ghcr.io/need4deed-org/fe:develop` returns 404 from the registry. The frontend cannot be deployed until that pipeline exists; worth raising separately.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,8 @@ FROM node:22-alpine AS builder
 WORKDIR /app
 
 ARG API_URL=http://localhost:8000
+ARG NEXT_PUBLIC_CLOUDFRONT_URL=https://d2nwrdddg8skub.cloudfront.net/images
+ARG NEXT_PUBLIC_CLOUDFRONT_DATA_URL=https://d2nwrdddg8skub.cloudfront.net/data
 
 COPY package.json yarn.lock* ./
 RUN yarn install --frozen-lockfile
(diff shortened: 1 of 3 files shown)


arturasmckwcz (approved):
LGTM thank you!

PR #912, opened 2026-08-07
885: fix, adds autocomplete operator edit in AgentOrganisationDetails

### LINKED ISSUE #885: NGO profile organisaional details section bugs
### Summary

several issues not saving and no dropdown 

### Steps to Reproduce

in editing mode: 
1. NGOs complain they cannot edit organisation details (maybe we need to make feilds about,website optionals)
2. operator / träger **DOES NOT SAVE** and it should be a drop down made of all the domains (ib/drk/albetros..)

### Expected Behavior

saveable 

### Actual Behavior

does not save

### Environment

_No response_

DarrellRoberts:
## Description
On edit for `agent` profile for `OrganisationDetails` this allows a new autocomplete `EditableField` to field `operator`.

## Related Issues
Closes #885 

## Changes
- Adds new `EditableField` type: `autocomplete`

## Screenshots / Demos

https://github.com/user-attachments/assets/1914050b-5958-49eb-ba07-c16018a00f3f

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes


src/components/Dashboard/common/statusMaps.ts
```diff
@@ -15,6 +15,7 @@ import { AgentEngagementStatusType, AgentTrustType, AgentVolunteerSearchType } f
 
 export type StatusValue = AgentEngagementStatusType | AgentVolunteerSearchType | AgentTrustType;
 
+// @ts-expect-error - TODO: Add INCONTACT and TRIED_TO_CONTACT types
```
DarrellRoberts, line 18: will implement when I do to [Issue 796](https://github.com/need4deed-org/fe/issues/796)

src/components/Dashboard/common/statusMaps.ts
```diff
@@ -30,6 +31,7 @@ export const statusColorMap: Record<StatusValue, string> = {
 
 type IconComponent = React.ComponentType<{ size?: number; color?: string }>;
 
+// @ts-expect-error - TODO: Add INCONTACT and TRIED_TO_CONTACT types
```
DarrellRoberts, line 34: will implement when I do to [Issue 796](https://github.com/need4deed-org/fe/issues/796)

src/components/Dashboard/Profile/sections/ProfileHeader/agent/constants.ts
```diff
@@ -2,12 +2,14 @@ import { EMPTY_PLACEHOLDER_VALUE } from "@/config/constants";
...
 import { AgentEngagementStatus, AgentTrustLevel, AgentVolunteerSearch, ApiAgentProfileGet } from "../../../types/agent";
 
-export const createEngagementStatusLabelMap = (t: TFunction): Record<AgentEngagementStatus, string> => ({
-  [AgentEngagementStatus.NEW]: t("dashboard.agentProfile.status.engagement.new"),
-  [AgentEngagementStatus.ACTIVE]: t("dashboard.agentProfile.status.engagement.active"),
-  [AgentEngagementStatus.UNRESPONSIVE]: t("dashboard.agentProfile.status.engagement.unresponsive"),
-  [AgentEngagementStatus.INACTIVE]: t("dashboard.agentProfile.status.engagement.inactive"),
-});
+export const createEngagementStatusLabelMap = (t: TFunction): Record<AgentEngagementStatus, string> =>
+  // @ts-expect-error TODO - Add INCONTACT and TRIED_TO_CONTACT types
```
DarrellRoberts, line 6: will implement when I do to [Issue 796](https://github.com/need4deed-org/fe/issues/796)

package.json
```diff
@@ -21,7 +21,7 @@
     "date-fns": "^4.1.0",
     "email-validator": "^2.0.4",
     "i18next": "^25.3.2",
-    "need4deed-sdk": "0.0.139",
+    "need4deed-sdk": "0.0.141",
```
DarrellRoberts, line 24: neccessary for `ApiOrganizationGetList[]` type above

nadavosa:
## Follow-up

Thanks for the detailed responses — the exact-match and typing-UX decisions make sense given the "aus" vs "ausbildung-ota.de" example, and the uncontrolled-component fix looks correct now.

One thing still not quite right in the case-insensitivity fix:
```ts
return operatorOptions.filter((op) => op.includes(value.toLowerCase()));
```
Only `value` gets lowercased — `op` (the actual organization title, e.g. "Caritas") doesn't. So `"Caritas".includes("car")` is still `false`, and typing the organization's own name in natural lowercase still won't surface it in the list. Needs `op.toLowerCase().includes(value.toLowerCase())`.

Saw your note about #923 for the `@ts-expect-error` cleanup — left a comment there too, since the two `statusColorMap`/`statusIconMap` files ended up with inconsistent colors/icons for the new statuses (one file makes INCONTACT/TRIED_TO_CONTACT/NEW all visually identical).

arturasmckwcz (commented):
Automated review findings (4 confirmed issues).

src/components/Dashboard/Profile/sections/OrganisationDetails/organisationDetailsSchema.ts
```diff
@@ -26,7 +26,12 @@ export const createOrganisationDetailsSchema = (t: (key: string) => string) => {
     // Stores the translated title (like `district` elsewhere), resolved
     // back to an id at submit time via AgentType/Service option mappings.
     organizationType: z.string().min(1, required),
-    operator: z.string().min(1, required),
+    operator: z
+      .string()
+      .min(3, t(`${i18nPrefix}.operatorRequired`))
+      .refine((val) => validOperators.includes(val), {
```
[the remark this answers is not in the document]
DarrellRoberts: do we currently have NGOs in production with stored `operators` which don't fall into the list? If so, wouldn't we want to phase this out? Or is the autocomplete list just a suggestion? 

this could be a confusion on my part

src/components/Dashboard/Profile/sections/OrganisationDetails/OrganisationDetailsEdit.tsx
```diff
@@ -31,6 +34,16 @@ export const OrganisationDetailsEdit = ({
     formState: { errors, isDirty, isValid },
   } = useFormContext<OrganisationDetailsFormData>();
 
+  const displayOperators = useCallback(
+    (value: string) => {
+      if (value?.length >= 3) {
+        return operatorOptions.filter((op) => op.includes(value.toLowerCase()));
```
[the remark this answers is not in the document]
DarrellRoberts: good point and relates to @nadavosa comment. I've added an extra `toLowerCase()` for the `operator` (or `op`).

src/components/EditableField/EditableField.tsx
```diff
@@ -559,6 +559,65 @@ export const EditableField = forwardRef(function EditableField<T extends string
             )}
           </DropdownWrapper>
         )}
+
+        {type === "autocomplete" && (
+          <DropdownWrapper ref={wrapperRef}>
+            <DropdownButton $hasError={!!errorMessage} onClick={() => setOpen((o) => !o)}>
```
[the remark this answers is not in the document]
DarrellRoberts: good point and I found it was easier to just always set the `setOpen` to `true` whenever the user clicks on the input field. This means when the user clicks outside of the input, it closes the `autocomplete` but whenever they click on the input, e.g. when inserting a letter anywhere within the value, it opens the `autocomplete`.

Or in other words, the `autocomplete` will always be open so long as there are matches & the user-focus is on the text-input

src/components/EditableField/EditableField.tsx
```diff
@@ -559,6 +559,65 @@ export const EditableField = forwardRef(function EditableField<T extends string
             )}
           </DropdownWrapper>
         )}
+
+        {type === "autocomplete" && (
```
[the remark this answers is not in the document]
DarrellRoberts: originally I wanted to reuse the checkbox-list/radio-list but I found it was easier to create a new type, as otherwise you are complicating this checkbox/radio-lists with extra props and conditions (also risks accidentally mutating their existing behaviour).

Also now they have different behaviours and slightly different appearences (now no chevron arrow for `autocomplete` as it's no longer needed)

I think also if there is a future change to the checkbox-list/radio-list it shouldn't neccessarily mean that the same change should apply to the autocomplete?

DarrellRoberts:
thanks @nadavosa and @arturasmckwcz , I've implemented some changes.

- Added additional `toLowerCase()` for `operator` in operator filter function
- Changed the `onClick` handler for `autocomplete, EditableField` so that the autoselect menu always open when user clicks on input

@arturasmckwcz for your points I've replied directly

src/components/Dashboard/Profile/sections/OrganisationDetails/organisationDetailsSchema.ts
```diff
@@ -26,7 +26,12 @@ export const createOrganisationDetailsSchema = (t: (key: string) => string) => {
     // Stores the translated title (like `district` elsewhere), resolved
     // back to an id at submit time via AgentType/Service option mappings.
     organizationType: z.string().min(1, required),
-    operator: z.string().min(1, required),
+    operator: z
```
arturasmckwcz: **Correctness:** The `operator` field's Zod validation requires an exact match against `validOperators`, sourced from a separately-cached `useGetOrganization()` query, while the pre-filled `operator` value comes from the independently-cached agent query — the two caches can diverge.

**Failure scenario:** The organizations list is fetched once (5 min staleTime) per tab. Another coordinator renames the linked organization. The agent profile then loads a fresh `operator` reflecting the new title, but the stale cached organizations list still has the old title, so `validOperators.includes(operator)` is false, `isValid` stays false, and the whole Organisation Details save button is blocked — even for edits to unrelated fields — with only a cryptic error under the untouched operator field.

src/components/EditableField/EditableField.tsx
```diff
@@ -559,6 +559,68 @@ export const EditableField = forwardRef(function EditableField<T extends string
...
```
arturasmckwcz: **Simplification:** The new `type === "autocomplete"` branch fully duplicates the DropdownWrapper/DropdownButton/DropdownList/OptionRow scaffold already used by the checkbox-list/radio-list branch instead of extending it, and has already silently diverged from it in two ways (see follow-up comment below).

**Failure scenario:** radio-list's button toggles open/closed, autocomplete's only opens and never closes on click; radio-list's list renders on `open` alone while autocomplete additionally gates on `options.length > 0`. Any future fix to shared dropdown behavior (keyboard nav, a11y, styling) has to be applied to both blocks or they drift further, as already happened here.

arturasmckwcz (commented):

src/components/EditableField/EditableField.tsx
```diff
@@ -559,6 +559,68 @@ export const EditableField = forwardRef(function EditableField<T extends string
...
+          <DropdownWrapper ref={wrapperRef}>
+            <DropdownButton
+              $hasError={!!errorMessage}
+              onClick={() => {
+                if (value) setOpen(true);
```
arturasmckwcz: **Correctness:** `DropdownButton`'s onClick reads the external `value` prop instead of `localValue`, and the adjacent `ClearButton` doesn't call `e.stopPropagation()`, so clicking Clear also bubbles into the still-truthy `value` closure and forces `open` back to `true` right after clearing.

**Failure scenario:** User clicks the X to clear the operator field: ClearButton's handler sets value to "", then the same click event bubbles to DropdownButton, whose closure still has the pre-clear (truthy) value, so `setOpen(true)` fires. Currently masked because `options.length > 0` also gates the list and becomes 0 once cleared, but it leaves `open` stuck true — a latent bug that surfaces if empty-value gating on options is ever relaxed.

src/hooks/useGetOrganization.ts
```diff
@@ -0,0 +1,18 @@
+import { ApiOrganizationGetList } from "need4deed-sdk";
+import { useGetQuery } from "./useGetQuery";
+import { apiPathOrganization, cacheTTL } from "@/config/constants";
+
+export const useGetOrganization = () => {
```
arturasmckwcz: **Efficiency:** `useGetOrganization` omits `addLang: false`, unlike every other comparable hook fetching non-translated reference data, so the UI language becomes part of the request and the react-query cache key for data that doesn't depend on it.

**Failure scenario:** Backend `GET /organization/` never reads `request.query.language`, and organization titles are explicitly documented as untranslated. Toggling UI language between en/de causes react-query to treat the two languages as separate cache entries, triggering a redundant refetch and duplicate cached copies of identical data.

src/components/EditableField/EditableField.tsx
```diff
@@ -559,6 +559,68 @@ export const EditableField = forwardRef(function EditableField<T extends string
...
+              }}
+            >
+              <InputWrapper>
+                <input
```
arturasmckwcz: **Correctness:** The new autocomplete `<input>` wires neither `onBlur` nor `onKeyDown`, unlike every other edit-mode input type (text/textarea/number) in this shared component, which all support submit-on-blur and Enter-to-submit.

**Failure scenario:** Dormant today since no current `EditableField` caller passes a `submit` prop with `type="autocomplete"`, but a future caller adding it following the existing text-input pattern will find blur-to-save and Enter-to-save silently do nothing, and Enter won't even `preventDefault()` if the field is ever placed inside a real `<form>`.

arturasmckwcz:
@DarrellRoberts tagging you just in case you've missed the last review findings

DarrellRoberts:
@arturasmckwcz , have pushed my changes

src/components/Dashboard/Profile/sections/OrganisationDetails/OrganisationDetails.tsx
```diff
@@ -96,6 +99,7 @@ export const OrganisationDetails = forwardRef<EditableSectionRef, Props>(functio
     const serviceIds = values.services
       .map((title) => serviceMapping.titleToId[title])
       .filter((id): id is number => id !== undefined);
+    const organizationId = organizationMapping.titleToId[values.operator];
```
arturasmckwcz: **[CONFIRMED] Operator save silently no-ops when the typed text doesn't exactly match an organization title.**

If the user types an operator name but doesn't click a dropdown row (case/whitespace differs, or they ignore the list and hit Save), `organizationMapping.titleToId[values.operator]` (exact-case lookup) returns `undefined`, so `...(organizationId !== undefined && { organizationId })` (line 113) omits `organizationId` from the PATCH entirely. `ApiAgentPatch` has no raw `operator` string field, so nothing about the operator is sent at all. `organisationDetailsSchema.ts` only validates `operator: z.string().min(1, required)`, so the form is considered valid and Save succeeds anyway — `onSuccess` calls `reset(values); setIsEditing(false)`, and `OrganisationDetailsDisplay` renders the typed (unsaved) text via `watch()`. The user believes their change saved, but the backend operator/organizationId is untouched — silent data loss.

src/components/EditableField/EditableField.tsx
```diff
@@ -559,6 +559,72 @@ export const EditableField = forwardRef(function EditableField<T extends string
             )}
           </DropdownWrapper>
         )}
+
+        {type === "autocomplete" && (
+          <DropdownWrapper ref={wrapperRef}>
+            <DropdownButton
+              $hasError={!!errorMessage}
+              onClick={() => {
```
arturasmckwcz: **[CONFIRMED] The autocomplete dropdown closes on a plain click inside its own input, not just on option selection or outside click.**

With the dropdown open and a non-empty value, clicking inside the `<input>` to reposition the cursor or select text bubbles up (the input has no `onClick`/`stopPropagation`, unlike `ClearButton` which does call `e.stopPropagation()`) to the wrapping `DropdownButton`, whose handler is `onClick={() => { if (localValue) setOpen((o) => !o); }}` — toggling `open` from true to false and closing the suggestion list mid-edit.

public/locales/en/translations.json
```diff
@@ -1238,8 +1238,14 @@
         "validation": {
           "required": "This field is required",
           "websiteInvalid": "Website must start with http:// or https://",
+          "operatorRequired": "A minimum of 3 characters is needed",
```
arturasmckwcz: New i18n keys `operatorRequired` and `operatorInvalid` (and their `de` counterparts) are added but never referenced by any `t(...)` call outside the translation files. This copy looks like it was meant to surface the exact failure described in the operator-save finding, but it was never wired into the schema or the `EditableField` error path — translators now maintain strings nothing displays.

src/components/EditableField/EditableField.tsx
```diff
@@ -559,6 +559,72 @@ export const EditableField = forwardRef(function EditableField<T extends string
             )}
           </DropdownWrapper>
         )}
+
+        {type === "autocomplete" && (
```
arturasmckwcz: The new `autocomplete` branch duplicates the text-input+clear-button markup and the dropdown-list/option-row markup already implemented for the `text` and `checkbox-list`/`radio-list` branches (~65 lines re-authoring `InputWrapper`/`ClearButton`/`Boolean(localValue)` from the `text` branch and `DropdownWrapper`/`DropdownList`/`OptionRow` from `checkbox-list`/`radio-list`). A fix to one copy (e.g. the outside-click handling, or the clear-button stopPropagation flagged above) won't propagate to the other, and future edits must be made twice.

src/components/Dashboard/Profile/sections/OrganisationDetails/OrganisationDetailsEdit.tsx
```diff
@@ -123,14 +136,19 @@ export const OrganisationDetailsEdit = ({
...
           render={({ field }) => (
-            <EditableField
-              mode="edit"
-              type="text"
-              label={t(`${i18nPrefix}.operator`)}
-              value={field.value}
-              setValue={field.onChange}
-              errorMessage={errors.operator?.message}
-            />
+            <>
```
arturasmckwcz: The operator `Controller`'s render prop wraps a single `<EditableField>` in an unnecessary `<>...</>` fragment — no sibling element is added, and every other `Controller` in this file returns the `EditableField` directly.

src/components/Dashboard/Profile/sections/OrganisationDetails/OrganisationDetails.tsx
```diff
@@ -124,6 +129,7 @@ export const OrganisationDetails = forwardRef<EditableSectionRef, Props>(functio
             languagesForForm={languagesForForm}
             organizationTypeOptions={apiAgentTypes.map((agentType) => agentType.title)}
             servicesOptions={apiServices.map((service) => service.title)}
+            operatorOptions={organizations?.map((org) => org.title)}
```
arturasmckwcz: Redundant optional chaining: `const { data: organizations = [] } = useGetOrganization();` already guarantees `organizations` is never nullish, so `organizations?.map((org) => org.title)` carries a defensive `?.` that can never trigger.

arturasmckwcz (commented):
@DarrellRoberts Re-review findings (2 confirmed bugs, plus a few quality issues) — see inline comments below.

One more issue that couldn't be anchored inline because `organisationDetailsSchema.ts` isn't touched by this PR: operator validation isn't generalized into the schema the way the other closed-set fields (`services`, `organizationType`) are — `operator: z.string().min(1, required)` accepts any non-empty string. That's the root cause of the silent-save bug below: an invalid operator is only ever discovered inside `onSubmit`'s `organizationMapping.titleToId` lookup, with no mechanism to surface `errors.operator`. Consider a schema-level `.refine` against known organization titles instead.

need4deed:
@DarrellRoberts can you resolve conflicts here?

DarrellRoberts:
@arturasmckwcz , have pushed my changes. Sorry I'm getting confused from all the comments versus what is required

PR #923, opened 2026-08-11
796: adds two new engagement types to NGO engagement

DarrellRoberts:
## Description
Adds two new engagement types to NGO engagement:
- Tried to contact
- In contact

Currently blocked by `be` as new values not expected in `engagementStatus` validation

## Related Issues
Addresses #796 

## Changes

### Badges
image

image

#### EN
image

#### DE
image

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes


nadavosa:
## Review

Nice, thorough coverage of the two new engagement types across labels, colors, and icons — and confirms this properly closes the `@ts-expect-error` gaps I flagged on #912.

### Inconsistency between the two status-map files
`src/components/Dashboard/Profile/common/statusMaps.ts` gives the two new statuses distinct, sensible styling:
```ts
[AgentEngagementStatusType.INCONTACT]: "var(--color-green-500)",       // ChatsCircleIcon
[AgentEngagementStatusType.TRIED_TO_CONTACT]: "var(--color-grey-200)", // PhoneDisconnectIcon
```
But `src/components/Dashboard/common/statusMaps.ts` gives both the *same* color and icon as each other — and the same as `NEW`:
```ts
[AgentEngagementStatusType.NEW]: "var(--color-green-100)",           // SparkleIcon
[AgentEngagementStatusType.INCONTACT]: "var(--color-green-100)",     // SparkleIcon
[AgentEngagementStatusType.TRIED_TO_CONTACT]: "var(--color-green-100)", // SparkleIcon
```
Wherever this second file's maps are used (e.g. the Agents list/badges view, matching the "Badges" screenshot in the description), `New`, `In contact`, and `Tried to contact` would render visually identical — no way to tell them apart at a glance, which seems to defeat the point of adding two distinguishable statuses. Worth reusing the same color/icon choices from the `Profile/common` file here too (or intentionally choosing different-but-still-distinct ones), rather than leaving all three as `SparkleIcon`/`green-100`.

Nothing else stands out — translations read fine in both locales, and no other exhaustive `Record<AgentEngagementStatus...>` map was missed (checked all three usages in the repo).

DarrellRoberts:
thanks for checking @nadavosa & good spot. Have pushed my changes

src/components/Dashboard/common/statusMaps.ts
```diff
@@ -19,9 +19,9 @@ export const statusColorMap: Record<StatusValue, string> = {
   [AgentEngagementStatusType.ACTIVE]: "var(--color-green-100)",
   [AgentEngagementStatusType.UNRESPONSIVE]: "var(--color-grey-50)",
   [AgentEngagementStatusType.INACTIVE]: "var(--color-grey-50)",
-  [AgentEngagementStatusType.NEW]: "var(--color-green-100)",
-  [AgentEngagementStatusType.INCONTACT]: "var(--color-green-100)",
-  [AgentEngagementStatusType.TRIED_TO_CONTACT]: "var(--color-green-100)",
+  [AgentEngagementStatusType.NEW]: "var(--color-violet-100)",
+  [AgentEngagementStatusType.INCONTACT]: "var(--color-green-500)",
```
arturasmckwcz: **Correctness:** `statusIconMap` further down in this file was left unchanged for `INCONTACT`/`TRIED_TO_CONTACT` (still `SparkleIcon`, same as `NEW`), even though this diff gives them distinct colors here, and the sibling `Dashboard/Profile/common/statusMaps.ts` gives them distinct icons (`ChatsCircleIcon`/`PhoneDisconnectIcon`).

**Failure scenario:** Any future caller of `Dashboard/common/StatusBadge.tsx` for engagement status will render the new colors paired with the wrong, stale icon, inconsistent with how the same statuses render on the agent profile page. Currently dormant because `AgentCard.tsx` only feeds `volunteerSearch` (not `engagementStatus`) into this specific `StatusBadge`, so it won't visibly break until that column is wired up — but the next person to consume this map has no reason to suspect the inconsistency.

arturasmckwcz (commented):

**Simplification:** `src/components/Dashboard/common/statusMaps.ts` and `src/components/Dashboard/Profile/common/statusMaps.ts` are two hand-maintained, structurally parallel maps both keyed on overlapping `AgentEngagementStatusType` values with no shared source of truth — which is exactly what let this PR half-update one of them (see inline comment). A single canonical `AgentEngagementStatusType` → `{color, icon}` map imported by both consumers would make this class of drift structurally impossible.

DarrellRoberts:
thanks @arturasmckwcz , have pushed my changes

src/components/Dashboard/Profile/common/statusMaps.ts
```diff
@@ -95,10 +96,15 @@ export const statusColorMap: Record<StatusValue, string> = {
     [AgentTrustLevel.UNKNOWN, "var(--color-grey-50)"],
     [AgentTrustLevel.LOW, "var(--color-red-50)"],
     [AgentTrustLevel.HIGH, "var(--color-green-100)"],
-    [AgentEngagementStatusType.NEW, "var(--color-violet-100)"],
-    [AgentEngagementStatusType.ACTIVE, "var(--color-green-100)"],
-    [AgentEngagementStatusType.INACTIVE, "var(--color-grey-50)"],
-    [AgentEngagementStatusType.UNRESPONSIVE, "var(--color-grey-50)"],
+    [AgentEngagementStatusType.NEW, agentEngagementStatusConfig[AgentEngagementStatusType.NEW].color],
```
arturasmckwcz: **[CONFIRMED]** `statusColorMap`/`statusIconMap` still hand-list every `AgentEngagementStatusType` pair inside this `fromPairs([...])` array, which is cast `as Record<K, V>` a few lines above — a cast that bypasses TypeScript's exhaustiveness check, unlike the new `agentEngagementStatusConfig.ts` which is a directly-typed `Record` and does enforce it.

A future 7th `AgentEngagementStatusType` member gets added to the SDK: `agentEngagementStatusConfig.ts` fails to compile until a dev adds it there, but nothing forces them to also add the matching pair here. If they only satisfy that one compile error, `statusColorMap[status]`/`statusIconMap[status]` silently return `undefined` for the new status in this file — reproducing verbatim the bug (INCONTACT/TRIED_TO_CONTACT rendering with no color/icon) this PR was written to fix, with zero compiler signal. Deriving the pairs via `Object.entries(agentEngagementStatusConfig)` instead of hand-listing keys would make this impossible — the file's own `manualStatusColorMap`/`manualStatusIconMap` pattern a few lines below (directly-typed `Record<OpportunityManualStatusType, ...>`, not cast) already shows how.

src/components/Dashboard/common/statusMaps.ts
```diff
@@ -1,27 +1,25 @@
...
-  [AgentEngagementStatusType.NEW]: "var(--color-green-100)",
-  [AgentEngagementStatusType.INCONTACT]: "var(--color-green-100)",
-  [AgentEngagementStatusType.TRIED_TO_CONTACT]: "var(--color-green-100)",
+  [AgentEngagementStatusType.ACTIVE]: agentEngagementStatusConfig[AgentEngagementStatusType.ACTIVE].color,
+  [AgentEngagementStatusType.UNRESPONSIVE]: agentEngagementStatusConfig[AgentEngagementStatusType.UNRESPONSIVE].color,
+  [AgentEngagementStatusType.INACTIVE]: agentEngagementStatusConfig[AgentEngagementStatusType.INACTIVE].color,
+  [AgentEngagementStatusType.NEW]: agentEngagementStatusConfig[AgentEngagementStatusType.NEW].color,
+  [AgentEngagementStatusType.INCONTACT]: agentEngagementStatusConfig[AgentEngagementStatusType.INCONTACT].color,
+  [AgentEngagementStatusType.TRIED_TO_CONTACT]:
+    agentEngagementStatusConfig[AgentEngagementStatusType.TRIED_TO_CONTACT].color,
```
arturasmckwcz: Centralizing via `agentEngagementStatusConfig` silently changes the color/icon this `StatusValue` map previously assigned to `NEW`/`INCONTACT`/`TRIED_TO_CONTACT` — but this whole branch of the map looks dead: its only current consumer (`AgentCard.tsx` via `Dashboard/common/StatusBadge.tsx`) always passes an `AgentVolunteerSearchType`, never an engagement status.

If someone later wires `Dashboard/common/StatusBadge` (the near-duplicate of the already-wired `Profile/common/StatusBadge`) to an agent's `statusEngagement` — a plausible copy-paste since the two `StatusBadge` components look interchangeable — it'll silently pick up these values with no test coverage catching a mismatch against the Profile copy.

src/components/Dashboard/common/agentEngagementStatusConfig.ts
```diff
@@ -0,0 +1,44 @@
...
+  ChartLineIcon,
+  ChatsCircleIcon,
+  PhoneDisconnectIcon,
+  PhoneXIcon,
+  SparkleIcon,
+  StopCircleIcon,
+} from "@phosphor-icons/react";
+import { AgentEngagementStatusType } from "need4deed-sdk";
+
+type IconComponent = React.ComponentType<{ size?: number; color?: string }>;
```
arturasmckwcz: `type IconComponent = React.ComponentType<{ size?: number; color?: string }>` is a third verbatim copy of a type already defined in `Dashboard/common/statusMaps.ts` and `Dashboard/Profile/common/statusMaps.ts`, instead of being imported from one shared location. The three structurally-identical definitions can drift silently (e.g. one gains a `weight` prop for a new icon variant) with no compiler warning tying them together.

arturasmckwcz (commented):
@DarrellRoberts Re-review findings (2 confirmed, 2 lower-severity) — see inline comments below.

One more issue that couldn't be anchored inline because `Dashboard/Agents/Filters/constants.ts` isn't touched by this PR: `defaultAgentCardsFilter.engagementStatus` still only lists the four old statuses, and `public/locales/{en,de}/translations.json`'s `dashboard.agents.filters.engagementStatus` block has no labels for `INCONTACT`/`TRIED_TO_CONTACT` either. Once a coordinator sets an agent to 'In contact' or 'Tried to contact' via the profile dialog this PR now correctly colors and labels, the Agents list page's engagement-status filter sidebar has no checkbox for either new state — the feature is visible on the profile but not filterable on the list.

PR #928, opened 2026-08-12
🐛 Fix category card text overflow

### LINKED ISSUE #914: Landing page: opportunity/category card text overflows the card box (fixed height, no overflow handling)
Image

## Description
On the landing page's volunteering-categories section (the cards like "Veranstaltungen", "Allgemeine ehrenamtliche Tätigkeiten"), longer text (especially German, which runs longer than English) overflows past the bottom of the rounded card box instead of the box growing to fit it.

## Root cause
`src/components/VolunteeringCategories/CategoryCard.tsx`:
```tsx
const Card = styled(BaseCard)`
  background-color: var(--color-sand);
  width: var(--homepage-volunteering-category-card-width);
  height: var(--homepage-volunteering-category-card-height);
  ...
`;
```
`height` is a fixed pixel value per breakpoint (`globals.css`, 328px/380px/350px depending on viewport) instead of `min-height`, and `BaseCard` (`src/components/styled/container.tsx`) adds no overflow handling (`display: flex; flex-direction: column`, nothing else). Text longer than the fixed height simply spills out below the card.

## Fix
Change `height` to `min-height` on `Card` in `CategoryCard.tsx` so the box grows to fit its content, matching what the design visually needs. (Alternative: clip with `overflow: hidden` + line-clamp, but that would truncate real content, which seems worse than letting the card grow.)

## Related
Part of the "port remaining landing-page sections from www.need4deed.org" epic.

Justber79:
## Description
Fixes the volunteering category cards on the landing page so longer text no longer overflows the card container. Replaced the fixed `height` with `min-height`, allowing cards to grow with their content.

## Related Issues
Closes #914

## Changes
- Replaced fixed `height` with `min-height` in `CategoryCard`
- Verified the card expands correctly for longer content, including German text

### Before
before

### After
after

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/src/components/VolunteeringCategories/CategoryCard.tsx
+++ b/src/components/VolunteeringCategories/CategoryCard.tsx
@@ -7,7 +7,7 @@ import { BaseCard, IconDiv } from "../styled/container";
 const Card = styled(BaseCard)`
   background-color: var(--color-sand);
   width: var(--homepage-volunteering-category-card-width);
-  height: var(--homepage-volunteering-category-card-height);
+  min-height: var(--homepage-volunteering-category-card-height);
   padding-top: var(--homepage-volunteering-category-card-padding-top);
   padding-right: var(--homepage-volunteering-category-card-padding-right);
   padding-bottom: var(--homepage-volunteering-category-card-padding-bottom);


arturasmckwcz (approved):
@Justber79 
LGTM thank you!

nadavosa:
Confirmed — this is exactly the fix from #914 (`height` → `min-height` on `CategoryCard`'s `Card`, so the box grows to fit longer German text instead of clipping). Minimal, correct, matches the root cause exactly. Nothing to add.

PR #734, opened 2026-06-24
🐛 fix: resolve TypeScript errors from SDK drift

arturasmckwcz:
## Summary

- Fixes 70+ TypeScript errors caused by the fe codebase drifting behind SDK changes
- Adds `.dockerignore` to exclude `node_modules`/`.next`/`.git` from build context (drops it from ~2 GB to a few MB)

### SDK changes accommodated

| Area | What changed |
|---|---|
| `OpportunityMatchStatusType` | Renamed → `OpportunityMatchStatus` |
| `OpportunityStatusType.INACTIVE` | Removed → mapped to `PAST` |
| `VolunteerStateMatchType.PAST` | Removed → mapped to `NO_MATCHES` |
| `OpportunityMatchStatus.{NO_MATCHES,PAST}` | Removed from enum |
| `ContactMethodType.VIDEO_CALL` | Removed |
| `AgentMembershipStatus`, `ApiAgentMembership`, `ApiAgentRegister*` | Not in SDK → defined locally in `AgentRegistration/types.ts` |
| `ApiUserGet.personId` | Not in SDK → extended locally where needed |
| `ApiAgentGetList.email`, `ApiVolunteerGetList.{statusMatch,email,statusCommunication}` | Not in SDK → extended locally |
| `ApiVolunteerOpportunityGetList.{statusMatch,accompanyingDetails,agentTitle,numberOfVolunteers}` | Not in SDK → extended locally |
| `ApiOpportunityAgent.id` | Not in SDK → extended locally |
| `ApiOpportunityAccompanyingDetails.{refugeeLanguage,appointmentPostcode,appointmentLanguage}` | Not in SDK → extended locally |
| `ApiVolunteerGet.{statusVaccinationDate,statusCGCDate,statusCGCApplicationDate}` | Not in SDK → extended locally |
| `ApiAgentPatch.{addressStreet,addressPostcode}` | Not in SDK → local `OrganizationDetailsPatch` type |
| `textarea-caret` | No `@types` package → local declaration in `src/types/textarea-caret.d.ts` |

## Test plan

- [ ] `yarn typecheck` passes with zero errors
- [ ] `yarn lint` passes
- [ ] Docker build context is small (verify with `docker build .` — context transfer should be < 10 MB)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #929, opened 2026-08-13
exclude inactive records from default volunteer and opportunity views

### LINKED ISSUE #637: feat: exclude inactive records from default view — volunteers, opportunities, and summary counts
## Description

Inactive volunteers and inactive opportunities should not appear in the dashboard by default. They should be hidden from both the list/card views and from the summary counts shown at the top of the search views. A user can opt in to seeing them by explicitly selecting the inactive status in the filters — at that point the inactive records appear and are included in the counts.

## Current behaviour

Inactive volunteers and opportunities appear in search results and are included in the total counts at the top of the search view regardless of the active filter selection.

### Default state (no filter selected)
- Inactive volunteers are excluded from the volunteer list/card view
- Inactive opportunities are excluded from the opportunity list/card view
- The summary counts at the top of each search view reflect only active records (inactive excluded)

### When "Inactive" status is explicitly ticked in the filters
- Inactive records appear in the list/card view alongside active ones
- The summary counts include inactive records

## Scope

- Volunteer search view — list, card, and summary counts
- Opportunity search view — list, card, and summary counts
- The filter for status (inactive checkbox/option) must remain available so users can opt in

## Acceptance Criteria

- [ ] Inactive volunteers are hidden by default and excluded from volunteer summary counts
- [ ] Inactive opportunities are hidden by default and excluded from opportunity summary counts
- [ ] Ticking the inactive filter option reveals inactive records and includes them in counts
- [ ] Unticking the inactive filter re-hides them and restores the default counts
- [ ] Coordinator/admin behaviour is unchanged in any view that is already role-restricted

ivannissimrch:
## Description
Inactive volunteers and inactive opportunities no longer show up in the dashboard by default, and are not counted in the results number at the top. Ticking "Inactive" in the filters brings them back.

The reason nothing was filtered before: the query only got a `status` (or `engagement`) param for boxes the user had ticked. On a fresh page load, nothing is ticked, so no param was sent at all, so the server had nothing to filter by and returned everything.

The fix supplies a default set when the user has selected nothing, and it does that at the API call site rather than inside the shared serializer. 

Both default lists are derived from the SDK enums rather than written by hand, so a new status added later shows up by default instead of silently disappearing.

The counts needed no work. `count` comes back from the server with the filtered list, so filtering the query filters the count.

useCopyEmails gets a snapshot of the filter taken before the defaults are applied, so copy-all still includes inactive volunteers

**Also open:** the issue only mentions inactive, so past opportunities are still visible by default. I was on the fence about including opp-past too and went with what the issue says. Happy to add it, it is one line.

## Related Issues
Closes #637 

## Changes
- Restored the Inactive option to the opportunity status filter. It was added by #435 on 4 May and removed by #734 (`834888b4`) on 24 June, when the SDK had temporarily dropped `OpportunityStatusType.INACTIVE`. The SDK has the value back, and the `opp-inactive` translation keys were still in both locale files, so this is a one-line revert.
- `DEFAULT_OPPORTUNITY_STATUSES` and `DEFAULT_VOLUNTEER_ENGAGEMENTS`, both derived from the SDK enums
- Both list controllers supply those defaults when no filter is selected
- The volunteer side strips the `vol-` prefix the same way the existing serializer does, since the backend re-adds it

## Screenshots / Demos

https://github.com/user-attachments/assets/a05d0a24-03a4-4b25-a8b9-abaca752ccdd

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes

### DIFF
--- a/src/components/Dashboard/Opportunities/Filters/constants.ts
+++ b/src/components/Dashboard/Opportunities/Filters/constants.ts
@@ -54,6 +54,8 @@ export const DEFAULT_OPPORTUNITY_STATUSES = Object.values(OpportunityStatusType)
   (status) => status !== OpportunityStatusType.INACTIVE,
 );
 
+export const STATUS_PARAM = "status";
+
 export const SEPARATOR = "~";
 export type AvailabilityKeys = keyof OpportunityCardsFilter["availability"];
 export type AvailabilitySubKeys = TimeSlot | ByDay | OccasionalType;
--- a/src/components/Dashboard/Opportunities/Filters/constants.ts
+++ b/src/components/Dashboard/Opportunities/Filters/constants.ts
@@ -17,6 +17,7 @@ export const defaultOpportunityCardsFilter: OpportunityCardsFilter = {
     [OpportunityStatusType.NEW]: false,
     [OpportunityStatusType.SEARCHING]: false,
     [OpportunityStatusType.ACTIVE]: false,
+    [OpportunityStatusType.INACTIVE]: false,
     [OpportunityStatusType.PAST]: false,
   },
   type: {
@@ -47,6 +48,11 @@ export const defaultOpportunityCardsFilter: OpportunityCardsFilter = {
     },
   },
 };
+
+// Every status except inactive. Derived so new SDK values are visible by default
+export const DEFAULT_OPPORTUNITY_STATUSES = Object.values(OpportunityStatusType).filter(
+  (status) => status !== OpportunityStatusType.INACTIVE,
+);
 
 export const SEPARATOR = "~";
 export type AvailabilityKeys = keyof OpportunityCardsFilter["availability"];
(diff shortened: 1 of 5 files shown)


nadavosa:
## Review

Good root-cause explanation (no param sent on fresh load → server returns everything), and deriving the default sets from the SDK enums instead of hand-writing them is the right call for future-proofing.

### Copy-emails vs. visible list can now silently disagree
```ts
const emailsFilter = new URLSearchParams(serializedFilter);

if (!serializedFilter.has(QueryParamsKeys.ENGAGEMENT)) {
  DEFAULT_VOLUNTEER_ENGAGEMENTS.forEach((defaultEngagement) => ...);
}
```
On a fresh page load (no explicit filter), the visible list/count now excludes inactive volunteers, but `emailsFilter` is snapshotted *before* the default is applied, so "copy emails" still includes them. That means a coordinator could see "42 volunteers" on screen but copy more than 42 emails, with nothing indicating why the numbers don't match. The PR description calls this out as deliberate, but I'd double check that's actually the desired UX rather than just the easiest thing to preserve — it seems like it'd be confusing in practice. If keeping it, might be worth a tooltip/label near the copy button noting it includes inactive volunteers.

### Scoping question (opp-past)
Agreed with going with the issue's literal scope (inactive only) rather than folding in `opp-past` here — issue #637 only mentions inactive, and expanding scope in the same PR that fixes a specific reported bug adds risk for an unrequested change. Worth its own issue if past-opportunities-hidden-by-default is also wanted.

No test coverage added (checklist notes this), consistent with other PRs in this repo.

Issue #917, opened 2026-08-10
Landing page: implement Events section (currently a stub), button linking to /event-page, title fetched from the event page


nadavosa:
## Description
Fill in the existing placeholder for the "Upcoming Events" banner section on the new landing page (app.need4deed.org) — matching the one on www.need4deed.org: a background-image banner with an "Upcoming Events" heading, the current event's title/blurb, and a "Learn More & Register" button.

## Current state
`src/components/EventsSection/EventsSection.tsx` is a stub:
```tsx
<div style={{ border: "1px dashed ..." }}>Events Section (Coming in #137)</div>
```

## Requirements (per request)
- Button links to the current event page: `https://www.need4deed.org/event-page`.
- The section's text should show the current event's title, **fetched from the event page** rather than hardcoded a second time in `fe`.

## Open question: where does the title come from?
This needs a decision before implementation, flagging what I found so far:
- `website`'s own homepage "Upcoming Events" card (`src/components/EventsSection/EventCard.tsx`) is *also* currently broken — it calls a `be` endpoint (`GET /api/event/`) that doesn't exist, so it always shows "No upcoming event available currently," even though a real event already sits in `website`'s `public/data/{en,de}/events.json`, and `website`'s own `/event-page` (`src/pages/EventPage.tsx`) hardcodes its event content directly in JSX rather than reading from that same file.
- So there isn't yet a single, working, authoritative source for "the current event's title" to fetch from — `website`'s own homepage card has the identical problem this issue is asking `fe` to avoid.
- Options to resolve before building this: (a) fetch `website`'s public `events.json` directly (e.g. `https://www.need4deed.org/data/en/events.json`, if that's how it's actually served — needs confirming), (b) wait for `website`'s own event data plumbing to be fixed and share whatever ends up being the real source then, or (c) some other shared source. Recommend deciding this explicitly rather than hardcoding the event title a third time (`fe` would then be a third place, after `website`'s homepage card and `EventPage.tsx`, that can drift out of sync).

## Ask
- Implement the banner UI (background image, heading, event title/blurb, "Learn More & Register" button → `https://www.need4deed.org/event-page`).
- Resolve the open question above for where the title text comes from.

## Related
Closes fe#137 (referenced by the existing stub). Part of the "port remaining landing-page sections from www.need4deed.org" epic.

PR #930, opened 2026-08-13
Landing page: Add YouTube video section ("How you can support refugees in your community")

gshiva53:
## Description
- Adds the YouTube video and header title mentioned in issue #915. 
- Uses the existing styling for the video. 
- Adds both `en` and `de` translations for the section title.
- Styling is screen responsive.

## Related Issues
Closes #915 

## Changes
- Translations for en and de
- `RefugeeSupport` component for homepage

## Screenshots / Demos
The homepage will look like the following screenshot after this PR is merged 
Screenshot 2026-08-13 at 10 50 51 pm

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes


public/locales/de/translations.json
```diff
@@ -29,6 +29,9 @@
         "cookies": "Cookies"
       }
     },
+    "iframeTitles": {
+      "howToVolunteer": "Wie Du Geflüchtete in Deiner Nachbarschaft unterstützen können"
+    },
     "testimonials": {
```
gshiva53, line 35: German translation for the heading and video title. Taken from the need4deed-org website.

public/locales/en/translations.json
```diff
@@ -31,6 +31,9 @@
         "cookies": "Cookies"
       }
     },
+    "iframeTitles": {
+      "howToVolunteer": "How you can support refugees in your community"
+    },
     "testimonials": {
```
gshiva53, line 37: English translation for the heading and video title. Taken from the need4deed-org website.

src/components/RefugeeSupport/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./RefugeeSupport";
```
gshiva53, line 1: I don't think this is required. I may be wrong. Added just to maintain consistency with other components.

src/components/Website/Landing/Landing.tsx
```diff
@@ -32,6 +33,7 @@ export function Landing({ lang }: { lang: Lang }) {
       <VolunteeringCategoriesSection />
       {/* Transferred Components */}
       <VolunteeringOpportunitiesSection />
+      <RefugeeSupport />
       <EventsSection />
```
gshiva53, line 37: Importing and rendering the refugee support component on the landing page.

gshiva53:
Hi @need4deed, requesting for review for this PR 😄. I didn't know who else to tag for this.

nadavosa:
## Review

Matches issue #915 well — reuses the plain-iframe pattern from the old site's `website` repo (same video id), and the security attributes (`sandbox`, `referrerPolicy`, `loading="lazy"`) are a nice addition the old site didn't have.

### Reused CSS variables from an unrelated section
```ts
const IFrame = styled.iframe`
  width: var(--homepage-process-section-video-width);
  height: var(--homepage-process-section-video-height);
  border-radius: var(--homepage-process-section-video-border-radius);
`;
```
These variable names belong to the `ProcessSteps` section's own video, not this new `RefugeeSupport` section. Reusing them works today because the values happen to be what you want, but it couples two unrelated sections' styling — if `ProcessSteps`'s video size ever changes for its own reasons, this section silently changes too. Worth its own `--homepage-refugee-support-video-*` variables instead.

### Minor: `sandbox="allow-scripts allow-same-origin ..."`
This combination is generally flagged (MDN calls it out) because scripts + same-origin together let embedded content escape its own sandbox restrictions — low risk here since the `src` is a fixed, trusted YouTube URL rather than user input, and YouTube's player likely needs `allow-same-origin` to work correctly, but worth being aware of if this pattern gets copied elsewhere for less-trusted content.

### Minor: section ordering
On the old site, this video section sits directly above "How to start" (ProcessSteps). Here it's placed between the still-stubbed `VolunteeringOpportunitiesSection` and `EventsSection` instead. Probably fine since the stub sections around it are being reworked anyway (#917/#918), just flagging in case the order matters for now.

### Heads up: likely conflicts with #931
#931 removes `<VolunteeringOpportunitiesSection />` from `Landing.tsx` on the same lines this PR adds `<RefugeeSupport />` next to. Worth merging #931 first (already approved) and rebasing this on top, or coordinating the merge order.

Justber79:
The section works well and the video displays correctly. 😊 I noticed one small responsiveness issue on mobile devices: the video is shifted to the right and touches the right edge instead of being centered with equal spacing on both sides. I’ve attached screenshots for reference.

#930 mobile review
#930 mobile review

public/locales/de/translations.json
```diff
@@ -29,6 +29,9 @@
         "cookies": "Cookies"
       }
     },
+    "iframeTitles": {
+      "howToVolunteer": "Wie Du Geflüchtete in Deiner Nachbarschaft unterstützen können"
```
Justber79: The German text has a conjugation mismatch: “Du … können” should be “Du … kannst”. Could we change this to “Wie Du Geflüchtete in Deiner Nachbarschaft unterstützen kannst”?

src/components/RefugeeSupport/index.ts
Justber79: Yes, this barrel export is needed for the current import style: @/components/RefugeeSupport. Without it, the import would need to reference RefugeeSupport/RefugeeSupport directly.

src/components/RefugeeSupport/RefugeeSupport.tsx
```diff
@@ -0,0 +1,32 @@
+import styled from "styled-components";
+import { useTranslation } from "react-i18next";
+import { Heading2 } from "../styled/text";
+import { FullWidthContainer, SectionContainer } from "../styled/container";
+
+const IFrame = styled.iframe`
+  width: var(--homepage-process-section-video-width);
+  height: var(--homepage-process-section-video-height);
+  border-radius: var(--homepage-process-section-video-border-radius);
```
Justber79: Could we also add border: 0 here? Browsers may otherwise render the iframe with a default border.

src/components/RefugeeSupport/RefugeeSupport.tsx
```diff
@@ -0,0 +1,32 @@
+import styled from "styled-components";
+import { useTranslation } from "react-i18next";
+import { Heading2 } from "../styled/text";
+import { FullWidthContainer, SectionContainer } from "../styled/container";
+
+const IFrame = styled.iframe`
+  width: var(--homepage-process-section-video-width);
```
Justber79: Have we tested this at the smallest supported mobile width? The fixed 360px mobile width may be clipped on narrower screens, especially when section padding is included.

Justber79 (changes requested):
Thanks for adding this section. I found one German translation issue that should be corrected. I also added two suggestions about the iframe styling and mobile responsiveness.

PR #933, opened 2026-08-18
Add fe image build workflow

### LINKED ISSUE #932: No image build workflow: fe cannot be built and published from the org
The fe repo has no workflow that builds and publishes a container image, so the image running in production is built from a fork. Completing the move of production onto org-owned infrastructure (need4deed-org/infra#7) needs the org to be able to build its own fe image.

A \`build-fe.yaml\` workflow (manual \`workflow_dispatch\`, publishes \`ghcr.io/need4deed-org/fe:develop\`) fixes this. The image bakes the production build args: \`API_URL=http://be:8000\` (the fe server proxies \`/api/*\` to the be in-cluster; the browser stays same-origin) and the \`cdn.need4deed.org\` asset hosts.

theodore-evans:
Adds `.github/workflows/build-fe.yaml`: a manually triggered (`workflow_dispatch`) workflow that builds the fe container image and publishes it to `ghcr.io/need4deed-org/fe:develop`.

This is the image serving production at `app.need4deed.org` right now, currently built from my fork because the org repo has no image workflow. Merging this (and running it once) lets the org build its own fe image, which is one of the steps to move production fully onto org-owned repos - see need4deed-org/infra#7 and the pins in its `overlays/prod/kustomization.yaml`.

Notes on the baked build args (Next.js `NEXT_PUBLIC_*` values are fixed at build time, so they live here, not in the deployment env):

- `API_URL=http://be:8000`: the fe server proxies `/api/*` to the be service in-cluster; the browser stays same-origin, so no public API host is needed.
- `NEXT_PUBLIC_CLOUDFRONT_URL` / `NEXT_PUBLIC_CLOUDFRONT_DATA_URL` point at `cdn.need4deed.org`, the Traefik->cdn-proxy route to Infomaniak Object Storage.
- `GIT_COMMIT_SHA` is stamped for traceability.

The GHCR namespace comes from `github.repository_owner`, so the same file builds to the org namespace here and to a fork's namespace when run there. Publishing uses the workflow's own `GITHUB_TOKEN` (`packages: write`); no new secrets needed.

Closes #932

### DIFF
--- a/.github/workflows/build-fe.yaml
+++ b/.github/workflows/build-fe.yaml
@@ -1,0 +1,32 @@
+name: build-fe
+
+on:
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Log in to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push fe image
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          push: true
+          tags: ghcr.io/${{ github.repository_owner }}/fe:develop
+          build-args: |
+            API_URL=http://be:8000
+            NEXT_PUBLIC_CLOUDFRONT_URL=https://cdn.need4deed.org/images
+            NEXT_PUBLIC_CLOUDFRONT_DATA_URL=https://cdn.need4deed.org/data
+            GIT_COMMIT_SHA=${{ github.sha }}


nadavosa:
## Review

Straightforward, minimal-permissions workflow (`contents: read` + `packages: write` only, no new secrets) — makes sense as a step toward org-owned build infra per the linked `infra#7`.

### The pushed tag doesn't reflect what branch was actually built
```yaml
tags: ghcr.io/${{ github.repository_owner }}/fe:develop
```
Since this is `workflow_dispatch`-only, whoever triggers it can pick *any* branch from the GitHub UI's branch selector, but the image always gets pushed to the `:develop` tag regardless of what was actually checked out and built. If someone runs this from a feature branch by mistake, it'd silently overwrite the tag that (per the description) is "the image serving production at app.need4deed.org right now." Worth either using `${{ github.ref_name }}` in the tag, or adding a guard that fails the run unless triggered from `develop`.

### Minor: no immutable tag alongside `:develop`
Only the mutable `:develop` tag gets pushed — no per-commit tag (e.g. `:${{ github.sha }}`) for rollback/provenance. `GIT_COMMIT_SHA` is baked into the app itself for runtime introspection, which helps, but a multi-tag push (`tags: |` with both `:develop` and `:${{ github.sha }}`) is a common pairing so a specific past build can still be pulled/rolled back to directly.

Not blocking either way — this is a big improvement over building from a fork, just flagging the branch/tag mismatch as worth closing before this becomes the routine way production images get built.

arturasmckwcz (approved):

PR #936, opened 2026-08-20
add skills filter to the opportunity list

### LINKED ISSUE #905: Filter opportunities by skills in the dashboard list view
## Context

Coordinators viewing the opportunity list can filter by type, status, district, language, activity, and availability — but not by skill, even though skill data already exists on opportunities and a skill-picker UI already exists elsewhere in the app for a different purpose.

## Current state (researched against `develop` on both repos)

- **FE**: no `skill` field exists in `OpportunityCardsFilter` (`src/components/Dashboard/Opportunities/Filters/types.ts`), `defaultOpportunityCardsFilter` (`constants.ts`), `FiltersContent.tsx`, or the URL (de)serialization helpers (`serializeOpportunityFilters`/`deserializeOpportunityFilters` in `Opportunities/helpers.ts`). The existing `activity` filter is functionally identical to what a `skill` filter needs and is the exact pattern to copy end-to-end (`[EntityTableName.ACTIVITY]` → `[EntityTableName.SKILL]`, same `generateNestedFilterControlItems` helper, same accordion structure).
- **Existing skill-picker precedent**: `src/components/Dashboard/NewOpportunity/fields/SkillsField.tsx` already renders a checkbox-list of skills for the create/edit opportunity form. Different UX shape than a multi-select list filter, but confirms skill data + labels are already available client-side via the same `apiSkills`/`createMapping` pattern.
- **BE gap**: `QuerystringOpportunityFiltering` (`be`'s `src/server/types/endpoint-handlers.ts`) already declares `filter.skill?: string` — but it's unused. `getOpportunityWhere` (`src/server/utils/data/get-opportunity-where.ts`) has no `skill` branch, even though it has a near-identical `activity` branch (`deal.dealActivity.activity.id`) that a `skill` branch (`deal.dealSkill.skill.id`) would mirror exactly. The list GET route already eager-loads `deal.dealSkill.skill`, so the relation is available, just not filterable.
- **No SDK change needed**: `EntityTableName.SKILL` already exists in `need4deed-sdk`, and `QuerystringOpportunityFiltering` lives in `be`'s own types, not the SDK package — this can ship without a contract change.

## Scope

1. Add a skills filter to the opportunity list (FE), following the `activity` filter's pattern exactly: `OpportunityCardsFilter`, `defaultOpportunityCardsFilter`, `createOpportunityFilterItems`, `serializeOpportunityFilters`/`deserializeOpportunityFilters`, and a new `AccordionFilter` in `FiltersContent.tsx`.
2. Wire the BE side: add a `skill` branch to `getOpportunityWhere` mirroring the existing `activity` branch. This is `be` work, not `fe` — needs a companion `be` issue/PR (see below) unless picked up as part of the same effort.

## Depends on

BE wiring for `filter.skill` in `getOpportunityWhere` — the querystring type already declares the field, but the actual `WHERE` clause branch doesn't exist yet. The FE filter is inert without it.

## Acceptance criteria

- [ ] Coordinators can filter the opportunity list by one or more skills, same interaction pattern as the existing activity filter.
- [ ] The skill filter is reflected in the URL (shareable/bookmarkable), consistent with other filters.
- [ ] Filtering by skill returns only opportunities whose `Deal` has a matching `DealSkill`.

ivannissimrch:
## Update 2026-08-26

 be#892 landed on Aug 22 and added the `skill` branch to `getOpportunityWhere`, so the note below about the filter being inert no longer applies. Tested end to end: ticking a skill narrows the list and the param round-trips through the URL.

## Description
Adds a skills filter to the opportunity list, following the existing activity filter pattern.

 **This needs a backend change first.** `getOpportunityWhere` has no `skill` branch. So the filter renders and reaches the URL. Still, nothing is filtered until that lands. `filter.skill` is already declared in `QuerystringOpportunityFiltering` and the list route already eager-loads `deal.dealSkill.skill`, so only the where-clause is missing. No SDK change needed.

I have the backend side working locally with a unit test, verified end-to-end against my own backend, but I could not open the BE PR. The pre-push hook runs the full test suite, and it does not pass on my machine: `getDateObj` expects day 11 and I get 10, and the onetimer test expects 2026-08-01, and I get 2026-07-31. Both are one day off because I am in UTC-5, and those tests assume a European time zone. `swagger.test.ts` also times out at 5s intermittently. None of them touch my change.

**One more thing worth knowing.** While adding the skill branch, I found a separate bug in the same function: `language`, `district` and `activity` all write the same `deal` key, and object spread replaces rather than merges on a repeated key, so combining two of them silently drops one. Measured: one activity alone returns 1, and that same activity plus a language that does not apply still returns 1 instead of 0. It is on `develop` now and unrelated to this feature.

**And when you test this:** the skills dropdown depends on `/option` returning skills. If an environment has no `option` rows with `item_type='skill'`, the accordion renders empty. My local database had none until I seeded them.

## Related Issues
Closes #905

## Changes  
- `skill` added to `OpportunityCardsFilter` and `defaultOpportunityCardsFilter`
- `skillFilters` in `createOpportunityFilterItems`, plus the selected-filters flat array
 - One more `AccordionFilter` in `FiltersContent.tsx`
 - URL serialize and deserialize, so the filter is shareable like the others
 - `skills` translation key in en and de

## Screenshots / Demos
Screenshot from 2026-08-19 19-39-35

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes


nadavosa (commented):
Reviewed the diff and checked it out locally.

**FE code looks correct.** The skill filter mirrors the activity filter pattern exactly across `types.ts`, `constants.ts`, `helpers.ts` (both the Filters one and the parent one), `FiltersContent.tsx`, and `Opportunities.tsx`. I grepped every `EntityTableName.ACTIVITY` usage in `src/components/Dashboard/Opportunities` and confirmed a matching `SKILL` counterpart was added at each spot, nothing missed.

Confirmed `EntityTableName.SKILL` already exists in the published `need4deed-sdk` (`dist/types/core.d.ts`), so the "no SDK change needed" claim in the description holds. `tsc --noEmit` and lint both pass clean locally too.

**One thing worth deciding before merge:** since this ships without the BE `getOpportunityWhere` skill branch, a user can select a skill filter and it will silently do nothing to the results, no error, just no effect. Is that an acceptable gap to ship now, or should this wait and land together with the backend change so the filter isn't visibly inert in the meantime?

**Separate from this PR:** the pre-push test failures you hit (day-off-by-one date assertions assuming a European timezone, plus the flaky `swagger.test.ts` timeout) sound like a real problem blocking non-UTC-negative contributors from opening BE PRs at all. Might be worth its own BE issue so it doesn't keep biting people.

No test added here, but that matches the existing sibling filters (activity/language/district are untested too), so not blocking on that basis.

nadavosa (commented):
Update: checked the BE repo, this is already handled. need4deed-org/be#892 merged to develop on 2026-08-22, adds the `dealSkill` branch to `getOpportunityWhere` and also fixes the same-key-overwrite bug you flagged (language/district/activity/skill now accumulate into one shared `deal` object instead of clobbering each other, with tests added).

So my earlier concern about shipping a silently inert filter doesn't apply, the BE half is already merged. This should just work end to end once develop picks up both. No blocker from me.

arturasmckwcz (commented):
Automated review pass on this PR. One thing worth double-checking before merge: the PR description says the backend `skill` where-clause hasn't landed and that the BE PR couldn't be opened yet ("nothing is filtered until that lands"). But `be`'s `develop` at commit `ee963e60` (2026-08-22, same author) already added the `skill` branch to `getOpportunityWhere`, and this branch's merge-base with `origin/develop` includes that commit. So the filter is likely already fully live against current `develop`, not inert — worth confirming and testing end-to-end rather than merging on the assumption it's a no-op.

Two smaller inline notes below.

arturasmckwcz (approved):

src/components/Dashboard/Opportunities/helpers.ts
```diff
@@ -104,6 +104,15 @@ export function serializeOpportunityFilters(
     }
   });
 
+  params.delete(EntityTableName.SKILL);
```
arturasmckwcz: NIT: serializeOpportunityFilters now has a 4th near-verbatim copy-pasted block (district/language/activity/skill) instead of a shared helper parameterized by filter key. Worth generalizing so a future filter/fix doesn't need manual replication across all blocks.

src/components/Dashboard/Opportunities/helpers.ts
```diff
@@ -150,6 +159,11 @@ export function deserializeOpportunityFilters(
     newFilter.activity[l] = true;
   });
 
+  const querySkills = searchParams.getAll(EntityTableName.SKILL);
```
arturasmckwcz: NIT: deserializeOpportunityFilters similarly gains a 4th near-verbatim copy-pasted block (searchParams.getAll + forEach + set true) instead of a shared helper. Same generalization opportunity as the serialize side.

src/components/Dashboard/Opportunities/Filters/helpers.ts
```diff
@@ -42,9 +42,24 @@ export const createOpportunityFilterItems = (
     (key) => key,
   );
 
+  const skillFilters = generateNestedFilterControlItems(
```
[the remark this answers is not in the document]
ivannissimrch: Understood, I'll work on it. Thank you.

ivannissimrch:
I did the serialize and deserialize half here in fccd07f. Both loop a shared key list now.
 I started on the full config too, but it grows past those two functions. The filter type keys district and language off QueryParamsKeys, while createFilterFromOption needs EntityTableName, so one config value can't satisfy both consumers until the interface itself changes. That pulls in the types, the item builder, and FiltersContent.
 That felt like more risk on a PR you'd already approved. Do you think it's worth its own issue, or would you rather I keep going here?

arturasmckwcz:
> Do you think it's worth its own issue, or would you rather I keep going here?
@ivannissimrch yeah, please backlog an issue

ivannissimrch:
Backlogged as #970

PR #937, opened 2026-08-20
feat: coordinator-created NGO/agent profiles from the Agents table (fe#911)

### LINKED ISSUE #911: feat: coordinator-created NGO/agent profiles (no registration) from the Agents dashboard table
## Description

Add a "+" button on the upper right of the NGO/Agent list table (`src/components/Dashboard/Agents/Agents.tsx`, header rendered via `src/components/Dashboard/common/CardsHeader/CardsHeader.tsx` — which currently has no create-button slot) that lets a coordinator create a new NGO/agent profile from scratch, without going through the normal self-service `/register/agent` registration flow.

## Why

Coordinators sometimes need to add an NGO/agent to the system before that NGO has registered itself (e.g. they already have the org's details from an in-person meeting, email, etc.) — currently the only way to create an Agent is via the public registration form, which requires the NGO itself to sign up and create a login.

## Requirements

- "+" button in the top-right of the Agents table header, coordinator-only (mirrors the coordinator-only pattern already used for delete/transfer actions in #863).
- Opens a form to fill in the agent's profile fields directly (same/similar fields as `/register/agent`, minus anything that only makes sense for self-registration, e.g. password/login credentials).
- The resulting agent profile is **not tied to a user account** — no login, no registration email/verification flow. It should only be visible/manageable by coordinator (and admin) users, not surfaced anywhere an NGO user could see or claim it, unless/until a separate future flow links it to a real registered user.
- Since this creates an Agent without a corresponding User, this likely needs a `be` change too (Agent currently may assume a linked User/AgentPerson relation) — worth scoping the backend side before implementation.

## Related
- #895 (multiple NGOs per user) and #863 (coordinator-only delete/transfer on NGO/Opportunity) — same general area of coordinator-only agent-management tooling.
- #568 (closed epic: NGO/Agent User View — Permissions & Visibility Layer) for prior context on visibility rules between coordinator/agent/public views.

## Checklist
- [ ] Confirm `be` support (or scope a follow-up `be` issue) for an Agent with no linked User
- [ ] "+" button + create form, coordinator/admin-only
- [ ] New profile hidden from anyone who isn't a coordinator/admin until claimed/linked to a real registration

nadavosa:
## What
Closes https://github.com/need4deed-org/fe/issues/911

Adds a coordinator/admin-only "+" button to the Agents dashboard table header for creating a bare NGO/agent profile before it has self-registered, per the issue.

## Changes
- `CardsHeader`: new optional `headerAction` slot next to the title (no other consumer needs it yet).
- `CreateAgentDialog` (new): minimal form — organization name (required) + optional street/postcode — using `react-hook-form` + `zod`, mirroring the existing agent-contact add flow (`useCreateAgentContact` / `NewContactRow`).
- `useCreateAgent`: `POST /agent` mutation, invalidates the `["agents"]` query on success.
- Only wired into `Agents.tsx`, gated on `UserRole.COORDINATOR`/`ADMIN`.
- en/de translation keys under `dashboard.agents.createAgent`.

Deliberately does **not** reuse the multi-step self-registration wizard (`AgentRegistration/*`) — that's the right UX for a brand-new org signing itself up, not for a coordinator jotting down a placeholder record from a phone call. Richer fields (type, services, languages, about) are left for the coordinator to fill in afterward on the created agent's own profile page, same as any other agent.

## Depends on
`be` https://github.com/need4deed-org/be/pull/891 (`POST /agent`) — needs to merge/deploy first for this to actually create anything; the button will otherwise 404/500.

## Test plan
- [x] `yarn typecheck`, `yarn lint`, `yarn build` all clean.
- [ ] Not manually verified in a browser — no browser tooling available in this environment. Please click through the golden path (button visible only to coordinator/admin, form validation, success creates + shows in the list, duplicate title shows the 409 message) before merging.


nadavosa:
Self-review follow-up — fixed all 3 findings:

- Bumped \`need4deed-sdk\` 0.0.139 → 0.0.145 and switched \`useCreateAgent\` to the real (now-published) \`ApiAgentCreateResponse\` instead of the local shadow type.
- Added an "Unclaimed" badge to \`AgentCard\`/\`AgentTableRow\`, consuming the \`Agent.unclaimed\` field the follow-up review on be#891 added — without it, a coordinator had no way to visually tell a placeholder agent apart from a real one.
- Deduped the circular icon-button CSS (\`AddContactButton\` / \`CreateAgentButton\`) into a shared \`circleIconButtonStyles\` mixin instead of two copies.

Side effect: the SDK bump also pulls in \`AgentEngagementStatusType.INCONTACT\`/\`TRIED_TO_CONTACT\` (be#796), which broke typecheck in two files this PR doesn't otherwise touch. Added the exact same entries fe#923 already proposes for those two files, so this won't conflict with that PR when it lands — the real UX/copy for those statuses is still #923's job, not reinvented here.

`yarn typecheck`, `yarn lint`, and `yarn build` all pass. Still no browser available in this environment to click through it live — please verify the golden path before merging, same as noted in the PR description.

PR #950, opened 2026-08-23
give NGO users the full opportunity view

### LINKED ISSUE #934: limit ngos opportunities listst to their domain (träger)
### Description

in the ngo view of opportunities https://app.need4deed.org/de/dashboard/opportunities

they should not see the list of opportunities of everyone without the ability to click them (as it is now) but they should be able to see only the opportunities assigned to their own AGent / NGO profile and those opportunities they should be able to click and see all the information there in the card that is now available to them  

### Proposed solution

_No response_

ivannissimrch:
## Update 2026-08-27

Nadav settled the scope question: anyone from the same NGO/agent sees that NGO/agent's opportunities. Backend half is be#916 and needs to land before this one; otherwise, NGO users get a clickable full view of every organisation's opportunities.

## Description
NGO users get the same opportunity view as coordinators: the list tab, full table columns, clickable rows and cards, and the schedule filter. Previously, they got a read-only version with fewer columns and nothing clickable.

Needs the backend scoping first, so the list only contains their own opportunities. I'll open the backend issue shortly. 

Quick clarification before I open the be issue.
The one thing I want to confirm: does "their own domain" mean every shelter under that domain, or just the shelter the user belongs to? The issue says "their own operator their email domain" but then "for their own shelter", and those give different lists if a domain has more than one shelter.

## Changes
- `useAuth` returns `isAgent`; `isOwnProfile` 
- Opportunity table, cards, and filters use `isAuthorized || isAgent` instead of `isAuthorized`
- All three tabs list, cards, and map are shown to agents; the list was hidden
- `isAuthorized` itself is unchanged, so the other places that use it keep their current meaning

 ## Related Issues
  Closes #934

## Screenshots / Demos
Screenshot from 2026-08-23 13-58-19

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes

### DIFF
--- a/src/components/Dashboard/Opportunities/Filters/FiltersContent.tsx
+++ b/src/components/Dashboard/Opportunities/Filters/FiltersContent.tsx
@@ -13,7 +13,8 @@ type Props = {
 
 export default function FiltersContent({ setFilter, filter }: Props) {
   const { t } = useTranslation();
-  const { isAuthorized } = useAuth();
+  const { isAuthorized, isAgent } = useAuth();
+  const canSeeFullView = isAuthorized || isAgent;
 
   const { districtFilters, languageFilters, statusFilters, typeFilters, activityFilters, availabilityFilters } =
     createOpportunityFilterItems(filter, setFilter, t);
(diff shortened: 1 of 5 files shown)


need4deed:
@ivannissimrch to make it simple now - anyone from the same NGO/agent can see the oppotunities there of that NGO/agent

ivannissimrch:
Thanks, scoping by agent then. I created a PR on be, [916](https://github.com/need4deed-org/be/pull/916) it limits the list to the caller's own agent.

PR #955, opened 2026-08-28
feat: volunteer self-registration with email confirmation

### LINKED ISSUE #954: feat: volunteer self-registration with email confirmation
## Description

Add a public "Become a volunteer" registration page, mirroring the existing NGO/agent registration screen (`/register/agent`, `AgentRegistration.tsx` + `AccountStep.tsx`), but for volunteers: same fields (first name, last name, email, password, confirm password, phone, consent), headline "Become a volunteer" instead of "Register your organisation".

## Verified: no backend or SDK change needed

- `POST /user` (be) already accepts `role: UserRole.VOLUNTEER`. Its pre-handler comment says explicitly: "Volunteers and users self-register freely" — unlike `AGENT`, there's no email-domain restriction to work around.
- It already triggers `fastify.notify.emailVerification(savedUser)` for any role, so the email-confirmation flow (`/verify-email/[token]`) works out of the box.
- `verify-email/[token]/page.tsx` already redirects non-"agent" roles straight to `/dashboard` after verification (the agent-only branch sends to `/register/agent/complete` for org profile completion) — volunteers need no extra completion step.
- The consent checkboxes' linked pages (Privacy, `/rac-guidelines` "Need4Deed Guidelines", `/agreement` "Voluntary Participation Agreement") already address volunteers directly in their copy (e.g. "Agreement for Volunteer Participation" opens with "As a volunteer, you will..."), so the same three links apply unchanged.

## Scope

- New route, e.g. `/register/volunteer`, with a single "Your account" step identical in fields/validation to the NGO one, submitting `role: UserRole.VOLUNTEER` to `POST /user`.
- Same "check your email" success screen after submit.
- Out of scope: any post-verification profile-completion step (skills/languages/availability) — this issue is just the account-creation screen shown in the NGO flow's first step.

## Related
- Mirrors `src/components/AgentRegistration/` (`AgentRegistration.tsx`, `steps/AccountStep.tsx`, `helpers.ts`, `types.ts`, `styled.ts`).

nadavosa:
## Description

Adds a public "Become a volunteer" registration page, mirroring the existing NGO/agent registration screen (`/register/agent`) but for volunteers. Same account fields (first name, last name, email, password, confirm password, phone, consent), headline "Become a volunteer" instead of "Register your organisation".

No backend or SDK change was needed:
- `POST /user` already accepts `role: UserRole.VOLUNTEER` and self-registers it freely (its pre-handler comment: "Volunteers and users self-register freely", unlike `AGENT` which requires a matching org email domain).
- It already sends the verification email for any role.
- `verify-email/[token]/page.tsx` already redirects any non-"agent" role straight to `/dashboard` after verification, so no extra profile-completion step is needed for volunteers.
- The consent checkboxes' linked pages (Privacy, Need4Deed Guidelines, Voluntary Participation Agreement) already address volunteers directly in their copy, so the same three links apply unchanged.

## Related Issues
Closes #954

## Changes
- New route `/register/volunteer` → `VolunteerRegistration` component, POSTs `role: UserRole.VOLUNTEER`.
- Generalized `AccountStep` and `validateStep` (under `AgentRegistration/`) with an optional `namespace` param (defaults to `"agentRegistration"`, so the existing agent flow is unchanged) instead of duplicating the ~150 lines of identical field markup/validation for a second flow.
- Reused `AgentRegistration`'s `styled.ts` primitives and added a `volunteerRegistration` i18n namespace (en/de) for the copy that does differ (headline, subtitle, success screen).

## Verification
- `tsc --noEmit` and `yarn lint` pass.
- Ran the dev server and confirmed `/en/register/volunteer` renders the headline, all fields, and the consent links correctly; verified client-side validation triggers on empty submit. Did not exercise the live `POST /user` call (no backend running locally), that endpoint itself is unchanged.

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes


arturasmckwcz:
## Review findings

**1. Consent checkbox still shows agent-only legal copy to volunteers** (`src/components/AgentRegistration/steps/AccountStep.tsx:96`)

`AccountStep` was generalized with a `namespace` prop and every other label switched to `t(\`${namespace}...\`)`, but the consent line's guidelines link still uses the non-namespaced `homepage.footer.legal.guidelines` key and `Subpage.RAC_GUIDELINES` — both written for accommodation centers. A volunteer signing up at `/register/volunteer` has to check a required legal-consent box reading "...RAC Guidelines..." linking to content addressed to accommodation centers ("request matching", "volunteer screening"), not to volunteers.

**2. `PageWrapper` override duplicated instead of shared** (`src/components/VolunteerRegistration/VolunteerRegistration.tsx:31`)

The `min-height: 0; flex: 1` override (and its explanatory comment, which literally says "see the same override in AgentRegistration.tsx") is copy-pasted verbatim rather than extracted into the shared `styled.ts` next to `Wrapper`/`Card`. A future fix to how `PageLayout` composes with `Wrapper` now has to be applied in two places and can silently drift.

**3. No dedupe check against an existing `Volunteer`/`Person` by email** (`src/components/VolunteerRegistration/VolunteerRegistration.tsx:57`)

`handleSubmit` POSTs straight to `/user` with no `person.id` and no prior lookup. On the backend, `POST /user` (`src/server/routes/user.ts` ~lines 318-350) only looks up an existing `Person` when `person.id` is explicitly passed — otherwise it unconditionally creates a new one, even when a `Person` with that exact email already exists (e.g. via an existing `Volunteer` row from the legacy public form). Someone who already submitted the "Become a Volunteer" form and now self-registers here with the same email ends up with a second, disconnected `Person`/`User` pair instead of being linked to their existing volunteer profile — silently, with no error surfaced to them or a coordinator.

This third one is a pre-existing backend gap, not something this PR introduces, so I filed it as a standalone follow-up rather than blocking this PR on it: https://github.com/need4deed-org/be/issues/923 (tracked under the epic https://github.com/need4deed-org/be/issues/922).

src/components/VolunteerRegistration/VolunteerRegistration.tsx
```diff
@@ -0,0 +1,151 @@
...
+  PageTitle,
+  SuccessText,
+  SuccessTitle,
+  SuccessWrapper,
+  Wrapper,
+} from "@/components/AgentRegistration/styled";
+import { AccountRegistrationData, defaultAccountRegistrationData } from "@/components/AgentRegistration/types";
+import Link from "next/link";
+
+const NAMESPACE = "volunteerRegistration";
```
DarrellRoberts: this seems pointless to me as you're not really reusing a component, just duplicating it. Also in the `AgentRegistration` component you're using the raw string instead 
e.g.
```typescript
<PageTitle>{t("agentRegistration.title")}</PageTitle>
```

compared to

```typescript
<PageTitle>{t(`${NAMESPACE}.title`)}</PageTitle>
```

src/components/VolunteerRegistration/VolunteerRegistration.tsx
```diff
@@ -0,0 +1,151 @@
...
+export function VolunteerRegistration() {
+  const { t, i18n } = useTranslation();
+  const router = useRouter();
+  const user = useCurrentUser(true);
+  const [formData, setFormData] = useState<AccountRegistrationData>(defaultAccountRegistrationData);
+  const [errors, setErrors] = useState<Partial<Record<keyof AccountRegistrationData, string>>>({});
+  const [submitError, setSubmitError] = useState<string | null>(null);
+  const [isSubmitting, setIsSubmitting] = useState(false);
+  const [isSuccess, setIsSuccess] = useState(false);
+
```
DarrellRoberts: with our legacy `BecomeVolunteer` form found on `components/forms/BecomeVolunteer/BecomeVolunteer.tsx`, we used `tanstack useForm` hook which reduced the need for all of these `useState`

src/components/VolunteerRegistration/VolunteerRegistration.tsx
```diff
@@ -0,0 +1,151 @@
...
+      let message = t("message.errorGeneric");
+      if (axios.isAxiosError(err)) {
+        const data = err.response?.data as { message?: string } | undefined;
+        message = data?.message ?? message;
+      }
+      setSubmitError(message);
+    } finally {
+      setIsSubmitting(false);
+    }
+  };
```
DarrellRoberts: line 70-99 would be cleaner as a custom hook. We do have a custom `usePostRequest` hook but it seems a bit outdated (uses `fetch API` over `axios`)

src/components/VolunteerRegistration/VolunteerRegistration.tsx
```diff
@@ -0,0 +1,151 @@
...
+const NAMESPACE = "volunteerRegistration";
+
+// Wrapper's own min-height: 100vh would double up with PageLayout's flex: 1
+// container, adding a spurious extra viewport of empty space (see the same
+// override in AgentRegistration.tsx).
+const PageWrapper = styled(Wrapper)`
+  min-height: 0;
+  flex: 1;
+`;
+
```
DarrellRoberts: if you move this `PageWrapper` to and export this from `@/components/AgentRegistration/styled` (like the other styled components) then you don't have to declare it twice such as here & in the `AgentRegistration.tsx` component

DarrellRoberts (commented):
Just a few code-cleaniless points from me.

Also I imagine we want this component to also be what the "Join as a volunteer" to link to as well?

Image

This links to the legacy `BecomeVolunteer` form and sits on route: `/[lang]/forms/volunteer`

src/components/VolunteerRegistration/VolunteerRegistration.tsx
nadavosa: Fair, fixed — this file doesn't actually reuse any of AgentRegistration's JSX (only AccountStep/validateStep do, and those already take an explicit `namespace` param), so I dropped the const and every other `t()` call here is back to a plain literal string like `t("volunteerRegistration.title")`, matching AgentRegistration.tsx exactly.

src/components/VolunteerRegistration/VolunteerRegistration.tsx
nadavosa: Done — moved PageWrapper into styled.ts (next to Wrapper/Card) and both AgentRegistration.tsx and VolunteerRegistration.tsx now import the shared one.

src/components/VolunteerRegistration/VolunteerRegistration.tsx
nadavosa: Fair point, and I don't disagree tanstack useForm would be nicer here. I held off because AgentRegistration.tsx (what this PR was asked to mirror) uses the same plain useState approach, migrating just the volunteer flow would leave the two nearly-identical account steps on two different form-handling patterns, which reads as more inconsistent than the boilerplate itself. If you'd like, happy to do it as a follow-up that touches both flows together rather than just this one.

src/components/VolunteerRegistration/VolunteerRegistration.tsx
nadavosa: Same reasoning as the useForm comment below, AgentRegistration.tsx's handleSubmit is inline too, so I mirrored that rather than extracting only the volunteer side. Agreed a shared hook (axios-based, like useCreateEvent/useUpdateEvent in the Calendar feature) would be the cleaner long-term shape, open to doing that as a follow-up covering both registration flows.

nadavosa:
Pushed a follow-up commit addressing the review feedback:

**@DarrellRoberts** — yes, wired up. The homepage hero's "Join as a volunteer" button (`src/components/Website/Landing/Hero/Content.tsx`) now points at `/register/volunteer` instead of the legacy `/forms/volunteer` form. This is the app.need4deed.org landing page in this repo, not the separate `need4deed-org/website` marketing site, so that's the only place I touched. Also fixed the `PageWrapper` duplication and the pointless `NAMESPACE` indirection in `VolunteerRegistration.tsx` (replies inline). Left the `useForm`/custom-hook suggestions as-is for now with reasoning inline, wanted to flag rather than silently skip them.

**@arturasmckwcz** on the guidelines-link finding — I dug into `racGuidelines`'s actual content (`public/locales/en/translations.json`) before deciding what to do here. It's not agent-only content: the heading is generically "Need4Deed Guidelines" and it explicitly covers volunteers directly, e.g. "Screening of Volunteers: All new volunteers participate in a personal phone call with us... required to apply for a certificate of good conduct... proof of measles vaccination", "For volunteers who apply for accompaniment services, we conduct an additional video introductory meeting...". Same for the Agreement link, it's literally headed "Agreement for Volunteer Participation" and addresses "you" as a volunteer throughout, so that one's correct as-is too.

The one real wrinkle: the route itself is `/rac-guidelines` (an internal naming leftover — "RAC" from the accommodation-center side), so a volunteer clicking that link sees `rac-guidelines` in the URL despite the content being written for both audiences. That predates this PR (the exact same link is already live on the agent registration flow today) and renaming a live route is a bigger, separate change than this PR's scope, so I didn't fold it in here. Happy to file a follow-up if the team wants that URL renamed to something neutral like `/guidelines`, let me know.

Also, thanks for filing be#923 for the Person-dedup gap, agreed that's a real pre-existing issue and correctly out of scope for this PR.

Issue #137, opened 2025-12-01
landing events section


arturasmckwcz:
### Description

copypaste website event section

### Proposed solution

`src/components/EventsSection`

PR #971, opened 2026-09-01
feat: implement landing page events section (#917)

### LINKED ISSUE #917: Landing page: implement Events section (currently a stub), button linking to /event-page, title fetched from the event page

Justber79:
## Description

Replaces the landing page’s placeholder Events section with a responsive, API-connected “Upcoming Events” banner.

The section uses the existing Events API, which means events created through the dashboard Calendar can automatically appear on the landing page. It displays the nearest active current or upcoming event in the selected language.

## Related Issues

Closes #917

## Changes

- Replaced the “Events Section (Coming in #137)” placeholder
- Connected the section to the existing `/event` API
- Displayed the nearest active current or upcoming event
- Added the localized event title, short description and date range
- Added responsive background images based on the event type
- Added the “Learn More & Register” CTA
- Added loading and empty states
- Added English and German translations
- Added a fallback URL for CloudFront landing-page assets
- Preserved the existing landing-page styling and responsive layout

### Before

917-events-section-before

### After

917-events-section-after

## Checklist

- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -32,6 +32,12 @@
     "iframeTitles": {
       "howToVolunteer": "Wie Du Geflüchtete in Deiner Nachbarschaft unterstützen können"
     },
+    "events": {
+      "headline": "Kommende Veranstaltungen",
+      "button": "Mehr erfahren & anmelden",
+      "loading": "Kommende Veranstaltungen werden geladen…",
+      "empty": "Aktuell ist keine kommende Veranstaltung verfügbar."
+    },
     "testimonials": {
       "header": "ERFAHRUNGSBERICHTE",
       "headerParagraph": "Was Ehrenamtliche und Unterkünfte sagen"
--- a/public/locales/en/translations.json
+++ b/public/locales/en/translations.json
@@ -34,6 +34,12 @@
     "iframeTitles": {
       "howToVolunteer": "How you can support refugees in your community"
     },
+    "events": {
+      "headline": "Upcoming Events",
+      "button": "Learn More & Register",
+      "loading": "Loading upcoming events…",
+      "empty": "No upcoming event is currently available."
+    },
     "testimonials": {
       "header": "TESTIMONIALS",
       "headerParagraph": "What Volunteers & Coordinators Say"
--- a/src/components/EventsSection/EventsSection.tsx
+++ b/src/components/EventsSection/EventsSection.tsx
@@ -1,3 +1,167 @@
-export function EventsSection() {
-  return <div style={{ padding: "50px", border: "1px dashed red" }}>Events Section (Coming in #137)</div>;
+"use client";
+
+import { Button } from "@/components/core/button";
+import { ImageWithGradient } from "@/components/core/image";
+import { FullWidthContainer, OverlayingSectionContainer } from "@/components/styled/container";
+import { CustomHeading, Heading1, Heading4 } from "@/components/styled/text";
+import { eventsPublicLandingUrl, ScreenTypes } from "@/config/constants";
+import { useScreenType } from "@/context/DeviceContext";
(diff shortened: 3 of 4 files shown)


nadavosa (commented):
Reviewed the diff and checked it out locally. `tsc --noEmit` and lint both pass, CI is green. The API integration, upcoming-event selection, and date-range formatting all look correct, and I confirmed the hardcoded CloudFront fallback URL matches the one already documented elsewhere in the repo (README.md, fe-CLAUDE.md, next.config.ts, docker-compose.yml), so that's not a made-up value.

One real issue worth fixing:

**`shortDescription.replace(/\\n/g, "\n")` (`EventsSection.tsx:142`) has no visible effect.** This converts literal `\n` text sequences into actual newline characters, but the element it's rendered in (`Heading4`, `src/components/styled/text.tsx`) has no `white-space: pre-line`/`pre-wrap` set. Browsers collapse raw newline characters to a single space by default, so even when the replace does find something to convert, a multi-line short description still won't visually break onto separate lines. If the intent is to let coordinators write multi-line descriptions for the landing page teaser, `Heading4` (or a wrapping element) needs `white-space: pre-line` too; otherwise this replace call is dead code, worth double-checking whether it's needed at all, since a normal multi-line textarea entry already round-trips through JSON as real `\n` characters, not the two-character escaped form this regex targets.

Everything else looks solid, no other findings.

Justber79:
@nadavosa Thanks for catching this! I removed the unnecessary replacement and added white-space: pre-line to a component-scoped event description style, so real textarea line breaks now render correctly without affecting other headings. The fix has been pushed.

PR #972, opened 2026-09-02
956: feat volunteer profile completion after email confirmation

### LINKED ISSUE #956: feat: volunteer profile completion after email confirmation
## Description

Continuation of #954 / #955 (volunteer self-registration). Once a volunteer confirms their email, they currently land nowhere meaningful, this issue adds the missing step: a "volunteer profile" form, pre-filled with the account details they already entered (name, email, phone) and empty for everything else, that opens right after they click the confirmation link in their email.

This mirrors the existing agent flow exactly:
- `verify-email/[token]/page.tsx` already branches on the registering role: after verifying, an "agent" is sent to `/register/agent/complete?token=<verify-token>`, forwarding the same JWT as a query param (not a logged-in session), where `ProfileCompletion.tsx` collects the rest of the org's details and does `POST /agent/register?token=...`.
- Today a volunteer isn't branched at all, `verify-email` falls into the `else` case and pushes straight to `/dashboard` with no session established, which likely just bounces to login. Extend that page to also check for `role=volunteer` and redirect to the new completion route, same token-forwarding pattern.

## Update (2026-09-03): existing volunteers get a different branch

https://github.com/need4deed-org/be/pull/947 (merged) means a volunteer who self-registers with an email that already has a `Person`/`Volunteer` record (e.g. from the legacy public form) now links to that existing record instead of getting a duplicate. That changes what should happen after email verification for them specifically:

- **New volunteer** (no existing `Volunteer` profile): unchanged from the original plan below, straight into the completion form, token-based, no login yet.
- **Existing volunteer** (already has a `Volunteer` profile, just never had a login before): skip the completion form entirely, they already have real data, redirect straight to `/login` instead. After logging in they should land on their existing profile, not an empty form.

To tell these apart, `verify-email/[token]/page.tsx` needs the `hasVolunteerProfile` flag from https://github.com/need4deed-org/sdk/pull/220 (merged, published as `need4deed-sdk@0.0.149`) — `be` still needs to consume it and return it from `POST /user/verify-email` (tracked in https://github.com/need4deed-org/be/issues/943, bundle it there or split out, reviewer's call).

Also surfaced while tracing this end-to-end: even once someone reaches `/dashboard` post-login, `/dashboard/profile/page.tsx` is currently hardcoded agent-only and has no path for a volunteer to see their own profile at all — filed separately, see linked issue.

## Fields to collect

The public "become a volunteer" form (`src/components/forms/BecomeVolunteer/`, `VolunteerFormData`) already defines the full field shape we need: postcode, districts, availability, languages, activities, skills, certificate of good conduct, measles vaccination proof, how they heard about us, comments. Reuse that shape/component where practical rather than re-designing the fields. First/last name, email, and phone are already known from the account step and shouldn't be asked again.

## Backend work required (unlike #954, this one isn't fe-only)

Checked the `be` repo: there's no self-service way for an authenticated volunteer to create their own profile today.
- `POST /volunteer/legacy` is public/anonymous and not linked to any Person/User account, it's the legacy form's endpoint, not reusable as-is.
- `POST /volunteer` (the real volunteer route) requires `fastify.authenticate()` **and** `role: UserRole.COORDINATOR`, a self-registering volunteer can't call it.
- `Volunteer.personId` is a straightforward `@ManyToOne(Person)`, the same `Person` a self-registered volunteer's `User` already points to, so linking is structurally simple, but the endpoint/authorization to do it doesn't exist.

This needs a paired `be` issue for a self-service endpoint analogous to `POST /agent/register`, e.g. `POST /volunteer/register?token=...`, that verifies the token and creates the `Volunteer` linked to the already-created `Person`. Tracked at https://github.com/need4deed-org/be/issues/943.

## Scope
- Extend `verify-email/[token]/page.tsx` to redirect a confirmed volunteer to the new completion route **if** they don't already have a `Volunteer` profile (`hasVolunteerProfile: false`); redirect straight to `/login` if they do.
- New route/component (e.g. `/register/volunteer/complete`), analogous to `AgentRegistration/ProfileCompletion/`, pre-filled with nothing (name/email/phone aren't shown again since they're already saved), collecting the remaining `VolunteerFormData` fields.
- Depends on the paired `be` endpoint (be#943) and the `hasVolunteerProfile` flag on `POST /user/verify-email` (also be#943 or split out).

## Related
- #954, #955 (volunteer account-creation step, already shipped)
- https://github.com/need4deed-org/be/pull/947 (Person-linking, merged — makes "existing volunteer" detection possible)
- https://github.com/need4deed-org/be/issues/943 (backend endpoint + verify-email flag)
- Mirrors `src/components/AgentRegistration/ProfileCompletion/ProfileCompletion.tsx` and `src/app/[lang]/register/agent/complete/`

DarrellRoberts:
## Description
Completes new volunteer registration process.

Dependant on `be` endpoint which needs to be created

This branch is based on `nadavosa/954-volunteer-registration` which is used on PR #955 

## Related Issues
Closes #956 

## Changes
- Replicated `AgentRegistration` somewhat but I refactored it so that we use `tanstack-form` along with `tanstack` useMutation for consistency and better robustness
- The previous `BecomeVolunteer` form used a pilllike checkboxes for fields such as `activities` and `skills` . I felt it was better UX to use our existing checkbox Dropdown menu on the `EditableField` component

### step1
image

### step2
image

### step3
image

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "date-fns": "^4.1.0",
     "email-validator": "^2.0.4",
     "i18next": "^25.3.2",
-    "need4deed-sdk": "0.0.147",
+    "need4deed-sdk": "0.0.152",
     "next": "15.3.8",
     "react": "^19.0.0",
     "react-day-picker": "^9.13.0",
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -2211,6 +2211,10 @@
     "submit": "Absenden",
     "alreadyUser": "Bereits registriert?",
     "loginLink": "In Ihrem Konto anmelden",
+    "completion": {
+      "title": "Schließen Sie die Einrichtung Ihres Kontos ab",
+      "subtitle": "Erzählen Sie uns etwas über sich"
+    },
     "steps": {
       "account": {
         "title": "Ihr Konto",
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -2207,12 +2207,30 @@
     "title": "Ehrenamtlich werden",
     "subtitle": "Erstellen Sie ein Konto, um sich ehrenamtlich für Geflüchtete in Berlin zu engagieren.",
     "next": "Weiter",
+    "back": "Zurück",
+    "submit": "Absenden",
     "alreadyUser": "Bereits registriert?",
     "loginLink": "In Ihrem Konto anmelden",
     "steps": {
       "account": {
         "title": "Ihr Konto",
         "description": "Geben Sie Ihre persönlichen Daten ein, um ein Login zu erstellen."
+      },
+      "availabilityInfo": {
+        "title": "Verfügbarkeit, Aktivitäten & Fähigkeiten",
+        "description": "Wann sind Sie verfügbar, an welchen Aktivitäten sind Sie interessiert und welche Fähigkeiten bringen Sie mit?"
+      },
+      "address": {
+        "title": "Standort & Sprachen",
+        "description": "Wo befinden Sie sich und welche Sprachen sprechen Sie?"
+      },
+      "certificates": {
+        "title": "Letzte Details",
+        "description": "Geben Sie an, ob Sie die folgenden Zertifikate besitzen und wie Sie von uns erfahren haben."
+      },
+      "review": {
+        "title": "Überprüfung",
+        "description": "Bitte überprüfen Sie Ihre Angaben vor dem Absenden."
       }
     },
     "fields": {
@@ -2230,7 +2248,12 @@
     "errors": {
       "passwordTooShort": "Das Passwort muss mindestens 8 Zeichen lang sein",
       "passwordMismatch": "Die Passwörter stimmen nicht überein",
-      "phoneTooShort": "Die Telefonnummer muss mindestens 7 Zeichen lang sein"
+      "phoneTooShort": "Die Telefonnummer muss mindestens 7 Zeichen lang sein",
+      "missingToken": "Dein Registrierungslink ist ungültig oder abgelaufen. Bitte verwende den Link aus deiner Bestätigungs-E-Mail."
+    },
+    "success": {
+      "title": "Registrierung erfolgreich!",
+      "description": "Ihr Konto wurde erstellt. Sie werden in Kürze zur Anmeldeseite weitergeleitet."
     },
     "checkEmail": {
       "title": "Prüfen Sie Ihre E-Mail",
--- a/public/locales/de/translations.json
+++ b/public/locales/de/translations.json
@@ -2247,7 +2247,77 @@
       "consent": {
         "header": "Hiermit stimme ich zu:",
         "and": "und"
-      }
+      },
+      "postcode": "PLZ (Berlin) Deines Wohnorts*",
+      "locations": {
+        "header": "Bevorzugte Bezirke* (Berlin)",
+        "para": "*Denke daran, dass die meisten Unterkünfte außerhalb des Rings liegen!",
+        "error": "",
+        "helpText": "Die möglichen ehrenamtlichen Tätigkeiten, die wir Dir schicken, basieren sowohl auf Deinen bevorzugten Bezirken als auch auf der Postleitzahl Deines Wohnorts.",
+        "helpTitle": ""
+      },
+      "availability": {
+        "header": "Verfügbarkeit*",
+        "para": "*Denke daran, dass die meisten Tätigkeiten innerhalb der regulären Arbeitszeit von 9-17 Uhr stattfinden, zumindest am Anfang!",
+        "error": "",
+        "helpText": "Die Verfügbarkeiten von Mo-So beziehen sich auf regelmäßige ehrenamtliche Tätigkeiten. Wenn Deine Verfügbarkeiten auf bestimmte Zeiträume begrenzt sind oder erklärt werden müssen, schreibe dies bitte im letzten Abschnitt in die Kommentare.",
+        "helpTitle": ""
+      },
+      "languages": {
+        "header": "Sprachen*",
+        "headerWithoutAsterick": "Sprachen",
+        "para": "*Wenn Du Deine Sprache nicht findest, wähle bitte \"Andere\" und gib Deine bevorzugte Sprache im Kommentarfeld unten an",
+        "addLanguage": "Sprache hinzufügen",
+        "chooseLanguage": "Sprache auswählen",
+        "chooseLevel": "Niveau auswählen",
+        "removeLanguage": "Entfernen",
+        "language": "Sprache",
+        "level": "Niveau",
+        "languagesNative": {
+          "header": "Muttersprache",
+          "error": ""
+        },
+        "languagesFluent": {
+          "header": "Fließend",
+          "error": "Bitte wähle mindestens eine Sprache aus"
+        },
+        "languagesIntermediate": {
+          "header": "Mittelstufe"
+        },
+        "singleLevelError": "Bitte gib nur ein Niveau pro Sprache an",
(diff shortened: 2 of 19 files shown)


nadavosa (commented):
Reviewed the diff and checked it out locally. `tsc --noEmit` and lint both pass on this branch.

**Blocking, not code-related: this PR has real merge conflicts against `develop` right now** (`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`). It doesn't include the #955 commit that landed ~an hour before this was opened, so once conflicts are resolved, please double check `AccountStep.tsx`/`VolunteerRegistration.tsx` come out with the current `develop` versions (no `PageWrapper` duplication, no `NAMESPACE` const) rather than reverting back to the pre-cleanup versions this branch was forked from.

**Two real findings once that's sorted:**

**1. `EditableField.tsx`'s dropdown position regresses for every existing usage, not just this PR's new steps.** `DropdownList` used to render at `top: var(--editableField-dropdownList-top)`, which `globals.css` sets to `110%`. The new version hardcodes `top: 100%` for the default ("down") direction and drops the CSS var entirely, so every already-shipped `EditableField` checkbox/radio dropdown across the app (Opportunities filters, volunteer profile fields, etc.) shifts 10% closer to its trigger, not just the new `AddressStep`/`AvailabilityStep`/`CertificateStep` usages that motivated the new `dropdownDirection` prop. Worth keeping the CSS var for the "down" case and only hardcoding the new "up" case.

**2. `goodConductCertificate`/`measlesVaccination` store a translated UI label as the field value instead of a stable value.** In `CertificateStep.tsx`, the radio-list `options` are `t("form.becomeVolunteer.fields.certOfGoodConduct.true"/"false")`, and `setValue` passes whatever string the user clicked straight into `field.handleChange(value as string)`. So the actual submitted value is literally "Yes"/"No" in English or "Ja"/"Nein" in German, whatever the UI happened to render at click time, not a fixed value.

This also doesn't match the domain these fields are supposed to have. The already-shipped legacy `BecomeVolunteer` form types these as `boolean | undefined` (`dataStructure.ts`), and the SDK's own `VolunteerFormData` (imported by `types.ts` in this same PR, via the now-unused `ApiVolunteerRegister` type) types them as `DocumentStatusType`, a real enum (`YES`/`NO`/`ASKED_TO_APPLY`/`APPLIED_SELF`/`APPLIED_N4D`/`UNDEFINED`), not a free string. `ProfileCompletionData` (the type actually used for the submitted payload) isn't tied to `VolunteerFormData` at all, hence nothing catches this at compile time.

Given there's no SDK contract yet for the new `POST /volunteer/register` endpoint (per the PR description, the `be` side doesn't exist yet), this seems like the moment to get that payload shape defined in the `sdk` repo first, rather than have `ProfileCompletionData` diverge from the closest existing type it already imports. Locale-dependent stored values will also break as soon as this ships in more than one language.

Nothing else stood out, the wizard flow, step navigation, and `useForm`/`useRegisterVolunteer` wiring all look sound once the two items above are fixed.

nadavosa:
Checked this against the flow end to end, two gaps before this can merge as functional:

1. `verify-email/[token]/page.tsx` isn't part of this diff. It still only branches on `role=agent` and pushes everyone else (including volunteers) straight to `/dashboard`, which bounces to `/login` since no session is established yet. That redirect wiring for `role=volunteer` was called out explicitly in the issue's scope (https://github.com/need4deed-org/fe/issues/956) but isn't in this PR, so `/register/volunteer/complete` is currently unreachable from the actual registration flow.
2. `useRegisterVolunteer.ts` posts to `POST /volunteer/register?token=...`, which doesn't exist on `be` yet (per this PR's own description). I filed the paired BE issue: https://github.com/need4deed-org/be/issues/943.

Also currently showing merge conflicts against `develop`.

Recommend holding this until the BE endpoint lands and the `verify-email` redirect is wired up, otherwise this ships a page nothing routes to and that would 404 on submit.

DarrellRoberts:
thanks @nadavosa , there's a bit of ambiguity for the types of `certGoodConduct` and `certMeasles` , and I realised we need new `sdk` types which I should have establised first.

So will confirm these types in this `sdk` [#PR 222](https://github.com/need4deed-org/sdk/pull/222)

After that, will update this PR

nadavosa:
Heads up: the plan is for https://github.com/need4deed-org/sdk/pull/222 to merge first and be the formal contract for this form's submit payload (`ApiVolunteerRegisterNew`). It doesn't currently match what this PR sends, not just missing fields, different shapes:

- `locations`/`activities`/`skills`/`leadFrom`: sdk#222 wants `OptionById[]`/`OptionItem[]` (objects with `id`/`title`), this PR sends bare `number[]`.
- `languages`: sdk#222 wants `ApiLanguage[]` (`{id, title, proficiency, purpose}`), this PR sends `LanguageObject[]` (`{id, language, level}`).
- `availability`: sdk#222 wants `ApiAvailability[]` (`{day, daytime}` name-based entries), this PR sends the local `Availability` grid shape (`{weekday, timeSlots: {id, selected}[]}[]`).

I already reworked the backend side (https://github.com/need4deed-org/be/pull/950) to match sdk#222 field-for-field, so once this form is updated to send that shape, the two should line up. Full discrepancy writeup is on sdk#222 and be#950 if useful: https://github.com/need4deed-org/sdk/pull/222#issuecomment-5538913915

arturasmckwcz:
Heads up: this PR's payload is currently nested — \`ApiVolunteerRegister = { volunteer: VolunteerFormData }\` and \`registerMutation.mutateAsync({ volunteer: value })\` (register-mutation hook + the form submit).

We just decided (on sdk#222, in response to an open review question) to **not nest** — the request body should be the flat \`ApiVolunteerRegisterNew\` shape directly, no \`volunteer\` wrapper key. be#950 (the backend endpoint this posts to) has already been updated to expect the flat body instead of \`{ volunteer: ... }\`.

Once sdk#222 lands flat, this PR will need the same change: drop the wrapper in both the local type and the \`mutateAsync\` call, so the request body is just the form data itself.

arturasmckwcz:
Update: sdk#222 has merged and published as `need4deed-sdk@0.0.152` — flat `ApiVolunteerRegisterNew`, no `{ volunteer: ... }` wrapper, matching be#950. Here's the full list of what this PR's submission needs to change to match it (beyond the wrapper I flagged earlier), based on nadavosa's field-by-field analysis on sdk#222:

**1. Drop the wrapper.** `useRegisterVolunteer.ts` currently POSTs `{ volunteer: payload }` — should POST the converted body directly (no `volunteer` key).

**2. `locations`/`activities`/`skills`/`leadFrom`: `number[]` → `OptionById[]`/`OptionItem[]`.** `ProfileCompletionData` currently sends bare id arrays; the contract needs `{ id, title }` objects. Since the EditableField dropdowns already have the title for each selected option (that's what's rendered), resolve each id back to its `{id, title}` from whatever options list the form already has loaded, rather than sending raw ids.

**3. `languages`: `LanguageObject[]` → `ApiLanguage[]`.** Current shape is `{ id, language, level }` where (per nadavosa's comment on sdk#222) `language` is a stringified option id, not a title, and `id` is just the form row key. Target shape is `{ id, title, proficiency? }` — `id`/`title` need to come from resolving the selected language option, and `level` (`LanguageLevel`: native/fluent/intermediate) maps onto `proficiency` (`LangProficiency`: adds beginner/advanced) directly for the 3 shared values; omit `proficiency` when `level` is `""`.

**4. `availability`: `Availability` (`{weekday, timeSlots}[]`) → `ApiAvailability[]` (`{day, daytime}[]`).** Good news — this conversion already exists and is reused elsewhere in this repo: `formToApiAvailability` in `src/components/Dashboard/Profile/sections/VolunteerProfile/availabilityUtils.ts`. Just call that on `form.availability` instead of sending it raw.

Happy to pair on this if useful — the shapes are now locked in `need4deed-sdk@0.0.152`, so this is the target to convert to.

DarrellRoberts:
thanks @arturasmckwcz , that was really helpful. I've pushed my changes & basically just processed everything just before submit rather than changing the onChange handler for the fields

Haven't had a chance to test manually yet but will do.

also fyi the `optionslist` for `languages` seem to be incomplete and the `leadFrom` list from options seems to be empty

Issue #963, opened 2026-08-31
feat: Post card — author/tags/opportunity chips, three-dot edit/delete menu with confirm dialog


arturasmckwcz:
## Context

Post display + management for the Posts page rebuild. Not blocked on new backend work — uses the existing `PATCH`/`DELETE /post/:id` endpoints (be#680).

## Scope

- Post card: author name/avatar, timestamp, text with tag rendering (reuse the same rendering used for Comments), linked-opportunity chips that link to the opportunity's profile
- Three-dot menu on a post: Edit / Delete (author-only, or admin/coordinator per existing role conventions)
- Confirmation pop-up before delete
- Removes the old placeholder's hardcoded-disabled Send button and no-op textarea (final cleanup of the #892 placeholder)

## Acceptance criteria

- [ ] Post card shows author, timestamp, tagged users, and clickable opportunity chips
- [ ] Three-dot menu offers Edit and Delete to the post's author
- [ ] Deleting a post requires confirmation first
- [ ] Old placeholder composer/button code is removed

Part of Epic: #969

Issue #961, opened 2026-08-31
feat: Posts composer — free text, @mention (incl. @all), multi-opportunity picker, emoji picker


arturasmckwcz:
## Context

Base composer for the Posts page rebuild (fe#892 updated per prototype proposal + amendment). Not blocked on any new backend work — uses the existing `POST /post` endpoint (be#680).

## Scope

- Textarea composer, anchored at the **bottom** of the feed (per the amendment — supersedes #892's original top-composer assumption)
- Same composer component is reused for replies (see #964) — switches into reply mode showing the target's author, timestamp, and original-message context
- @mention: reuse `Autocomplete`/`useCommentTag` from the Comments section (no rebuild) — live suggestions as the user types `@`, including a special `@all` entry, used consistently; mentions insert at the current cursor position, including into existing or pasted text
- Opportunity linking: search-driven picker, **multiple** opportunities selectable per post (the `Post` entity's `linkedOpportunities` is already a m2m, so no BE change needed here). Picker stays open while selecting, closes via a "Done" action or on leaving it; selected opportunities shown as removable chips
- Emoji picker inserted directly into the composer (amendment added this on top of the original proposal)
- `useCreatePost` hook wrapping `useMutationQuery`, POSTing to `/post`

## Acceptance criteria

- [ ] Composer sits at the bottom of the feed
- [ ] Typing `@` surfaces user suggestions incl. `@all`; selecting one tags them; mentions insert at the cursor position, including into existing/pasted text
- [ ] Multiple opportunities can be attached via a picker that stays open during selection and closes on "Done" or on leaving it; selections shown as removable chips
- [ ] Emoji picker inserts an emoji into the composer text
- [ ] Submitting creates a post via `POST /post` and it appears in the feed

Part of Epic: #969 (see Round 3 corrections, 2026-09-01, points 1, 7, 8)

PR #973, opened 2026-09-02
feat: add reverse infinite posts feed (#962)

### LINKED ISSUE #962: feat: Posts feed — chat-direction ordering (oldest top, newest bottom) + pagination
## Context

Feed layout for the Posts page rebuild. The amendment reversed the original proposal's ordering: posts now read like a chat, oldest at the top, newest at the bottom, with the composer pinned below the feed (see fe#961 for the composer itself). This also reverses #892's original "newest first" acceptance criterion — that's superseded.

## Scope

- Feed renders oldest → newest top-to-bottom, auto-scrolled to the bottom (newest) on load, matching chat UX
- "Load older" fetches the previous page and prepends it at the top (reverse-infinite-scroll), replacing #892's original "More posts" button-at-bottom pattern
- Uses existing `GET /post` pagination — no BE change needed

## Acceptance criteria

- [ ] Feed opens scrolled to the newest post at the bottom
- [ ] Scrolling up near the top loads and prepends older posts
- [ ] No layout jump when older posts are prepended

Part of Epic: #969

Justber79:
## Description

Implements the Posts feed as a chat-style timeline. Posts are displayed oldest to newest, the feed opens at the newest post, and older pages load as the user scrolls upward without changing their visible position.

## Related Issues

Closes #962

## Changes

- Add paginated post fetching with the existing `GET /post` endpoint
- Display loaded posts in chronological order
- Automatically position the feed at the newest post on initial load
- Load and prepend older posts near the top of the feed
- Preserve scroll position when older posts are inserted
- Add loading, error, and empty feed states
- Keep the existing placeholder composer at the bottom for follow-up issue #961

## Screenshots / Demos

Tested locally with 45 posts across three API pages. The final post-card styling is covered by #963.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes


arturasmckwcz (commented):
@Justber79 thanks for the PR! Ran a review and found a few correctness issues worth fixing before merge, plus two smaller polish items — left inline. CI (lint/typecheck) is green; the main gaps are around edge cases the 45-post/3-page manual test wouldn't have hit (no overflow, a failed background page-fetch, or a language switch mid-session).

src/components/Dashboard/Posts/PostFeed.tsx
```diff
@@ -0,0 +1,125 @@
...
+    if (result.isError) pendingScrollMetrics.current = null;
+  }, [fetchNextPage, hasNextPage, isFetchingNextPage]);
+
+  const handleScroll = useCallback(() => {
+    const feed = feedRef.current;
+    if (!feed) return;
+
+    const currentScrollTop = feed.scrollTop;
+    const isScrollingUp = previousScrollTop.current !== null && currentScrollTop < previousScrollTop.current;
+    previousScrollTop.current = currentScrollTop;
```
arturasmckwcz: **Infinite scroll never fires without overflow.** Loading older posts is wired only to the native `scroll` event on `feed`. If the loaded posts don't overflow the container — few total posts, or a tall container on a large viewport (`height: max(520px, calc(100dvh - ...))`) — `feed.scrollHeight <= feed.clientHeight`, so it's not scrollable and `handleScroll` never runs. `loadOlderPosts` becomes unreachable even while `hasNextPage` is true, silently stranding older pages a user can never trigger by scrolling.

src/components/Dashboard/Posts/PostFeed.tsx
```diff
@@ -0,0 +1,125 @@
...
+    );
+  }
+
+  if (isError) {
+    return (
+      <EmptyState role="alert">
+        <Paragraph>{error?.message || t("message.errorGeneric")}</Paragraph>
+      </EmptyState>
+    );
+  }
```
arturasmckwcz: **A failed background page-fetch wipes the whole feed.** When `loadOlderPosts`'s `fetchNextPage()` fails, TanStack Query sets `isError: true` while `data.pages` (and `posts`) still holds everything already loaded — but this `isError` branch runs before the main render branch, so a transient network blip while scrolling replaces the entire loaded feed + scroll position with a generic full-page error, discarding valid on-screen data. Consider distinguishing the initial-load error from a failed `fetchNextPage` (e.g. only short-circuit to the error state when `posts.length === 0`).

src/components/Dashboard/Posts/PostFeed.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  const posts = useMemo(() => data?.pages.flatMap((page) => page.data).reverse() ?? [], [data?.pages]);
+
+  useLayoutEffect(() => {
+    const feed = feedRef.current;
+    if (!feed || posts.length === 0) return;
+
+    if (!didScrollToNewest.current) {
+      feed.scrollTop = feed.scrollHeight;
+      previousScrollTop.current = feed.scrollTop;
```
arturasmckwcz: **Scroll-to-newest can be skipped after an in-place language switch.** `didScrollToNewest`/`pendingScrollMetrics`/`previousScrollTop` are refs on this component instance and never reset when the underlying query key changes without a remount. `usePostsFeed()`'s query key includes `lang`, and the app's language switcher does a client-side route change that swaps only the `[lang]` segment — `PostFeed` isn't remounted. Switching language while viewing Posts starts a new query; `FeedScrollContainer` re-renders at `scrollTop 0`, but `didScrollToNewest.current` is already `true` from before, so the 'scroll to bottom' branch is skipped and the feed opens scrolled to the oldest post instead of the newest.

src/components/Dashboard/Posts/PostFeed.tsx
```diff
@@ -0,0 +1,125 @@
...
+  if (isError) {
+    return (
+      <EmptyState role="alert">
+        <Paragraph>{error?.message || t("message.errorGeneric")}</Paragraph>
+      </EmptyState>
+    );
+  }
+
+  if (posts.length === 0) {
+    return (
```
arturasmckwcz: **Raw Axios error message, not localized.** `error` here is a raw `AxiosError`, so `error.message` is always a truthy generic string (e.g. "Request failed with status code 500", "Network Error") — never the backend's own `response.data.message` that `useGetQuery.tsx` normally unwraps elsewhere in the app. The `|| t("message.errorGeneric")` fallback is effectively dead code, and users see an untranslated, technical string instead of the established error-message pattern.

src/components/Dashboard/Posts/styles.ts
```diff
@@ -4,16 +4,27 @@ export const NewPostSection = styled.div`
...
   flex-direction: column;
   gap: var(--spacing-4);
+  flex-shrink: 0;
 `;
 
 export const PostsContainer = styled.div`
   display: flex;
   flex-direction: column;
   gap: var(--spacing-16);
+  height: max(
```
arturasmckwcz: **Height calc omits the mobile nav-bar padding, can hide the composer.** This only subtracts `--dashboard-base-container-padding-bottom`, but on mobile (<=767px) `DashboardBaseContainer`'s actual bottom padding also adds `--dashboard-navigation-bar-mobile-height` to clear the fixed bottom nav. `PostsContainer` ends up taller than the visible area by that amount, pushing `NewPostSection` (the composer, right after this in the flex column) down behind/under the fixed mobile nav bar.

src/components/Dashboard/Posts/styles.ts
```diff
@@ -22,6 +33,52 @@ export const EmptyState = styled.div`
...
+  flex: 1;
+  min-height: 0;
+  flex-direction: column;
+  gap: var(--spacing-12);
+  overflow-y: auto;
+  overscroll-behavior: contain;
+  padding: 0 var(--spacing-8) var(--spacing-8);
+`;
+
+export const LoadOlderIndicator = styled.div`
```
arturasmckwcz: **Minor:** `LoadOlderIndicator` always reserves `min-height: var(--spacing-24)` even once `hasNextPage` is `false` (no more older posts), leaving a permanent empty gap at the top of the feed. Might be worth conditionally rendering it only while `hasNextPage` is true.

Justber79:
@arturasmckwcz Thanks for the detailed review! I’ve addressed all six points and pushed the fixes. I also reran TypeScript and lint successfully.

arturasmckwcz (approved):
LGTM

PR #976, opened 2026-09-03
feat: add post cards and management controls (#963)

### LINKED ISSUE #963: feat: Post card — author/tags/opportunity chips, three-dot edit/delete menu with confirm dialog

Justber79:
## Description

Adds the complete post-card presentation and management controls for the Posts page. Posts now match the dashboard styling and clearly show their author, timestamp, mentions, and linked opportunities, while permitted users can edit or delete posts safely.

## Related Issues

Closes #963

## Changes

- Add styled post cards with author name, avatar or branded initials, and timestamp
- Render tagged people clearly inside post text
- Render linked opportunities as clickable chips leading to their profiles
- Add an author/admin/coordinator three-dot menu with Edit and Delete actions
- Add inline post editing using the existing PATCH endpoint
- Add a confirmation dialog before permanent deletion
- Remove the old disabled composer placeholder in preparation for #961
- Add English and German labels for the new controls

## Screenshots / Demos

963-post-card-with-menu png

963-post-delete-confirmation png

## Checklist

- [ ] Tests added/updated
- [ ] Documentation updated
- [ ] CI passes

### DIFF
--- a/src/components/Dashboard/Posts/PostActionMenu.tsx
+++ b/src/components/Dashboard/Posts/PostActionMenu.tsx
@@ -1,0 +1,49 @@
+import { useEffect, useRef } from "react";
+import { useTranslation } from "react-i18next";
+
+import { ActionMenu, ActionMenuItem } from "./styles";
+
+type Props = {
+  isOpen: boolean;
+  onClose: () => void;
+  onEdit: () => void;
+  onDelete: () => void;
+};
+
+export function PostActionMenu({ isOpen, onClose, onEdit, onDelete }: Props) {
+  const { t } = useTranslation();
+  const menuRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (!isOpen) return;
+
+    const closeOnOutsideClick = (event: MouseEvent) => {
+      if (!menuRef.current?.contains(event.target as Node)) onClose();
+    };
+    const closeOnEscape = (event: KeyboardEvent) => {
+      if (event.key === "Escape") onClose();
+    };
+
+    document.addEventListener("mousedown", closeOnOutsideClick);
+    document.addEventListener("keydown", closeOnEscape);
+    return () => {
+      document.removeEventListener("mousedown", closeOnOutsideClick);
+      document.removeEventListener("keydown", closeOnEscape);
+    };
+  }, [isOpen, onClose]);
+
+  if (!isOpen) return null;
+
+  return (
+    <ActionMenu ref={menuRef} role="menu">
+      <ActionMenuItem role="menuitem" onClick={onEdit}>
+        {t("dashboard.posts.edit")}
+      </ActionMenuItem>
+      <ActionMenuItem role="menuitem" $danger onClick={onDelete}>
+        {t("dashboard.posts.delete")}
+      </ActionMenuItem>
+    </ActionMenu>
+  );
+}
+
+export default PostActionMenu;
--- a/src/components/Dashboard/Posts/PostActionMenu.tsx
+++ b/src/components/Dashboard/Posts/PostActionMenu.tsx
@@ -1,3 +1,4 @@
+import { useClickOutside } from "@/hooks";
 import { useEffect, useRef } from "react";
 import { useTranslation } from "react-i18next";
 
@@ -13,21 +14,17 @@ type Props = {
 export function PostActionMenu({ isOpen, onClose, onEdit, onDelete }: Props) {
   const { t } = useTranslation();
   const menuRef = useRef<HTMLDivElement>(null);
+  useClickOutside(menuRef, onClose);
 
   useEffect(() => {
     if (!isOpen) return;
 
-    const closeOnOutsideClick = (event: MouseEvent) => {
-      if (!menuRef.current?.contains(event.target as Node)) onClose();
-    };
     const closeOnEscape = (event: KeyboardEvent) => {
       if (event.key === "Escape") onClose();
     };
 
-    document.addEventListener("mousedown", closeOnOutsideClick);
     document.addEventListener("keydown", closeOnEscape);
     return () => {
-      document.removeEventListener("mousedown", closeOnOutsideClick);
       document.removeEventListener("keydown", closeOnEscape);
     };
   }, [isOpen, onClose]);
--- a/src/components/Dashboard/Posts/PostCard.tsx
+++ b/src/components/Dashboard/Posts/PostCard.tsx
@@ -24,6 +24,7 @@ import {
   PostHeader,
   PostHeaderText,
   PostMenuButton,
+  PostMenuWrapper,
   PostText,
   PostTimestamp,
 } from "./styles";
@@ -38,13 +39,14 @@ export function PostCard({ post }: Props) {
   const [isEditing, setIsEditing] = useState(false);
   const [isDeleteOpen, setIsDeleteOpen] = useState(false);
   const [editText, setEditText] = useState(post.text);
-
-  const closeEdit = useCallback(() => {
-    setIsEditing(false);
-    setEditText(post.text);
-  }, [post.text]);
   const updatePost = useUpdatePost(post.id, () => setIsEditing(false));
   const deletePost = useDeletePost(post.id, () => setIsDeleteOpen(false));
+
+  const closeEdit = useCallback(() => {
+    if (updatePost.isPending) return;
+    setIsEditing(false);
+    setEditText(post.text);
+  }, [post.text, updatePost.isPending]);
 
   const canManage =
     currentUser?.personId === post.author.id ||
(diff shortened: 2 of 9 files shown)


src/components/Dashboard/Posts/PostCard.tsx
```diff
@@ -0,0 +1,179 @@
...
+    setIsEditing(true);
+    setIsMenuOpen(false);
+  };
+
+  const saveEdit = () => {
+    let formattedText = editText.trim();
+    post.taggedPersons.forEach((person) => {
+      formattedText = formattedText.replaceAll(`@${person.fullName}`, `<@${person.id}>`);
+    });
+    updatePost.mutate({ text: formattedText });
```
arturasmckwcz: **[Mandatory]** Saving an edit only PATCHes `{ text }`, never `taggedPersonIds`/`linkedOpportunityIds`, even though editing lets the user remove an `@mention` from the text. `be/src/server/routes/post.routes.ts` only updates `post.taggedPersons` when `taggedPersonIds` is explicitly present in the body — since this never sends it, deleting an `@mention` from the text during edit leaves that person permanently listed in the post's `taggedPersons` relation, diverging from what the text actually shows. Same applies to `linkedOpportunityIds` if opportunity chips can be removed during edit.

src/components/Dashboard/Posts/PostCard.tsx
```diff
@@ -0,0 +1,179 @@
...
+    });
+    setEditText(editableText);
+    setIsEditing(true);
+    setIsMenuOpen(false);
+  };
+
+  const saveEdit = () => {
+    let formattedText = editText.trim();
+    post.taggedPersons.forEach((person) => {
+      formattedText = formattedText.replaceAll(`@${person.fullName}`, `<@${person.id}>`);
```
arturasmckwcz: **[Mandatory]** `saveEdit()` converts `@FullName` back into `<@id>` tags via a literal `String.replaceAll` with no word-boundary check, so one tagged person's name being a prefix/substring of another's corrupts the save. Example: post tags "Anna" (id 1) and "Anna Marie" (id 2) in text `Hi <@1> and <@2>`. Edit opens as `Hi @Anna and @Anna Marie`. On save, the forEach hits Anna first: `replaceAll('@Anna', '<@1>')` also matches the substring inside `@Anna Marie`, producing `Hi <@1> and <@1> Marie` — Anna Marie's tag is silently replaced with Anna's id and "Marie" becomes stray text. Needs a word-boundary-aware replace (or working from stored mention positions instead of re-parsing the displayed text).

arturasmckwcz (commented):
@Justber79 thanks for the PR! Found 6 issues — 4 marked **mandatory** to fix before merge, 2 left as optional polish. CI (lint/typecheck) is green.

src/components/Dashboard/Posts/PostActionMenu.tsx
```diff
@@ -0,0 +1,49 @@
...
+  onClose: () => void;
+  onEdit: () => void;
+  onDelete: () => void;
+};
+
+export function PostActionMenu({ isOpen, onClose, onEdit, onDelete }: Props) {
+  const { t } = useTranslation();
+  const menuRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
```
arturasmckwcz: **[Mandatory]** This re-implements outside-click/escape-close logic and an Edit/Delete action-menu pattern that already exists in this codebase — `src/hooks/useClickOutside.ts` (already used by `TrustLevelDropdown`/`StepTitle`/`Menu.tsx`) and `src/components/Dashboard/Profile/sections/Comments/common/CommentActionMenu.tsx`, which has the identical `isOpen`/`onClose`/`onEdit`/`onDelete` shape. Per shared-rules.md's "reuse before you create": please reuse `useClickOutside` and ideally the existing `CommentActionMenu` pattern rather than maintaining a second near-duplicate three-dot menu — a future fix to the close/escape handling (e.g. a scroll/touch edge case) won't propagate here otherwise.

src/components/Dashboard/Posts/PostCard.tsx
```diff
@@ -0,0 +1,179 @@
...
+          <Avatar src={getImageUrl(post.author.avatarUrl)} alt="" />
+        ) : (
+          <AvatarInitials aria-hidden="true">{authorInitials}</AvatarInitials>
+        )}
+        <PostHeaderText>
+          <PostAuthor>{post.author.fullName}</PostAuthor>
+          <PostTimestamp dateTime={createdAt.toISOString()}>{createdAt.toLocaleString(i18n.language)}</PostTimestamp>
+        </PostHeaderText>
+        {canManage && (
+          <div style={{ position: "relative", marginLeft: "auto" }}>
```
arturasmckwcz: **[Mandatory]** This positioning wrapper uses a raw inline `style` object instead of a styled-component, inconsistent with every other element in this file (`Avatar`, `PostHeaderText`, etc. are all styled-components). Please pull this into a styled-component (e.g. `PostMenuWrapper`) so theming/spacing-token changes applied elsewhere reach it too, and future layout refactors don't need to special-case this one raw-styled div.

src/components/Dashboard/Posts/PostCard.tsx
```diff
@@ -0,0 +1,179 @@
...
+      </PostBody>
+
+      {isDeleteOpen && (
+        <ConfirmationDialog
+          title={t("dashboard.posts.deleteDialog.title")}
+          message={t("dashboard.posts.deleteDialog.message")}
+          confirmText={t("dashboard.posts.delete")}
+          cancelText={t("dashboard.posts.cancel")}
+          onCancel={() => setIsDeleteOpen(false)}
+          onConfirm={() => deletePost.mutate()}
```
arturasmckwcz: **[Optional]** The confirm button in `ConfirmationDialog` isn't disabled while `deletePost.isPending`, so a fast double-click fires two DELETE requests — the first succeeds, the second 404s on the now-missing post and surfaces an unexpected error toast right after a successful deletion. Consider disabling/guarding on `isPending`.

src/components/Dashboard/Posts/PostCard.tsx
```diff
@@ -0,0 +1,179 @@
...
+          </div>
+        )}
+      </PostHeader>
+
+      <PostBody>
+        {isEditing ? (
+          <>
+            <EditTextArea value={editText} onChange={(event) => setEditText(event.target.value)} autoFocus />
+            <EditActions>
+              <EditButton type="button" onClick={closeEdit}>
```
arturasmckwcz: **[Optional]** The Cancel button in edit mode isn't disabled while `updatePost.isPending`, so a user can cancel out of an edit whose PATCH is still in flight — the editing UI closes, but the in-flight request still completes and fires a success toast plus a cache refresh the user believed they'd discarded. Consider disabling Cancel (or ignoring the stale response) while pending.

Justber79:
@arturasmckwcz Thanks for the review! I addressed all four mandatory findings and the two optional improvements:
- Edit requests now include updated taggedPersonIds and preserve linkedOpportunityIds.
- Mention conversion now processes longer names first and uses boundary-aware matching.
- The post action menu now reuses useClickOutside.
- The inline positioning style was moved into PostMenuWrapper.
- Delete confirmation controls are disabled while deletion is pending.
- Edit Cancel is disabled while saving.

I also tested mention removal end to end and confirmed that the linked opportunities remain attached. Typecheck and lint pass. The fixes are included in commit 6e051b9e.

src/components/Dashboard/Posts/PostActionMenu.tsx
```diff
@@ -0,0 +1,46 @@
...
+  isOpen: boolean;
+  onClose: () => void;
+  onEdit: () => void;
+  onDelete: () => void;
+};
+
+export function PostActionMenu({ isOpen, onClose, onEdit, onDelete }: Props) {
+  const { t } = useTranslation();
+  const menuRef = useRef<HTMLDivElement>(null);
+  useClickOutside(menuRef, onClose);
```
arturasmckwcz: Minor/non-blocking: `useClickOutside` attaches its `mousedown` listener unconditionally here, whereas the previous hand-rolled version only attached it while `isOpen`. Functionally harmless (closing an already-closed menu is a no-op re-render), just means the listener is technically always live rather than only while the menu is open.

arturasmckwcz (commented):
@Justber79 re-reviewed after the fix commit — all 6 findings from the previous pass are correctly addressed: taggedPersonIds/linkedOpportunityIds now sync on edit, the mention re-serialization is fixed properly (longest-name-first + word-boundary regex, verified against the "Anna" / "Anna Marie" case by hand), PostActionMenu now reuses useClickOutside, the inline style is now a styled-component, and both the delete-confirm and edit-cancel buttons are disabled while their mutation is pending (nice touch also gating ConfirmationDialog's backdrop close on cancelDisabled). One trivial, non-blocking observation left inline. CI is green.

arturasmckwcz (approved):
LGTM

Issue #647, opened 2026-06-13
feat: public event registration form — replace Google Forms with in-house signup per event


nadavosa:
## Overview

Coordinators currently share Google Form links for event signups. This issue replaces that with an in-house public registration form, accessible without login, scoped to a specific event (dated non-accompanying opportunity).

Each event gets a shareable URL. Coordinators can view registrations in the dashboard on the opportunity profile.

---

## Public form: \`/forms/event/[id]\`

Publicly accessible — no login required. Same pattern as \`/forms/volunteer\` (TanStack Form, \`usePostRequest\`).

The form pre-loads the event title and date from the opportunity (read-only display at the top) so the registrant knows what they're signing up for.

### Fields

| Field | Required | Notes |
|---|---|---|
| Full name | Yes | |
| Email | Yes | validated |
| Phone | No | |
| Number of people | Yes | default 1 (for groups) |
| Language preference | No | dropdown, same options as volunteer form |
| Message / notes | No | free text, max 500 chars |

### Behaviour
- Submit → success screen ("You're registered! We'll be in touch.") — same thank-you pattern as BecomeVolunteer
- If opportunity ID is invalid or event is in the past → show "This event is no longer available" message
- Form is bilingual (en/de) following existing i18n pattern

### Shareable link
Coordinators copy the URL \`/en/forms/event/[id]\` from the opportunity profile and share it externally (WhatsApp, email, etc.).

---

## Dashboard: registrations view on opportunity profile

On opportunity profiles for events (dated, non-accompanying), add a **Registrations** section visible to Coordinators and NGO users:

- Table: Name | Email | Phone | People | Message | Registered at
- Row count shown in section header ("3 registrations")
- Export to CSV button (simple \`window.open\` with a download endpoint — see BE issue)
- No editing — registrations are read-only in the dashboard

---

## Copy-link button on opportunity profile

For event opportunities, show a **"Copy registration link"** button (clipboard copy of \`/en/forms/event/[id]\`) so coordinators can easily grab and share it.

---

## FE tasks

- [ ] New page: \`src/app/[lang]/forms/event/[id]/page.tsx\`
- [ ] \`EventRegistrationForm\` component (TanStack Form, mirrors BecomeVolunteer structure)
- [ ] Pre-load event title + date from \`GET /opportunity/:id\` (public, no auth)
- [ ] Success and invalid-event states
- [ ] \`Registrations\` section component for the opportunity profile (coordinator/NGO only)
[...]

PR #978, opened 2026-09-03
feat: add public event details page (#941)

### LINKED ISSUE #941: Public page: port /event-page (event details) from website to fe
## Description
`website`'s `/event-page/:lng` is a standalone page showing the current event's full details (title, meta, description) with a registration link. `fe`'s Header/homepage currently send the "Events" nav item to `eventsPublicLandingUrl` (`https://www.need4deed.org/event-page`, `src/config/constants.ts:51`) — `fe` has no internal page for this at all today.

## Reference implementation (website repo)
`need4deed-org/website`'s `src/pages/EventPage.tsx` — hardcodes the event's title/meta/description directly in JSX and links out to a Google Form (`REGISTRATION_URL`) for registration.

## Notes / overlaps with other in-flight work
- **Registration**: fe#647 already tracks replacing the Google Form registration flow with an in-house signup per event — this issue should NOT duplicate that; it covers the static content page/shell only, linking to whatever registration entry point fe#647 lands on (or the Google Form as an interim step).
- **Event data source**: fe#917 (homepage Events section, still open) flags that `website` has no single authoritative source for "the current event" yet (its own homepage card and its `/event-page` both hardcode independently). Whatever data source fe#917 settles on should be reused here rather than a third hardcoded copy.

## Ask
- Add a new public route (e.g. `/[lang]/event-page`) rendering the event's title/meta/description, sourced the same way fe#917 resolves for the homepage banner.
- Once live, update `eventsPublicLandingUrl` usages in `Header.tsx`/`MenuitemList.tsx` to point internally instead of the external `website` domain.

## Related
Depends on the data-source decision in fe#917. Part of the broader website→app content-transfer effort tracked in fe#919.

Justber79:
## Description

Adds a localized public event-details page inside the FE application. It reuses the same live event API and upcoming-event selection as the landing-page Events section, so event content created by coordinators is shown consistently without another hardcoded source.

## Related Issues

Closes #941

## Changes

- Add the public `/[lang]/event-page` route with title, description, date, address, additional information, and registration action
- Reuse shared upcoming-event selection and localized date formatting with the #917 landing banner
- Update header and landing-page Events links to navigate internally
- Remove the legacy redirect that sent `/event-page` directly to Google Forms
- Use each event's API-provided registration URL while keeping in-app registration outside this issue (#647)
- Add localized loading, empty, and event-page labels in English and German
- Keep the public route accessible when a stale authentication cookie is present

## Screenshots / Demos

941-event-details-page

941-landing-events-banner

## Checklist

- [ ] Tests added/updated — no automated test runner is configured; typecheck, lint, production build, and manual end-to-end event creation/navigation/deletion were verified
- [ ] Documentation updated — not required for this UI change
- [ ] CI passes — pending

### DIFF
--- a/src/components/EventPage/EventPage.tsx
+++ b/src/components/EventPage/EventPage.tsx
@@ -1,0 +1,168 @@
+"use client";
+
+import { Button } from "@/components/core/button";
+import { PageLayout } from "@/components/Layout";
+import { Heading1, Heading2, Paragraph } from "@/components/styled/text";
+import { useEvents } from "@/hooks/useEvents";
+import { formatEventDate, getHttpUrl, getUpcomingEvent } from "@/utils/events";
+import { CalendarBlankIcon, MapPinIcon } from "@phosphor-icons/react";
+import { useMemo } from "react";
+import { useTranslation } from "react-i18next";
+import styled from "styled-components";
+
+const PageContent = styled.main`
+  width: min(100% - 32px, 960px);
+  margin: 0 auto;
+  padding: clamp(40px, 7vw, 88px) 0;
+`;
+
+const EventCard = styled.article`
+  overflow: hidden;
+  border: 1px solid var(--color-orchid-light, var(--color-orchid));
+  border-radius: 24px;
+  background: var(--color-white);
+  box-shadow: 0 16px 40px rgb(40 25 47 / 10%);
+`;
+
+const Hero = styled.header`
+  padding: clamp(28px, 6vw, 64px);
+  background: linear-gradient(135deg, var(--color-orchid-subtle), var(--color-orchid));
+`;
+
+const EventType = styled.span`
+  display: inline-flex;
+  margin-bottom: var(--spacing-20);
+  padding: 8px 14px;
+  border-radius: 999px;
+  background: var(--color-midnight);
+  color: var(--color-white);
+  font-size: var(--font-size-xs);
+  font-weight: var(--font-weight-semibold);
+  text-transform: capitalize;
+`;
+
+const Subtitle = styled(Paragraph)`
+  margin-top: var(--spacing-12);
+`;
+
+const Body = styled.div`
+  display: grid;
+  grid-template-columns: minmax(0, 1fr) minmax(240px, 300px);
+  gap: clamp(32px, 6vw, 72px);
+  padding: clamp(28px, 6vw, 64px);
+
+  @media (max-width: 767px) {
+    grid-template-columns: 1fr;
+  }
+`;
+
+const Description = styled(Paragraph)`
+  margin-top: var(--spacing-20);
+  white-space: pre-line;
+`;
+
+const AdditionalInfo = styled.ul`
+  margin: var(--spacing-20) 0 0;
+  padding-left: 20px;
+  color: var(--color-midnight);
+`;
+
+const Details = styled.aside`
+  display: flex;
+  flex-direction: column;
+  gap: var(--spacing-20);
+`;
+
+const Detail = styled.div`
+  display: grid;
+  grid-template-columns: 24px minmax(0, 1fr);
+  gap: var(--spacing-12);
+  align-items: start;
+  color: var(--color-midnight);
+`;
+
+const EmptyState = styled.div`
+  padding: clamp(40px, 8vw, 96px) 24px;
+  text-align: center;
+`;
+
+export function EventPage() {
+  const { t, i18n } = useTranslation();
+  const { data: events, isLoading } = useEvents();
+  const event = useMemo(() => getUpcomingEvent(events), [events]);
+  const registrationUrl = getHttpUrl(event?.linkRSVP);
+
+  return (
+    <PageLayout>
+      <PageContent>
+        {isLoading ? (
+          <EmptyState aria-live="polite">
+            <Heading2>{t("eventPage.loading")}</Heading2>
+          </EmptyState>
+        ) : event ? (
+          <EventCard>
+            <Hero>
+              <EventType>{event.type}</EventType>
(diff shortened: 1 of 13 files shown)


src/components/EventPage/EventPage.tsx
```diff
@@ -0,0 +1,168 @@
...
+    <PageLayout>
+      <PageContent>
+        {isLoading ? (
+          <EmptyState aria-live="polite">
+            <Heading2>{t("eventPage.loading")}</Heading2>
+          </EmptyState>
+        ) : event ? (
+          <EventCard>
+            <Hero>
+              <EventType>{event.type}</EventType>
```
nadavosa: The event type badge renders the raw enum value (`event.type`, e.g. "party"/"workshop") instead of a translated label. The codebase already has translated labels for the same enum at `dashboard.calendar.createForm.typeParty`/`typeWorkshop` (in `public/locales/de/translations.json`, `typeParty` is "Feier"). On the German event page, a Party event will show "Party" (CSS-capitalized English text) instead of "Feier".

src/utils/events.ts
```diff
@@ -0,0 +1,34 @@
...
+    year: "numeric",
+  });
+  const start = format.format(new Date(event.date));
+  if (!event.dateEnd) return start;
+
+  const end = format.format(new Date(event.dateEnd));
+  return start === end ? start : `${start} – ${end}`;
+}
+
+export function getHttpUrl(value?: string) {
```
nadavosa: This duplicates the existing `getHttpUrl` in `src/components/Dashboard/Calendar/Calendar/EventCard.tsx` (same http(s)-URL validation, differing only in returning `undefined` here vs `null` there). Worth extracting one shared helper instead of maintaining two copies of the same logic.

src/components/EventPage/EventPage.tsx
```diff
@@ -0,0 +1,168 @@
...
+                <Heading2>{t("eventPage.about")}</Heading2>
+                <Description>{event.description}</Description>
+                {event.additionalInfo?.length ? (
+                  <>
+                    {event.additionalTitle && (
+                      <Heading2 margin="var(--spacing-32) 0 0">{event.additionalTitle}</Heading2>
+                    )}
+                    <AdditionalInfo>
+                      {event.additionalInfo.map((item) => (
+                        <li key={item}>{item}</li>
```
nadavosa: Using the item's own text as the React key (`key={item}`) will collide if `additionalInfo` contains two identical strings, a plausible content-entry case, causing a duplicate-key warning and unreliable reconciliation on updates. Prefer `key={index}` since this is a static, non-reorderable list rendered from array position.

src/components/EventPage/EventPage.tsx
```diff
@@ -0,0 +1,168 @@
...
+  grid-template-columns: minmax(0, 1fr) minmax(240px, 300px);
+  gap: clamp(32px, 6vw, 72px);
+  padding: clamp(28px, 6vw, 64px);
+
+  @media (max-width: 767px) {
+    grid-template-columns: 1fr;
+  }
+`;
+
+const Description = styled(Paragraph)`
```
nadavosa: A fetch error is indistinguishable from "no upcoming event" here — `useEvents()` (via `useGetQuery`) also returns `isError`, but only `data`/`isLoading` are destructured. If `GET /event` fails, `events` stays `undefined`, `getUpcomingEvent(undefined)` returns `undefined`, and the page silently renders the same "No upcoming event" empty state as a real API outage would. Worth destructuring `isError` too and showing a distinct error state, so a backend/network issue doesn't look like "nothing scheduled" to a visitor.

Rest of the diff looks solid: the `getUpcomingEvent`/`formatEventDate` extraction into `utils/events.ts` removes the duplication with `EventsSection`, `getHttpUrl`'s protocol check is a sensible guard before opening `event.linkRSVP` in a new tab, and the localized-path changes to `eventsPublicLandingUrl` are complete (only 3 call sites, all updated in this diff).

Justber79:
@nadavosa Thanks for the review — addressed all four findings in b8499ead:

- translated the event type badge using the existing calendar event-type labels
- extracted the duplicate URL validation by reusing the shared `getHttpUrl` helper
- changed the static additional-info list to use its array index as the key
- added a distinct localized error state for failed event requests

`npm run typecheck` and `npm run lint` both pass locally (only the existing unrelated lint warnings remain).

src/components/EventPage/EventPage.tsx
```diff
@@ -0,0 +1,168 @@
...
+import styled from "styled-components";
+
+const PageContent = styled.main`
+  width: min(100% - 32px, 960px);
+  margin: 0 auto;
+  padding: clamp(40px, 7vw, 88px) 0;
+`;
+
+const EventCard = styled.article`
+  overflow: hidden;
```
nadavosa: Per the org rule of always using the shared component library (`src/components/core/`, plus `src/components/styled/`) rather than one-off styled-components: this page reinvents a few things that already exist.

- `EventCard` (line 20) hand-rolls card styling (`border-radius: 24px`, border, background, shadow) instead of extending `BaseCard` from `src/components/styled/container.tsx`, which several other cards in the app build on. The hardcoded `24px` in particular duplicates the `var(--card-border-radius)` token `BaseCard` already uses, so this card won't stay in sync if that token changes.
- `EmptyState` (line 85, used for both the loading and "no upcoming event" states) reimplements a basic centered-message wrapper, the same pattern `CenteredWrapper` (`src/components/core/common/CenteredWrapper.tsx`) already provides and is used for elsewhere (e.g. `dashboard/profile/page.tsx`'s loading/empty states).

Not blocking, but worth aligning with the shared primitives rather than adding a third variant of "card" and "centered empty message" to the codebase.

PR #980, opened 2026-09-04
960: fixes filters that didn't deserialize

### LINKED ISSUE #960: bug: browser back button skips a filtered list session entirely, lands on the page before it
## Description

Applying filters/sort/page on a dashboard list (Opportunities, Volunteers, Agents), then using the browser's back button, doesn't return to that filtered list state. Instead it jumps past it, landing on whatever page was open before the list was ever visited (e.g. the dashboard Home). Expected: back should return to the last list state the user was actually looking at.

## Confirmed contributing cause

All three dashboard list pages share the same pattern:
- `Opportunities.tsx`, `Volunteers.tsx`, and `Agents.tsx` each have a `useEffect` that syncs filter/sort/view-mode state to the URL via `router.replace(...)`, not `router.push`.
- Pagination (`usePageParam.ts`) does the same, `router.replace` on page change.
- The ONE `router.push` in the flow is the initial sidebar navigation into the list (`NavigationBar.tsx`, `router.push(route)`).

Net effect: no matter how much filtering/sorting/paging a user does on a list, the entire visit is a single browser-history entry (the URL just gets rewritten in place each time via `replace`). So one press of "back" pops that single entry and returns straight to whichever page preceded the list visit, skipping over the filtered state itself rather than "un-filtering" one step at a time.

## Not a regression of the closed filter-persistence issues

#356, #487, and #729 (all closed) fixed a different, related problem: filters/pagination/sort resetting when navigating from a list to a Profile page and back. Those were fixed by moving that state into URL params in the first place. That fix is working as intended and is not what's reported here, this is a distinct, previously-unaddressed side effect of using `replace` for all of those URL updates: it solves "state survives a remount" but does nothing for "back button should step back through what I was actually browsing."

## Ask

Confirm the exact repro path (which page you're on when you hit back, and what "generic homepage" is, `/dashboard` Home, or the public site) and fix so back returns to the filtered list state rather than skipping it. Likely fix shapes to weigh: pushing one history entry when a user first lands on a list with an unfiltered URL vs. a filtered one, or intercepting back navigation to restore the last list URL explicitly, rather than relying on native multi-step history through `replace`-only updates.

DarrellRoberts:
## Description
Original issue expresses error with all filters not deserializing but only present on `Volunteer` with `engagement` and `match` filters .

This is because both have unique enum values, such as `vol-matched` which was causing a condition check within the `deserializeVolunteerFilters` to return `false`.

Have added additional `"vol-"` to match its enum values, pass the condition and correctly deserialize.

Not generated with AI ;) 

## Related Issues
Closes #960 

## Changes

## Screenshots / Demos

https://github.com/user-attachments/assets/719225a0-2cb0-40ab-a126-461266f0d66f

## Checklist
- [ ] Tests added/updated
- [ ] Documentation updated
- [x] CI passes

### DIFF
--- a/src/components/Dashboard/Volunteers/helpers.ts
+++ b/src/components/Dashboard/Volunteers/helpers.ts
@@ -188,15 +188,15 @@ export function deserializeVolunteerFilters(filter: VolunteerCardsFilter, search
 
   const queryEngagement = searchParams.getAll(QueryParamsKeys.ENGAGEMENT);
   queryEngagement.forEach((e) => {
-    if (newFilter.engagement[e] !== undefined) {
-      newFilter.engagement[e] = true;
+    if (newFilter.engagement[`vol-${e}`] !== undefined) {
+      newFilter.engagement[`vol-${e}`] = true;
     }
   });
 
   const queryStatusMatch = searchParams.getAll(VolunteerStatusMatch.MATCH);
   queryStatusMatch.forEach((e) => {
-    if (newFilter.match[e] !== undefined) {
-      newFilter.match[e] = true;
+    if (newFilter.match[`vol-${e}`] !== undefined) {
+      newFilter.match[`vol-${e}`] = true;
     }
   });
 


src/components/Dashboard/Volunteers/helpers.ts
```diff
@@ -188,15 +188,15 @@ export function deserializeVolunteerFilters(filter: VolunteerCardsFilter, search
 
   const queryEngagement = searchParams.getAll(QueryParamsKeys.ENGAGEMENT);
   queryEngagement.forEach((e) => {
-    if (newFilter.engagement[e] !== undefined) {
-      newFilter.engagement[e] = true;
+    if (newFilter.engagement[`vol-${e}`] !== undefined) {
```
nadavosa: Two minor things here (same pattern repeats at line 198 for `match`):

1. The `vol-` prefix add/strip logic is now hardcoded as a string literal in 4 places: here and the assignment on the next line, plus `serializeFilters`' `.replace(/^vol-/, "")` (lines 118, 125). This exact prefix-mismatch bug (#960) already recurred once because the transform lives in multiple ad-hoc spots — worth centralizing into a shared `withVolPrefix`/`withoutVolPrefix` helper so a future SDK enum/prefix change only needs updating in one place instead of four.
2. ``vol-${e}`` gets computed twice per iteration (once in the `if` check, once in the assignment). Minor, but `const key = `vol-${e}`;` then using `key` in both spots avoids the redundant string allocation.

Not blocking, the fix itself is correct.

nadavosa: Second-pass review: neither of the two suggestions from the prior review (https://github.com/need4deed-org/fe/pull/980#discussion_r3956639974) has been addressed. `git diff fbc84ae22ae2907486511f5535bc9f6f4800ea77..3b249584de58e3ed6b1fe2aab7689131e2b57271 -- src/components/Dashboard/Volunteers/helpers.ts` is empty: this file has not changed since the review. The only commits added afterward (https://github.com/need4deed-org/fe/pull/980/commits/3b249584de58e3ed6b1fe2aab7689131e2b57271) are merges from `develop` touching an unrelated file (`NavigationBar.tsx`, brought in via #984), so there are no new issues to flag from them either. Both points (centralize the `vol-` prefix add/strip logic; avoid recomputing the template literal) remain open.

